PR #17, opened 2017-09-18
Make brpc support GCC 4.4.7, tested on Ubuntu 12.04 LTS.

solrex:
This patch makes example/echo_c++ work on Ubuntu 12.04 with GCC 4.4.7.

### DIFF
--- a/config_brpc.sh
+++ b/config_brpc.sh
@@ -39,7 +39,7 @@ elif [ -z "$CXX" ]; then
 fi
 
 GCC_VERSION=$($CXX tools/print_gcc_version.cc -o print_gcc_version && ./print_gcc_version && rm ./print_gcc_version)
-if [ $GCC_VERSION -gt 0 ] && [ $GCC_VERSION -lt 40800 ]; then
+if [ $GCC_VERSION -gt 0 ] && [ $GCC_VERSION -lt 40407 ]; then
     >&2 $ECHO "GCC is too old, please install a newer version supporting C++11"
     exit 1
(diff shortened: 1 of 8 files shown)


chenzhangyi:
atomic and memory order is not fully supported before gcc4.8(checkout https://gcc.gnu.org/gcc-4.7/cxx0x_status.html).  It's unsafe since memory order matters in brpc, You'd better try boost::atomic (which the internal version of brpc really uses),  instead of using std::atomic.

solrex:
@chenzhangyi People all say that but nobody gives why. Does GCC under 4.8 have bugs on atomic and memory order? Is there any document or bug reporting marks that?

jamesge:
The link given by @chenzhangyi clearly states memory order is not supported in gcc4.4, so this patch will **not** be merged. However brpc can be built with gcc 3.4 and 4.4 inside Baidu, with deps on boost.atomic. It shouldn't be too hard to mimic that on this opensourced version. You can have another PR or fire an issue.

solrex:
You guys are really arrogant.

lingji:
Barrier_XXX，目测是为了跨平台，但是butil/atomicops.h中定义的atomic 并没有将Barrier_XXX这一系函数包装起来，不是很明确你们的意图。是不是缺少部分呢？ @chenzhangyi 。本来的想法是不是，如果编译器或标准库没有提供atomic,则会将atomic 重定义为Barrier_XXX 系列函数，从而实现跨平台呢？

PR #63, opened 2017-09-28
move butil/third_party to third_party

thinkerou:
Add one special dir for third_party.


jamesge:
third_party in butil is orginated from base/third_party of chromium, it is very special and extensively modified to adapt usages in brpc. Moving into a top-level third_party may confuse users with that: they're still similar with the original ones, however they're not.

thinkerou:
OK, thanks @jamesge and I will close the pull request.

PR #87, opened 2017-10-25
Not find google/profiler.h when run config_brpc.sh

ibyte2011:
Add "gtest-devel" from
```
sudo yum install  gflags-devel protobuf-devel protobuf-compiler leveldb-devel
```
change to
```
sudo yum install  gtest-devel gflags-devel protobuf-devel protobuf-compiler leveldb-devel
```


jamesge:
brpc does not search for google/profiler.h in HEAD

PR #263, opened 2018-03-13
Minor fixes.

Carmot:
Deleted unused code.
Added pointers check.

### DIFF
--- a/src/brpc/amf.cpp
+++ b/src/brpc/amf.cpp
@@ -63,7 +63,9 @@ AMFField::AMFField(const AMFField& rhs)
     if (rhs.IsString()) {
         if (!rhs._is_shortstr) {
             _str = (char*)malloc(rhs._strsize + 1);
-            memcpy(_str, rhs._str, rhs._strsize + 1);
+            if (_str != NULL) {
+                memcpy(_str, rhs._str, rhs._strsize + 1);
+            }
         }
     } else if (rhs.IsObject()) {
         _obj = new AMFObject(*rhs._obj);
@@ -73,6 +75,7 @@ AMFField::AMFField(const AMFField& rhs)
 }
 
 AMFField& AMFField::operator=(const AMFField& rhs) {
+    assert(this != &rhs);
     Clear();
     _type = rhs._type;
     _is_shortstr = rhs._is_shortstr;
@@ -81,7 +84,9 @@ AMFField& AMFField::operator=(const AMFField& rhs) {
     if (rhs.IsString()) {
         if (!_is_shortstr) {
             _str = (char*)malloc(rhs._strsize + 1);
-            memcpy(_str, rhs._str, rhs._strsize + 1);
+            if (_str != NULL) {
+                memcpy(_str, rhs._str, rhs._strsize + 1);
+            }
         }
     } else if (rhs.IsObject()) {
         _obj = new AMFObject(*rhs._obj);
@@ -145,8 +150,10 @@ void AMFField::SetString(const butil::StringPiece& str) {
         _type = AMF_MARKER_STRING;
         _is_shortstr = true;
         _strsize = str.size();
-        memcpy(_shortstr, str.data(), str.size());
-        _shortstr[str.size()] = '\0';
+        if (_shortstr != NULL) {
+            memcpy(_shortstr, str.data(), str.size());
+            _shortstr[str.size()] = '\0';
+        }
     } else {
         _type = (str.size() < 65536u ?
                  AMF_MARKER_STRING : AMF_MARKER_LONG_STRING);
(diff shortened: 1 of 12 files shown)


src/brpc/amf.cpp
```diff
@@ -63,7 +63,9 @@ AMFField::AMFField(const AMFField& rhs)
     if (rhs.IsString()) {
         if (!rhs._is_shortstr) {
             _str = (char*)malloc(rhs._strsize + 1);
-            memcpy(_str, rhs._str, rhs._strsize + 1);
+            if (_str != NULL) {
+                memcpy(_str, rhs._str, rhs._strsize + 1);
+            }
```
jamesge: When _str is NULL, immediate crash is better and intended

src/brpc/amf.cpp
```diff
@@ -81,7 +84,9 @@ AMFField& AMFField::operator=(const AMFField& rhs) {
     if (rhs.IsString()) {
         if (!_is_shortstr) {
             _str = (char*)malloc(rhs._strsize + 1);
-            memcpy(_str, rhs._str, rhs._strsize + 1);
+            if (_str != NULL) {
```
jamesge: When _str is NULL, immediate crash is better and intended

src/brpc/amf.cpp
```diff
@@ -145,8 +150,10 @@ void AMFField::SetString(const butil::StringPiece& str) {
         _type = AMF_MARKER_STRING;
         _is_shortstr = true;
         _strsize = str.size();
-        memcpy(_shortstr, str.data(), str.size());
-        _shortstr[str.size()] = '\0';
+        if (_shortstr != NULL) {
```
jamesge: _shortstr(array field) will never be NULL

src/brpc/builtin/common.cpp
```diff
@@ -71,6 +71,7 @@ inline bool url_wildcmp(const char* wild, const char* str) {
             ++str;
         } else {
             wild = mp;
+            // Is it guaranteed that cp is not nullptr always here?
```
jamesge: yes. If 'wild' does not contain *, *str should be 0 after the first loop and this while loop should be jumped over; If 'wild'  has a *, cp should be set (to str+1) before here.

src/brpc/input_messenger.cpp
```diff
@@ -328,9 +328,7 @@ void InputMessenger::OnNewMessages(Socket* m) {
         }
     }
 
-    if (read_eof) {
-        m->SetEOF();
-    }
+    m->SetEOF();
```
jamesge: should not be changed.

src/brpc/server.cpp
```diff
@@ -1369,45 +1366,47 @@ void Server::RemoveMethodsOf(google::protobuf::Service* service) {
...
-                }
-                butil::StringPiece v_svc_name(slash_sp.field(), slash_sp.length());
-                const ServiceProperty* vsp = FindServicePropertyByName(v_svc_name);
-                if (vsp == NULL) {
-                    if (_global_restful_map) {
-                        std::string path_str;
-                        path.CopyToString(&path_str);
-                        if (_global_restful_map->RemoveByPathString(path_str)) {
-                            continue;
+        if (mp != NULL) {
```
jamesge: Print fatal and continue is probably better

src/butil/containers/flat_map_inl.h
```diff
@@ -273,7 +273,7 @@ FlatMap<_K, _T, _H, _E, _S>::operator=(const FlatMap<_K, _T, _H, _E, _S>& rhs) {
         free(_buckets);
         _nbucket = rhs._nbucket;
         // note: need an extra bucket to let iterator know where buckets end
-        _buckets = (Bucket*)malloc(sizeof(Bucket) * (_nbucket + 1/*note*/));
+        _buckets = new Bucket[sizeof(Bucket) * (_nbucket + 1/*note*/)];
```
jamesge: _buckets is released by free which should be matched with malloc. And even if new[] is used, the "sizeof(Bucket) *" should be removed.

src/mcpack2pb/serializer-inl.h
```diff
@@ -228,9 +228,9 @@ inline Serializer::GroupInfo* Serializer::push_group_info() {
         return NULL;
     }
     if (_group_info_more == NULL) {
-        _group_info_more =
-            (GroupInfo*)malloc((MAX_DEPTH + 1 - arraysize(_group_info_fast))
-                               * sizeof(GroupInfo));
+        _group_info_more = new GroupInfo[(MAX_DEPTH + 1 - 
+                                arraysize(_group_info_fast))
+                               * sizeof(GroupInfo))];
```
jamesge: _group_info_more is released by free which should be matched with malloc. "sizeof(GroupInfo)" should be removed even if new[] is used.

PR #287, opened 2018-03-28
Update task_group.cpp

leerayven:
Seemed that I found a bug, and I added a comment there.

### DIFF
--- a/src/bthread/task_group.cpp
+++ b/src/bthread/task_group.cpp
@@ -142,7 +142,7 @@ void TaskGroup::run_main_task() {
     std::unique_ptr<bvar::PerSecond<bvar::PassiveStatus<double> > > usage_bvar;
     
     TaskGroup* dummy = this;
-    bthread_t tid;
+    bthread_t tid; //tid is not initialized, and it would be presented to sched_to if wait_task just returned -1. Any problem?
     while (wait_task(&tid)) {
         TaskGroup::sched_to(&dummy, tid);
         DCHECK_EQ(this, dummy);


src/bthread/task_group.cpp
```diff
@@ -142,7 +142,7 @@ void TaskGroup::run_main_task() {
     std::unique_ptr<bvar::PerSecond<bvar::PassiveStatus<double> > > usage_bvar;
     
     TaskGroup* dummy = this;
-    bthread_t tid;
+    bthread_t tid; //tid is not initialized, and it would be presented to sched_to if wait_task just returned -1. Any problem?
```
jamesge: wait_task returns bool.

leerayven: yeah, but if(_last_pl_state.stopped()) returns true, and wait_task returns -1, which is true for the while loop condition, then the uninitialized tid would be presented to sched_to(). Am I right?

jamesge: oh, I have noticed the "return -1" in wait_task(), which is incorrect and should be "return false". It will be fixed. Fortunately the buggy branch is never entered currently since worker threads are not stopped when the program quits (to avoid potential deadlocks).

leerayven: Haha actually I spent much time to convince myself that there should be something wrong with it, since it should be fully tested by numerous projects by baidu.

jamesge:
Fixed in [this commit](https://github.com/brpc/brpc/commit/322b87d993195e70b4d625df058d49c44ab6195c). Thanks for pointing out.

PR #322, opened 2018-05-12
Fix mac build

eric-jj:
1. Fix can't get correct mac version on latest mac OSX, use a common script to do that.
2. Fix build error when disable thrift support. Without --with-thrift parameter, the libbrpc_thrift.a should not be depends on.


zyearn:
This branch conflicts with the latest version. Please merge master and push again. @eric-jj

zyearn:
All the changes have already been fixed in the latest version. Close the pr.

PR #342, opened 2018-05-24
Fix regster thrift protocol issue.

kenshinxf:


src/brpc/policy/thrift_protocol.cpp
```diff
@@ -632,18 +632,20 @@ void PackThriftRequest(
 
 } // namespace policy
 
+extern "C" {
+
 void RegisterThriftProtocol() {
```
jamesge: 移到brpc外面

kenshinxf: done

PR #345, opened 2018-05-25
chore: fix undefined reference to butil::ReadCommandLine

hawkingrei:
When I use bazel to build brpc. it failed to build. so I check the `BUILD` file. i find that it loss `src/butil/process_util.cc`

I have another phenomenon that the Travis is green but bazel building still failed by check the log.


.travis.yml
```diff
@@ -25,4 +25,5 @@ install:
 - sudo apt-get install -y gdb  # install gdb
 
 script:
+- if [[ "$PURPOSE" == "compile-with-bazel" ]]; then bazel build -j 12 -c opt --copt -DHAVE_ZLIB=1 //...          ; fi
```
jamesge: "; fi"前的空格是必须的么

BUILD
```diff
@@ -32,12 +44,16 @@ COPTS = [
...
-    "-lssl",
-    "-lcrypto",
     "-ldl",
     "-lz",
-]
+] + select({
+    ":darwin": [],
+    "//conditions:default": [
+      "-lrt",
+      "-lssl",
```
jamesge: 区分下是编译问题还是链接问题，我记得是include path导致的编译问题。

PR #389, opened 2018-07-06
Add tag support for single connection case

kenshinxf:
Add tag support for single connection case


jamesge:
Tag support in SocketMaps will be added in #482

PR #710, opened 2019-04-01
add NOTICE.txt to address copyright

tanzhongyibidu:


jamesge:
1. 确认下为什么两个文件的后缀名不统一
2. 这两个文件和已有的LICENSE是什么关系

tanzhongyibidu:
Hi， gejun，
 
Let me explain it in more details：
 
1.     The LICENSE file located under the root of repo, should mention its license content and plus all third party license also.
For example , you can check https://github.com/apache/incubator-skywalking/blob/master/LICENSE
You can notice that it includes MIT /BSD license also for its included component.
 
2.     The NOTICE.txt located under the root of repo,  will be used to mention its copyright has belonged to apache.org
Here is one example by doris
https://github.com/apache/incubator-doris/blob/master/NOTICE.txt
 
3.     The Disclaimer file is used to describe its current status is under incubator.
Here is one example
https://github.com/apache/incubator-doris/blob/master/DISCLAIMER
and another one:
https://github.com/apache/incubator-skywalking/blob/master/DISCLAIMER
 
These files will be bunded into the release source tarball,
when user download and uncompress them,  user can easily find the license, copyright owner, and current status.

jamesge:
Thanks for the explanation, however the first question remains: why there's an additional ".txt" extension to "NOTICE"

tanzhongyibidu:
名字叫NOTICE，这是需要的，至于有的叫NOTICE.txt，有的叫NOTICE，都可以。
要不我改成NOTICE？

jamesge:
ok，保持一致吧

tanzhongyibidu:
BTW: I add another file named Disclaimer, when I update the diff, it is merged automatically.

PR #733, opened 2019-04-16
Add a gflag to enable tcp keepalive

### LINKED ISSUE #731: A question about supporting TCP keepalive
In the current implementation, the parameter that sets the `TCP keepalive` is not found. 

The problem is that if the client is abnormally down (such as restarting after power failure), the server may have a large number of semi-connected states, that is, The client side can't see these connections (`netstat -apno`), but it still maintains a large number of `ESTABLISHED` status links on the server side.
At the same time, since linux does not open keepalive by default, if these connections have no data transmission, then it will always exist, and it will consume the resources of the server. And if the server side has some resource restrictions (such as limiting the total number of fd), the system will face the problem of service failure.

Therefore, I feel that Brpc as an rpc library should be able to allow `TCP keepalive` to be manually enabled in the program.

If there is no other special consideration in the Brpc, I can submit a PR

lingbin:
Fix #731 

The gflag name is `socket_server_tcp_keepalive`

TCP keepalive can keeps connections active by enabling the periodic
transmission of messages

Enable this option to help the server release resources early
when the client is disconnected abnormally, such as power down

### DIFF
--- a/src/brpc/socket.cpp
+++ b/src/brpc/socket.cpp
@@ -70,7 +70,7 @@ DEFINE_bool(log_idle_connection_close, false,
             "Print log when an idle connection is closed");
 BRPC_VALIDATE_GFLAG(log_idle_connection_close, PassValidate);
 
-DEFINE_bool(socket_enable_tcp_keepalive, true,
+DEFINE_bool(socket_enable_tcp_keepalive, false,
         "Open tcp keepalive explicitly, if this value is false, "
         "use system default");
 
--- a/src/brpc/socket.cpp
+++ b/src/brpc/socket.cpp
@@ -70,6 +70,10 @@ DEFINE_bool(log_idle_connection_close, false,
             "Print log when an idle connection is closed");
 BRPC_VALIDATE_GFLAG(log_idle_connection_close, PassValidate);
 
+DEFINE_bool(socket_enable_tcp_keepalive, true,
+        "Open tcp keepalive explicitly, if this value is false, "
+        "use system default");
+
 DEFINE_int32(socket_recv_buffer_size, -1, 
             "Set the recv buffer size of socket if this value is positive");
 
@@ -548,6 +552,14 @@ int Socket::ResetFileDescriptor(int fd) {
     if (_tos > 0 &&
         setsockopt(fd, IPPROTO_IP, IP_TOS, &_tos, sizeof(_tos)) < 0) {
         PLOG(FATAL) << "Fail to set tos of fd=" << fd << " to " << _tos;
+    }
+
+    if (FLAGS_socket_enable_tcp_keepalive) {
+        int one = 1;
+        if (setsockopt(fd, SOL_SOCKET, SO_KEEPALIVE, &one, sizeof(one)) != 0) {
+            PLOG(FATAL) << "Fail to set enable keepalive of fd=" << fd;
+        }
+
     }
 
     if (FLAGS_socket_send_buffer_size > 0) {


src/brpc/socket.cpp
```diff
@@ -70,6 +70,10 @@ DEFINE_bool(log_idle_connection_close, false,
             "Print log when an idle connection is closed");
 BRPC_VALIDATE_GFLAG(log_idle_connection_close, PassValidate);
 
+DEFINE_int32(socket_server_tcp_keepalive, true,
```
jamesge: This name is inconsistent with the impl. which works for client as well.

lingbin: Thank you for review.
I will change the name to `socket_enable_tcp_keepalive`

zyearn:
when socket_enable_tcp_keepalive is true, UT is broken

lingbin:
@zyearn  thanks for your review

when the default value of `socket_enable_tcp_keepalive` is true, the error message in some UT is 

> Fail to set enable keepalive of fd=4: **Socket operation on non-socket**. 

So there should be some places where a non-socket fd is used to create the brpc::Socket object (a clue is from the above error, which is the value of fd is 4, indicating that its high probability is not a real socket fd).

Change the default value of `socket_enable_tcp_keepalive` to false, which can maintain the same behavior as the original, just like the existing socket configuration (such as FLAGS_socket_send_buffer_size). That is, if the default value of `FLAGS_socket_send_buffer_size` is greater than 0, then there will be similar problems.

BTW, I am sorry that I am unable to find a more detailed reason for now, because I am currently experiencing some problems during compilation (should be due to my machine environment), resulting in the inability to perform sufficient GDB. After I solved the current compile problem, I am willing to do it.

lingbin:
I found the reason.

The `HttpTest` class (in the brpc_http_rpc_protocol_unittest.cpp file) uses a **pipe fd** to create `brpc::Socket` object, so it will fail when using `setsockopt()`.

The code is as follows:
https://github.com/apache/incubator-brpc/blob/acce77994832f17666324837c448117cf41e5b5d/test/brpc_http_rpc_protocol_unittest.cpp#L98-103

jamesge:
Do we have any application here? The tcp-keepalive may make existing features like -idle_timeout not working as expected, which may confuse users. Without a real application, this PR will not be merged.

lingbin:
Thank @jamesge for your reply.

Let me describe our application. We use *[braft](https://github.com/brpc/braft)* (*brpc* based) to build our program, which will have a lot of raft groups, and RPC is required between the `Leader node` and the `Follower node` of each raft group. When we make some exceptions, such as drop all the network packets of a machine's brpc port, we will find a large number of semi-connections on the server side. 

As @zyearn stated in the[ previous reply](https://github.com/apache/incubator-brpc/issues/731#issuecomment-483601355) of issue #731 , this problem can be solved by setting `ServerOption.idle_timeout_sec` on the server side. However, when the server is abnormally down, the client is still at risk, and this problem seems can only be solved by turning on `Tcp keepalive`.

But after careful thought, I feel that **the risk of the client side may not be as serious as imagined**, because the number of connections made by the client side is controlled by itself, whether single link or connection pool, so the number of socket fd on client side must be a limited number, and will know which link is bad when send next message (in contrast, on the server side, the number of socket fds created depends on the number of connections initiated by the client, so in theory, there may be unlimited socket fd)

Anyway, thank you both for letting me know how to solve our problem.

chenzhangyi:
@lingbin 
As far as I know, braft has resolved this problem by issuing `heartbeats` from `leaders` to `followers` periodically. 
In your case, writing something to the semi connection would trigger RTO(if the server side doesn't receive the package) or RST(if the server side receives the package) and the client side would finally close the connection.
It might lead to a performance issue if adding the `tcp keepalive` flag to a large amount of connections. It's better to do a benchmark test to find out the overhead.

lingbin:
@chenzhangyi 
Yes, as you said, there's no problem with the **client side** in *braft*, because Leader sends heartbeat RPC periodically to Follower. We encountered this problem on the **server side** because we didn't set `ServerOption. idle_timeout_sec`.

What I said before
> However, when the server is abnormally down, **the client is still at risk**, and this problem seems can only be solved by turning on Tcp keepalive. 

It is not to say *braft*, and we did not encounter problems on the **client side** when using *braft*. What I really want to say is that in an application if the client side **does not periodically send RPC** to the **server side**, the client side has this risk, but I really **haven not encountered a specific application example yet**, so I will close this PR for the time being.

Thank you all for the patience reply

PR #748, opened 2019-04-28
Fix a typo in bvar_c++.md

lingbin:


lingbin:
It seems that the current code has a problem, not the PR, because this PR only modifies the markdown doc.

Issue #769, opened 2019-05-16
Compilation errors caused by ssl on MacOS


gavinchou:
**Describe the bug (描述bug)**
It failed to build brpc on MacOS with all dependencies given.

The essence of the bug is that `config_brpc.sh` does not understand the given libopenssl's (libssl) path, result that it failed to link libbrpc.

**To Reproduce (复现方法)**

```
sh config_brpc.sh --headers="`pwd`/dependency/gflags/include `pwd`/dependency/leveldb/include $HOME/installed-from-src/protobuf/protobuf-3.5.2/include `pwd`/dependency/glog/include $HOME/installed-from-src/openssl/openssl-1.1.1/include" --libs="`pwd`/dependency/gflags/libs `pwd`/dependency/glog/libs `pwd`/dependency/leveldb/libs $HOME/installed-from-src/protobuf/protobuf-3.5.2/lib  $HOME/installed-from-src/openssl/openssl-1.1.1/lib" --cc=clang --cxx=clang++
```
where `pwd` is the root of repo brpc.

**Expected behavior (期望行为)**
build pass

**Versions (各种版本)**
OS: MacOS 10.13.6
Compiler: Apple LLVM version 9.1.0 (clang-902.0.39.2)
brpc: 7e85374b
protobuf: 3.5.2

**Additional context/screenshots (更多上下文/截图)**

![image](https://user-images.githubusercontent.com/6919662/57828096-c7cb2300-77dc-11e9-8689-22f3a10c0791.png)

**Related issues or pull request**:    
* MacOS安装make报错 #328
* macos 10.12.6 make出错 #360
* Mac下链接生成动态库失败 #666
* mac 10.13.6 编译出错 #726
* fix build scripts under mac os env and add building section for README #723

and I have tried to fix it with, a pull request will be posted later.

gavinchou:
Corresponding  pull request #770

Issue #726, opened 2019-04-12
mac 10.13.6 编译出错


xuexiangjys:
**Describe the bug (描述bug)**

ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [libbrpc.dylib] Error 1

**To Reproduce (复现方法)**
$ sh config_brpc.sh --headers=/usr/local/include --libs=/usr/local/lib --cc=clang --cxx=clang++
$ make

**Expected behavior (期望行为)**
编译通过

**Versions (各种版本)**
OS: mac 10.13.6
Compiler:
brpc:
protobuf:3.6.1

**Additional context/screenshots (更多上下文/截图)**
Packing libbrpc.a
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libbrpc.a(atomicops_internals_x86_gcc.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libbrpc.a(asan_invalid_access.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libbrpc.a(thread_restrictions.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libbrpc.a(mesalink_ssl_helper.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libbrpc.a(baidu_naming_service.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libbrpc.a(giano_authenticator.o) has no symbols
Linking protoc-gen-mcpack
ld: warning: text-based stub file /System/Library/Frameworks//CoreFoundation.framework/CoreFoundation.tbd and library file /System/Library/Frameworks//CoreFoundation.framework/CoreFoundation are out of sync. Falling back to library file for linking.
[...]

Issue #666, opened 2019-02-22
Mac下链接生成动态库失败


fengqibin:
Mac编译正确，链接生成静态库也没问题，但是链接生成动态库时出现报错，查阅了一些资料也没解决，菜鸟一枚，只能在此请教了，报错如下，望大神解惑：
Linking libbrpc.dylib
Undefined symbols for architecture x86_64:
  "_CRYPTO_THREADID_set_callback", referenced from:
      brpc::SSLThreadInit() in ssl_helper.o
  "_CRYPTO_THREADID_set_numeric", referenced from:
      brpc::SSLGetThreadId(crypto_threadid_st*) in ssl_helper.o
  "_EVP_PKEY_base_id", referenced from:
      brpc::SSLGetDHCallback(ssl_st*, int, int) in ssl_helper.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [libbrpc.dylib] Error 1

Issue #360, opened 2018-06-07
macos 10.12.6 make出错


houjincheng1992:
编译的出错的信息如下：
Compiling src/brpc/policy/streaming_rpc_protocol.o
Compiling src/brpc/policy/ubrpc2pb_protocol.o
Compiling src/brpc/policy/weighted_round_robin_load_balancer.o
Packing libbrpc.a
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libbrpc.a(atomicops_internals_x86_gcc.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libbrpc.a(asan_invalid_access.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libbrpc.a(thread_restrictions.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libbrpc.a(baidu_naming_service.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libbrpc.a(giano_authenticator.o) has no symbols
Linking protoc-gen-mcpack
Linking libbrpc.dylib
Undefined symbols for architecture x86_64:
  "_CRYPTO_THREADID_set_callback", referenced from:
      brpc::SSLThreadInit() in ssl_helper.o
  "_CRYPTO_THREADID_set_numeric", referenced from:
      brpc::SSLGetThreadId(crypto_threadid_st*) in ssl_helper.o
  "_EVP_PKEY_base_id", referenced from:
      brpc::SSLGetDHCallback(ssl_st*, int, int) in ssl_helper.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [libbrpc.dylib] Error 1
麻烦相关人帮忙看看原因～～

zyearn:
brew装一下openssl，然后手动软连接到系统lib

houjincheng1992:
@zyearn hi,我是在使用make时候出现的这个问题，是因为我没有安装openssl吗?

houjincheng1992:
[...]

PR #770, opened 2019-05-16
Fix compilation errors caused by ssl on MacOS

gavinchou:
Bug report: #769 

Related issue: #328 #360 #666 #723 #726

According to the [instruction](https://github.com/apache/incubator-brpc/blob/master/docs/cn/getting_started.md)
given by incubator-brpc, openssl seems to be an optional lib but it is required
on all platforms, despite that openssl may be installed by default and no need
to install manually. 

However, openssl is not trusted by Apple due to [heartbleed](https://en.m.wikipedia.org/wiki/Heartbleed),
if we need to get openssl support, we need to install it manually.

We can compile/install leveldb, gflags, openssl, openssl, protobuf on mac by
simply using the package manager called `brew` as suggested by the brpc
installation instruction, or install by hand with `autotools + libtool + pkg-config` 
from source code published on github (or somewhere).

`brew` does a lot for users, however, users usually don't know what's under
the hood and the pkg installation path may differ for different users, which
may make compilation fail.

e.g fails to link openssl when we build brpc on mac

```
Linking libbrpc.dylib
Linking protoc-gen-mcpack
Copying to output/include
Undefined symbols for architecture x86_64:
  "_BN_get_rfc2409_prime_1024", referenced from:
      brpc::SSLDHInit() in ssl_helper.o
      brpc::policy::DHWrapper::do_initialize() in dh.o
  "_BN_get_rfc3526_prime_2048", referenced from:
      brpc::SSLDHInit() in ssl_helper.o
  "_BN_get_rfc3526_prime_4096", referenced from:
      brpc::SSLDHInit() in ssl_helper.o
  "_BN_get_rfc3526_prime_8192", referenced from:
```

To fix that, **we should treat libopenssl as a normal, not default, library**
needed by brpc, just the same as leveldb or gflags.

```
--- a/config_brpc.sh
+++ b/config_brpc.sh
@@ -136,6 +136,9 @@ find_dir_of_header_or_die() {
     $ECHO $dir
 }

+# User specified path of lib openssl, if not given it's empty
+OPENSSL_LIB=$(find_dir_of_lib ssl)
+
 # Inconvenient to check these headers in baidu-internal
 #PTHREAD_HDR=$(find_dir_of_header_or_die pthread.h)
 OPENSSL_HDR=$(find_dir_of_header_or_die openssl/ssl.h)
@@ -233,7 +236,7 @@ PROTOBUF_HDR=$(find_dir_of_header_or_die google/protobuf/message.h)
 LEVELDB_HDR=$(find_dir_of_header_or_die leveldb/db.h)

 HDRS=$($ECHO "$GFLAGS_HDR\n$PROTOBUF_HDR\n$LEVELDB_HDR\n$OPENSSL_HDR" | sort | uniq)
-LIBS=$($ECHO "$GFLAGS_LIB\n$PROTOBUF_LIB\n$LEVELDB_LIB\n$SNAPPY_LIB" | sort | uniq)
+LIBS=$($ECHO "$GFLAGS_LIB\n$PROTOBUF_LIB\n$LEVELDB_LIB\n$OPENSSL_LIB\n$SNAPPY_LIB" | sort | uniq)
```

With this patch, 
users can specify libopenssl when run  `config_brpc.sh` with option
`--libs="... ${path_to_libopenssl} ..."`, in which `path_to_libopenssl` may be
one of `/usr/local/lib`, `/usr/local/opt/lib`,
`/users/gavin/installed-from-src/openssl/lib` and etc.

If you failed to compile due to openssl, try to change the above lib path.

I think it's the purpose of option `--libs`. The original `config_brpc.sh`
treats openssl as a default library instead of a normal library, that may be not
correct.

The above patch should fix all the following issues or pull request:  
* MacOS安装make报错 #328
* macos 10.12.6 make出错 #360
* Mac下链接生成动态库失败 #666
* mac 10.13.6 编译出错 #726
* fix build scripts under mac os env and add building section for README #723

(There may be more issues which I didn't find.)

This is an example of compilation error caused by openssl when we build brpc, 
we can do the same to resolve other compilation error related to libs, including
version issues of libs.

### DIFF
--- a/config_brpc.sh
+++ b/config_brpc.sh
@@ -136,6 +136,9 @@ find_dir_of_header_or_die() {
     $ECHO $dir
 }
 
+# User specified path of openssl, if not given it's empty
+OPENSSL_LIB=$(find_dir_of_lib ssl)
+
 # Inconvenient to check these headers in baidu-internal
 #PTHREAD_HDR=$(find_dir_of_header_or_die pthread.h)
 OPENSSL_HDR=$(find_dir_of_header_or_die openssl/ssl.h)
@@ -233,7 +236,7 @@ PROTOBUF_HDR=$(find_dir_of_header_or_die google/protobuf/message.h)
 LEVELDB_HDR=$(find_dir_of_header_or_die leveldb/db.h)
 
 HDRS=$($ECHO "$GFLAGS_HDR\n$PROTOBUF_HDR\n$LEVELDB_HDR\n$OPENSSL_HDR" | sort | uniq)
-LIBS=$($ECHO "$GFLAGS_LIB\n$PROTOBUF_LIB\n$LEVELDB_LIB\n$SNAPPY_LIB" | sort | uniq)
+LIBS=$($ECHO "$GFLAGS_LIB\n$PROTOBUF_LIB\n$LEVELDB_LIB\n$OPENSSL_LIB\n$SNAPPY_LIB" | sort | uniq)
 
 absent_in_the_list() {
     TMP=`$ECHO "$1\n$2" | sort | uniq`


jamesge:
Thanks for the patch and detailed explanations :)

JasonWpy:
works like this:

cmake -DCMAKE_INSTALL_PREFIX=$INSTALL -DCMAKE_INCLUDE_PATH="../../third_part_openssl/include/;../../third_part_gflags/include/;../../third_part_protobuf/include/;../../third_part_leveldb/include/" -DCMAKE_LIBRARY_PATH="../../third_part_openssl/lib/;../../third_part_gflags/lib/;../../third_part_protobuf/lib/;../../third_part_leveldb/lib64/" .

PR #784, opened 2019-05-28
Improve cmake: make unit tests link against the brpc shared library ...

jasonszang:
... so that we don't build a world of huge binaries that are forced to take every single piece of object files from those object library targets.
Test binaries down size from somewhat 6GB to around 600M. Linking is also vastly speeded up.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -391,6 +391,7 @@ set(SOURCES
 
 add_subdirectory(src)
 if(BUILD_UNIT_TESTS)
+    enable_testing()
     add_subdirectory(test)
 endif()
 add_subdirectory(tools)
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -174,9 +174,9 @@ endif()
 
 # create executable
 add_executable(test_butil ${TEST_BUTIL_SOURCES}
-                          ${CMAKE_CURRENT_BINARY_DIR}/iobuf.pb.cc
-                          $<TARGET_OBJECTS:BUTIL_LIB>)
-target_link_libraries(test_butil gtest
+                          ${CMAKE_CURRENT_BINARY_DIR}/iobuf.pb.cc)
+target_link_libraries(test_butil brpc-shared
+                                 gtest
                                  ${GPERFTOOLS_LIBRARIES}
                                  ${DYNAMIC_LIB})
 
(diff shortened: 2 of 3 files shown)


zyearn:
Hi, the cxxflags used in shared_lib and test are different, please make sure that UTs generated by cmake can be passed. They are broken in my local environment. Since UTs are not run in cmake environment, so travis looks fine.

jasonszang:
Ah, I see you require that some different cxxflags be used to compile non-test code for your unit tests to properly function.
Personally I feel that this might not be a good choice since you are testing a piece of object that is different than what you actually ship with `make install`. Which means that you did not test what you ship.
But anyway I'll try to fix this and make it work for now.

jamesge:
UT may turn on additional debugging macros to check consistencies and would like the release build being unaffected. Creating different (and probably incompatible) binaries is the only solution.

jasonszang:
If creating binaries that have different logic and behave differently for UT cannot be avoided then perhaps we should build a separate brpc shared or static library for testing instead of using object libraries directly for binaries.
And still, the unability to test release library make me a little nervious :) 
Maybe we can solve this by separate tests that depend on these UT specific behaviours to different test binaries, link those against the specially built UT library, and run other normal tests against our release lib?

zyearn:
The logic in UT and release is not strict different but has a relation of Inclusion. It means that if UT is passed, then we can be sure that the code in release is correct. For example, in http2, in order to test the code work expectedly after stream id runs out, we start it at a very big number so that after only a short time, we can test whether the code is correct or not.

jasonszang:
Then you need to make sure this relation of inclusion is always true by writing good tests always. Might be a burden upon ut writers? But if you guys are confident with that then maybe its OK.

jasonszang:
So I for now give up on these tests that relies on non-test objects built with specific cxxflags and instead only change those who do not to directly link against the release library.
Test binaries downsized to 1.2G instead of 600M but its still a big improvement.
As a byproduct, ctest now works correctly and users can now `make test` and run various `ctest` commands with CMake builds.

jasonszang:
Again travis doesn't look right. The error is about `run_tests.sh` not finding `[[`, which I did not touch.
Would anyone please take a look?

jasonszang:
Same error after rebasing, `./run_tests.sh: 22: ./run_tests.sh: [[: not found`
Took a look around and it seems that master branch is currently failing for the same reason. @jamesge

jasonszang:
Looks good now

PR #824, opened 2019-06-27
fix lalb be disabled wrongly

zyearn:


src/brpc/policy/locality_aware_load_balancer.h
```diff
@@ -86,7 +86,7 @@ class LocalityAwareLoadBalancer : public LoadBalancer {
         void Describe(std::ostream& os, int64_t now);
 
         int64_t Disable();
-        bool Disabled() const { return _base_weight == 0; }
+        bool Disabled() const { return _base_weight == -1; }
```
jamesge: <0

zyearn: done

MR-workaholic:
想问问，这个提交是修复哪个问题的？我们在用LA的时候，遇到选服务失败的情况，是因为这个问题吗？

zyearn:
This pr fixed the issue that _base_weight is accidentally set to 0 and Disable() returns true since _base_weight is equal to 0. As a result, LALB would skip the server because it is 'disabled' which should not. If you run across selecting failure and use brpc before Jun, it may probably be caused by this issue.

MR-workaholic:
@zyearn 哪具体哪些情况会出现"_base_weight is accidentally set to 0"？看代码，下游服务出现超长延时或者每个服务的平均weight小于1才会出现这种情况。而且“每个服务的平均weight小于1”会出现吗？

PR #912, opened 2019-08-30
modify code: src/brpc/ssl_options.h annotation ciper to cipher

### LINKED ISSUE #911: ssl_options.h文件：注释中单词拼写错误
文件 src/brpc/ssl_options.h

第66行：// The format of this string should follow that in `man 1 cipers'.
cipers -> ciphers

第139行：// should follow that in `man 1 cipers'. If empty, OpenSSL will choose
cipers -> ciphers

JayTange:
fix #911


jamesge:
thanks for the corrections~

PR #939, opened 2019-10-11
remove core before running test

zyearn:


.travis.yml
```diff
@@ -17,6 +17,7 @@ env:
 
 before_script:
 - ulimit -c unlimited -S       # enable core dumps
+- sudo bash -c "sudo echo 'core.%e.%p' > /proc/sys/kernel/core_pattern"
```
zyearn: The real command should be put into another "bash -c" otherwise it would cause a permission issue. The reason may be  that the 'sudo' only promote the 'echo' part but not the whole command.

PR #953, opened 2019-11-01
Remove conflict typedefs with XCode 11.1.

guoyuhong:
This may resolve the problem of https://github.com/apache/incubator-brpc/issues/952 .


guoyuhong:
It is strange, that this test does not fail in my Mac.

jamesge:
Not seeing the code in [master](https://github.com/apache/incubator-brpc/blob/master/src/butil/mac/foundation_util.h#L53), closing this pr. thank you 😊

PR #966, opened 2019-11-13
Fix thread_local in clang

zyctree:
when building with **clang** and c++11, the c++11 check with `__GNUC__` and `__GNUC_MINOR__` macro will not pass, and there will be a macro `# define thread_local __thread`, which is not expected

### DIFF
--- a/src/butil/thread_local.h
+++ b/src/butil/thread_local.h
@@ -28,9 +28,7 @@
 // Provide thread_local keyword (for primitive types) before C++11
 // DEPRECATED: define this keyword before C++11 might make the variable ABI
 // incompatible between C++11 and C++03
-#if !defined(thread_local) &&                                           \
-    (__cplusplus < 201103L ||                                           \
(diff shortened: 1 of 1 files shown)


jamesge:
what's the version of your clang?

zyctree:
clang-6.0
a simple test
```
$ clang-6.0 -dM -E - < /dev/null  | grep GNU
#define __GNUC_MINOR__ 2
#define __GNUC_PATCHLEVEL__ 1
#define __GNUC_STDC_INLINE__ 1
#define __GNUC__ 4
```

jamesge:
Our travis-ci on clang seems working. Anyway, the definition of "thread_local" is already useless in this repo which turns on C++11 by default. The deprecated code will be removed soon, thanks for your contribution.

jamesge:
https://github.com/apache/incubator-brpc/commit/6efb0cff5f30f32437c660cef01c93549cf62679

PR #997, opened 2019-12-14
Http empty uri should check whether user registered restful_mappings, instead of directly use IndexServer.

joshuafc:
Http empty uri should check whether user registered restful_mappings, instead of directly use IndexServer.


joshuafc:
So we can 
`server.AddService(&srv, brpc::SERVER_DOESNT_OWN_SERVICE, "/   => Do2,  /test   => Do2 " );`

chenjianyu:
I met the same issue too

Issue #1007, opened 2019-12-30
Hash collision degrades the performance of SocketMap


pexeer:
**Describe the bug (描述bug)**
send lots of RPC to many server(with same port) will cause high cpu and bad latency. 

**To Reproduce (复现方法)**
1. use channel with `single` options and  set defer_close_second to 60 seconds.
2. send every RPC with a new channel object and deconstruct after used. 
3. startup a large number of server with different ip and same port.
4. startup a BRPC client and use some thread to send large number of RPC to different server at the same time.
5. high cpu usage and very high RPC latency in the client.

**Expected behavior (期望行为)**
Client latency is approximately the same as server-side latency， and normal cpu usage.

**Additional context/screenshots (更多上下文/截图)**

The main point maybe is the Hash function of EndPoint class.
The hash function `butil::HashPair(uint32_t ip,uint32_t port)` just  connect two uint32_t to  a uint64_t.
FlatMap used power rehash policy but not the prime rehash policy to speed up `%` operation.
So only low address of the hash value  will be used by FlatMap.
Total EndPoint with same port will linked by linked list in the same bucket of FlatMap.
Lookup EndPoint in the SocketMap is O(n) time complexity guard by a pthread mutex.
All Channel init will wait the mutex of SocketMap.

pexeer:
#1008 fixed

PR #1009, opened 2019-12-30
bugfix to the hash function HashInts32.

pexeer:
the another optional solution of Issue #1007
make right of the hash function of HashInts32.

### DIFF
--- a/src/butil/containers/hash_tables.h
+++ b/src/butil/containers/hash_tables.h
@@ -26,6 +26,7 @@
 #include "butil/basictypes.h"
 #include "butil/strings/string16.h"
 #include "butil/build_config.h"
+#include "butil/third_party/murmurhash3/murmurhash3.h"   // fmix64
 
 #if defined(COMPILER_MSVC)
 #include <hash_map>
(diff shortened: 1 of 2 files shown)


src/butil/containers/hash_tables.h
```diff
@@ -129,8 +129,10 @@ using BUTIL_HASH_NAMESPACE::hash_set;
 //
 // Contact danakj@chromium.org for any questions.
 inline std::size_t HashInts32(uint32_t value1, uint32_t value2) {
-  uint64_t value1_64 = value1;
-  uint64_t hash64 = (value1_64 << 32) | value2;
+  uint64_t short_random1 = 842304669U;
+  uint64_t short_random2 = 619063811U;
+
+  uint64_t hash64 =  short_random1 * value1 + short_random2 * value2;
```
jamesge: please use [fmix64](https://github.com/apache/incubator-brpc/blob/fc07b7bf45c4b2aa90ae5c832f3b48a8da124640/src/butil/third_party/murmurhash3/murmurhash3.h#L57)((value1_64 << 32) | value2) which is better studied.

pexeer: good idea!

PR #1012, opened 2020-01-02
bugfix: pid file may be overwritten

eric-buaa-cn:
The following step will mess up the pid file:
1. Start the server with pid file enabled. Suppose the pid is 12345.
2. Make the server core dump. 
3. Restart the server. Suppose the new pid is 8888. The pid file will be overwritten with the wrong content 88885.


jamesge:
Thanks for your fix 👍 , and I've checked other usages of `open`, which seem to be OK.

PR #1028, opened 2020-01-19
support unix socket

Suse130224:


jamesge:
Thanks for your contribution. However, the support for unix sockets are better encapsulated in EndPoint without changing too much other code, especially the Channel. Another limit is that EndPoints are widely used and adding a char array inside it is unacceptable. Since valid port is between [0, 65535], it's possible that we put a pointer inside EndPoint without changing the structure size.

PR #1032, opened 2020-01-19
replace_include_with_find_package

zyearn:


zyearn:
according to https://github.com/apache/incubator-brpc/issues/1023, find_package provide more options for user to set(such as CMAKE_PREFIX_PATH). The minimum cmake version that brpc use is [2.8.10](https://cmake.org/cmake/help/v2.8.10/cmake.html#command:find_package), and find_package is supported in this version. It is safe to replace include with find_package.

zyearn:
cmake/FindGFLAGS.cmake and cmake/FindGperftools.cmake can not be deleted because they are not builtin module of cmake, we still need it if user install them using other method, such as apt-get

PR #1058, opened 2020-03-06
pprof: load symbol when init PProfService

lorinlee:
We guess it is likely that there is something wrong with this lazy loading symbol map. When process's memory is too large, this operation, which forks 'nm' subprocess several times, will copy a large page table several times. Copy page table holds mm.mmap_sem, which blocks mmap() and brk().

### DIFF
--- a/src/brpc/builtin/pprof_service.cpp
+++ b/src/brpc/builtin/pprof_service.cpp
@@ -517,6 +517,11 @@ static void FindSymbols(butil::IOBuf* out, std::vector<uintptr_t>& addr_list) {
     }
 }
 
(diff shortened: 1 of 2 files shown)


jamesge:
This mod adds a delay for all servers which may not need the profiler, please time the LoadSymbols function for several regularly-sized binaries(100MB~200MB) to see if the delay is acceptable.

lorinlee:
Is it possible to parse the elf file's symbols, instead of using 'nm'? I think it can both load symbols lazily and no need to fork a subprocess. But I'm not sure if there are some problems like platform compatibility or something else.

PR #1062, opened 2020-03-13
Support to compile in mysql by changing LOG_NONE to LOG_TO_NONE

GOGOYAO:
This pr is for issue #1061.

LOG_NONE is a enum value but it is a macro definition in mysql. This will cause compile error. 

We can solve this problem by changing LOG_NONE to LOG_TO_NONE.


jamesge:
Thanks for the PR. The fix is already done in https://github.com/apache/incubator-brpc/commit/e3840c18cdd9e1ff81f8280b96ac14869007a122

PR #1121, opened 2020-05-22
`THRIFT_STDCXX` use std by default when _THRIFT_STDCXX_H_ not define

curoky:
Thrift [dropped support for support for C++03/C++98](https://thrift.apache.org/lib/cpp#100) and removed `lib/cpp/src/thrift/stdcxx.h` in [0.13.0](https://github.com/apache/thrift/commit/316723add4c368ffd144dd5beb55245832e073fa#diff-d590dd422756aa5d86fbb57a1cdb386d) which defined `_THRIFT_STDCXX_H_`. 

When build brpc with thrift-0.13.0 or a newer version, there will be a compilation error. Because brpc use `boost::shared_ptr` but thrift use `std::shared_ptr`.

Therefore we should use `std` by default when `_THRIFT_STDCXX_H_` is not defined.

### DIFF
--- a/src/brpc/policy/thrift_protocol.cpp
+++ b/src/brpc/policy/thrift_protocol.cpp
@@ -45,8 +45,7 @@
  #if defined(_THRIFT_STDCXX_H_)
  # define THRIFT_STDCXX apache::thrift::stdcxx
  #else
- # define THRIFT_STDCXX boost
- # include <boost/make_shared.hpp>
+ # define THRIFT_STDCXX std
  #endif
 #endif
 


lorinlee:
what if someone uses thrift whose version less than 0.13.0?

curoky:
> what if someone uses thrift whose version less than 0.13.0?

Thrift add `_THRIFT_STDCXX_H_` after `0.10.0`, remove it from `0.13.0`.

This means that users whose version <= `0.10.0` will fail to compile, but if we don't change it, all users whose version >= `0.13.0` will fail to compile.

Maybe there is a better way to solve this problem which not need to change default behavior. For example, we can determine whether file `stdcxx.h` exists in cmake or bazel.

But no matter what, we should make a change

PR #1142, opened 2020-06-16
Fix not to abort when checking the errorno with unicode string

tobegit3hub:
Do not abort when comparing the errorno message with the string and just print the warning.

Resolved https://github.com/apache/incubator-brpc/issues/863 .


tobegit3hub:
It is a quick fix. Can @zyearn @jamesge @TousakaRin help to review and merge this?

PR #1212, opened 2020-08-21
Enable using brpc to implement an Mysql-compatible server as quickly as possible.

eric-buaa-cn:


eric-buaa-cn:
May be later I'll fix the license problem.

PR #1247, opened 2020-09-18
update StreamOptions

cstarc:
update StreamOptions


jamesge:
Thanks for spoting the inconsistency

PR #1270, opened 2020-10-20
coding style fix in brpc/trackme.cpp

gydong:
1. change the prefix of static global variables from "g_" to "s_" in brpc/trackme.cpp.  
2. delete outdated comments in src/bthread/task_group.cpp.  
3. all other changes  are the trailing spaces.


src/brpc/trackme.cpp
```diff
@@ -43,11 +43,11 @@ static const int32_t TRACKME_MIN_INTERVAL = 30;
 static const int32_t TRACKME_MAX_INTERVAL = 600;
 static int32_t s_trackme_interval = TRACKME_MIN_INTERVAL;
```
serverglen, line 44: s_traceme_interval => g_traceme_interval?

gydong: "s" stands for "static"

PR #1278, opened 2020-10-26
fix coredump cause by bad growth_non_responsive http request

acelyc111:
Url from web browser request like 'host:port/hotspots/growth_non_responsive?console=abc' will cause coredump.
ref https://github.com/apache/incubator-brpc/issues/1112


acelyc111:
@jamesge 这是个bug吗，还是说正常情况下不应该出现？

PR #1279, opened 2020-10-27
Small alignment in examples

darionyaphet:


example/echo_c++/client.cpp
```diff
@@ -76,10 +76,10 @@ int main(int argc, char* argv[]) {
         stub.Echo(&cntl, &request, &response, NULL);
         if (!cntl.Failed()) {
             LOG(INFO) << "Received response from " << cntl.remote_side()
-                << " to " << cntl.local_side()
```
gydong, line 79: Just keep the same indent style as that in other soruce code files？

darionyaphet: yep There is no other meaning.

gydong: I mean it seems here the coding style was the same with other files. Why not just keeping what it was

PR #1297, opened 2020-11-28
fix ErrorToString

darionyaphet:


src/butil/files/file.cc
```diff
@@ -117,7 +117,9 @@ std::string File::ErrorToString(Error error) {
     case FILE_ERROR_IO:
       return "FILE_ERROR_IO";
     case FILE_ERROR_MAX:
-      break;
```
gydong, line 120: FILE_ERROR_MAX is not a specific error, it's just a upper limit for that enum.

PR #1308, opened 2020-12-21
Correct a spelling mistake

lishoujun:
correct a spelling mistake


src/brpc/controller.h
```diff
@@ -340,7 +340,7 @@ friend void policy::ProcessThriftRequest(InputMessageBase*);
     // and controller.cpp to know more.
     void allow_done_to_run_in_place()
     { add_flag(FLAGS_ALLOW_DONE_TO_RUN_IN_PLACE); }
-    // True iff above method was called.
```
wasphin, line 343: This is not a typo, ``iff`` means if and only if

lishoujun: O, I see.

PR #1313, opened 2020-12-25
fix bug: "bthread switch" result in "sched_to itself!"

zzjshaok:
sometimes it output: LOG(FATAL) << "bthread=" << g->current_tid() << " sched_to itself!";
it dosen't work as expected.


wwbmmm:
There is a better solution: https://github.com/apache/incubator-brpc/pull/1777

PR #1320, opened 2020-12-29
Fix registering multiple addresses to discovery

wasphin:
See API here at https://github.com/bilibili/discovery/blob/master/doc/api.md#%E6%B3%A8%E5%86%8Cregister

### DIFF
--- a/src/brpc/policy/discovery_naming_service.cpp
+++ b/src/brpc/policy/discovery_naming_service.cpp
@@ -277,7 +277,9 @@ int DiscoveryClient::DoRegister() {
     std::vector<butil::StringPiece> addrs;
     butil::SplitString(_params.addrs, ',', &addrs);
     for (size_t i = 0; i < addrs.size(); ++i) {
-        os << "&addrs=" << addrs[i];
+        if (!addrs[i].empty()) {
+            os << "&addrs=" << addrs[i];
+        }
     }
 
     os << "&env=" << _params.env
--- a/src/brpc/policy/discovery_naming_service.cpp
+++ b/src/brpc/policy/discovery_naming_service.cpp
@@ -21,6 +21,7 @@
 #include "butil/third_party/rapidjson/memorybuffer.h"
 #include "butil/third_party/rapidjson/writer.h"
 #include "butil/string_printf.h"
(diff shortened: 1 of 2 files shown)


gydong:
LGTM

zyearn:
Could you add the corresponding Unit Test of this case?

src/brpc/policy/discovery_naming_service.cpp
```diff
@@ -271,9 +272,15 @@ int DiscoveryClient::DoRegister() {
...
-        << "&hostname=" << _params.hostname
-        << "&addrs=" << _params.addrs
-        << "&env=" << _params.env
+        << "&hostname=" << _params.hostname;
+
+    std::vector<butil::StringPiece> addrs;
+    butil::SplitString(_params.addrs, ',', &addrs);
+    for (size_t i = 0; i < addrs.size(); ++i) {
+        os << "&addrs=" << addrs[i];
+    }
```
zyearn, line 281: According to the comment of butil::SplitString: If several instances of |c| are contiguous, or if |str| begins with or ends with |c|, then an empty string is inserted, we should check if addrs[i] is empty string.

test/brpc_naming_service_unittest.cpp
```diff
@@ -579,6 +579,14 @@ class DiscoveryNamingServiceImpl : public test::DiscoveryNamingService {
                  google::protobuf::Closure* done) {
         brpc::ClosureGuard done_guard(done);
         brpc::Controller* cntl = static_cast<brpc::Controller*>(cntl_base);
+        auto body = cntl->request_attachment().to_string();
+        for (brpc::QuerySplitter sp(body); sp; ++sp) {
+            if (!sp.key().empty()) {
```
zyearn, line 584: I would suggest this line can be deleted.

wasphin: Thanks for the review, and done.

src/brpc/policy/discovery_naming_service.cpp
wasphin: Thanks for the review, and done.

zyearn:
Thanks for the contribution.

PR #1358, opened 2021-03-16
Fix deadlock in TaskGroup::push_rq

SF-Zhou:
This is a known bug in src/bthread/task_group_inl.h. It will cause a deadlock when all workers are spinning in `TaskGroup::push_rq`. A reproducible code:
```c++
#include <atomic>
#include <thread>
#include <vector>

#include <bthread/bthread.h>

static std::atomic<size_t> value{0};

void* do_job(void*) {
  ++value;
  return nullptr;
}

void* start(void*) {
  const uint32_t M = 10000;  // bigger than default queue capacity
  std::vector<bthread_t> tids(M);
  for (auto& tid : tids) {
    bthread_start_background(&tid, &BTHREAD_ATTR_SMALL, do_job, nullptr);
  }
  for (auto& tid : tids) {
    bthread_join(tid, nullptr);
  }
  return nullptr;
}

int main(int argc, char* argv[]) {
  const uint32_t N = 10;  // bigger than default concurrency

  std::vector<bthread_t> tids(N);
  for (auto& tid : tids) {
    bthread_start_urgent(&tid, &BTHREAD_ATTR_SMALL, start, nullptr);
  }
  for (auto& tid : tids) {
    bthread_join(tid, nullptr);
  }
  printf("Value = %lu\n", value.load());
  return 0;
}
```

### DIFF
--- a/src/bthread/task_group_inl.h
+++ b/src/bthread/task_group_inl.h
@@ -93,7 +93,8 @@ inline void TaskGroup::push_rq(bthread_t tid) {
         // A better solution is to pop and run existing bthreads, however which
         // make set_remained()-callbacks do context switches and need extensive
         // reviews on related code.
-        ::usleep(1000);
+        TaskGroup* g = this;
+        TaskGroup::yield(&g);
     }
 }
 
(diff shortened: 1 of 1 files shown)


jamesge:
This PR does not solve the issue, the removed comment clearly explains the details.

PR #1359, opened 2021-03-17
Fix possible memory leak in socket.cpp

SF-Zhou:
When `bthread_start_background` fails to start, the closure `thrd_func` is not executed and released, which will cause a memory leak.
Fix: manually release when fail.


TousakaRin:
Maybe it is better to use an unique_ptr?

Issue #1401, opened 2021-05-12
Multi rdma device not working in rdma brpc branch


tbago:
We have multi rdma NIC in server. If we use the first rdma device by use **--rdma_device=mlx5_0**. The brpc client and server works well. But if we use other rdma device like **--rdma_device=mlx5_1**, the client cannot communicate to the server. It will show the following error:
<img width="827" alt="WechatIMG24" src="https://user-images.githubusercontent.com/12290248/117993016-efc0d300-b371-11eb-95f3-f8a3d18d4c20.png">

The [rdma_create_qp](https://github.com/apache/incubator-brpc/blob/rdma/src/brpc/rdma/rdma_communication_manager.cpp#L311)  will failed with errno=22 (invalid param). The root cause is that the GetRdmaProtectionDomain() is use the global context, I mean that the global context may support support well or have some bugs? I have write my custom client and use rdma_cm_id->verbs as context. Then I can connect the brpc server with other rdma device.
  But in the [rdma_helper.cpp](https://github.com/apache/incubator-brpc/blob/rdma/src/brpc/rdma/rdma_helper.cpp) the pd is init before the rdma_cm_id. 
    So my question is does someone have the same issue will me.  And how to fixed the bug, do we need the create the pd after the rdma_cm_id is created.

tbago:
I found the rdma_cm_id is not matched when use the second device. I have add the bind function in client side. And it can fixed the client connection isuse.
<img width="829" alt="WechatIMG27" src="https://user-images.githubusercontent.com/12290248/118115593-bd6bb000-b41b-11eb-90c1-b161939d237c.png">

tbago:
But in the brpc server side. The rdma device options still not working.

changchengx:
@ziruiliu @tbago 
Does brpc support RoCE LAG device (two RDMA NICs are bond into one logic port)?

wwbmmm:
brpc has no plan to support multi rdma device.

changchengx:
@wwbmmm 
I tested the RoCE LAG device with below command and it could work
```
[307 /brpc/example/rdma_performance/]$ ./perf_server --rdma_device mlx5_bond_1
[306 /brpc/example/rdma_performance/]$ ./perf_client --rdma_device mlx5_bond_1 --servers=192.168.30.7:8002 --attachment_size=1024 --thread_num=10
```
[...]

PR #1402, opened 2021-05-13
Fixed rdma multi device connect failed issue

tbago:
The root cause is that if you specified the rdma_device in GFLAGS, the rdma_helper will alloc the pd in the specified device. But in [RdmaCommunicationManager::Create()](https://github.com/apache/incubator-brpc/blob/rdma/src/brpc/rdma/rdma_communication_manager.cpp#L70) the cm_id will always bind to the first device. So the pd is not matched and create qp will return failed. #1401


wwbmmm:
Hi, 我们后续计划不再维护单独的RDMA分支，而是将RDMA功能合入master，相关PR：https://github.com/apache/incubator-brpc/pull/1836
所以这个PR先close了。如果有关于RDMA的改进建议，可以在新的PR下讨论

PR #1495, opened 2021-07-23
Fix apache thrift download failed

wwbmmm:
修复不稳定的持续集成


zyearn:
Thanks Weibing for the fix

PR #1535, opened 2021-08-19
update getting_started doc to remove gperftools since it is not needed

tanzhongyi003:


zyearn:
这个是为什么要删掉了？

tanzhongyi003:
because  we have removed gperftool dependency from the source
so I think that we may need to update the doc also.

zyearn:
But without gperftools, how can users do cpu/heap profiling? Do we remove this because of license issue?

tanzhongyi003:
yes, you remove it because of license issue, 
do you still remember it?

zyearn:
No I don't remember :) I can't find any context in the commit history too: https://github.com/apache/incubator-brpc/commits/master/LICENSE.

We still use those files, for example src/butil/gperftools_profiler.h, and their license is listed here https://github.com/apache/incubator-brpc/blob/master/LICENSE#L760, which is 3-clause BSD and it's apache compatible. 
There may be some misunderstandings here.

tanzhongyi003:
ok, maybe you are right, I will close it now.

PR #1536, opened 2021-08-20
Fix tools/get_brpc_revision.sh

### LINKED ISSUE #1046: adapt 'git log ... %cI' in tools/get_brpc_revision.sh to git 1.x
'%cI' in --format of 'git log' is supported since 2.4.0, so the time field is not displayed as expected if git with version less than 2.4.0 is used.

v1siuol:
Fix #1046

### DIFF
--- a/tools/get_brpc_revision.sh
+++ b/tools/get_brpc_revision.sh
@@ -16,10 +16,12 @@
 # limitations under the License.
 
 output=$(cat $1/RELEASE_VERSION)
-version=$(git log -1 --format="%h\\|%cI" 2> /dev/null)
+abbr_commit_hash=$(git log -1 --format="%h" 2> /dev/null)
+committer_date=($(git log -1 --format="%ci" 2> /dev/null))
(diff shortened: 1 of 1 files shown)


v1siuol:
I take a snippet of this change and test it in git1.8, which looks good to me.
```
[root@a2d83421c3b6 abstract_data_type]# git --version
git version 1.8.3.1
[root@a2d83421c3b6 abstract_data_type]# committer_date=($(git log -1 --format="%ci" 2> /dev/null))
[root@a2d83421c3b6 abstract_data_type]# committer_date="${committer_date[0]}T${committer_date[1]}${committer_date[2]:0:3}:${committer_date[2]:3:2}"
[root@a2d83421c3b6 abstract_data_type]# echo $committer_date
2020-10-19T12:03:03-04:00
```

And it won't break git2.x after the change.
```
$ ./tools/get_brpc_revision.sh .
0.9.7\|fix_get_brpc_revision\|74db116c\|2021-08-17T21:28:24+02:00
```

PR #1616, opened 2021-11-25
fix小typo

PengyiPan:
found when following the instruction


zyearn:
Thanks for the contribution

PR #1620, opened 2021-12-01
Fix HttpResponse error

jl2005:
When an incorrect HTTP request is received, the existing code returns an `HttpRequest`, which should actually return an `HttpResponse`.
before fixing:
```
# curl 127.0.0.1:8010/v1/queue/stop\ 1  -v
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 8010 (#0)
> GET /v1/queue/stop 1 HTTP/1.1
> Host: 127.0.0.1:8010
> User-Agent: curl/7.64.1
> Accept: */*
>
GET / HTTP/1.1
Host: 127.0.0.1:58674
Accept: */*
User-Agent: brpc/1.0 curl/7.0

* Closing connection 0
```
After modification:
```
# curl 127.0.0.1:8010/v1/queue/stop\ 1  -v
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 8010 (#0)
> GET /v1/queue/stop 1 HTTP/1.1
> Host: 127.0.0.1:8010
> User-Agent: curl/7.64.1
> Accept: */*
>
< HTTP/1.1 400 Bad Request
* no chunk, no close, no size. Assume close to signal end
<
* Closing connection 0
```

### DIFF
--- a/src/brpc/policy/http_rpc_protocol.cpp
+++ b/src/brpc/policy/http_rpc_protocol.cpp
@@ -1146,13 +1146,13 @@ ParseResult ParseHttpMessage(butil::IOBuf *source, Socket *socket,
                 return MakeParseError(PARSE_ERROR_NOT_ENOUGH_DATA);
             }
             // Send 400 back.
-            butil::IOBuf bad_req;
+            butil::IOBuf bad_resp;
             HttpHeader header;
             header.set_status_code(HTTP_STATUS_BAD_REQUEST);
-            MakeRawHttpRequest(&bad_req, &header, socket->remote_side(), NULL);
+            MakeRawHttpResponse(&bad_resp, &header, NULL);
             Socket::WriteOptions wopt;
             wopt.ignore_eovercrowded = true;
-            socket->Write(&bad_req, &wopt);
+            socket->Write(&bad_resp, &wopt);
             return MakeParseError(PARSE_ERROR_NOT_ENOUGH_DATA);
         } else {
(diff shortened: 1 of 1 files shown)


jl2005:
cc @chenzhangyi @wanglun @jbonofre @llhe @ke4qqq

src/brpc/policy/http_rpc_protocol.cpp
```diff
@@ -1146,13 +1146,13 @@ ParseResult ParseHttpMessage(butil::IOBuf *source, Socket *socket,
                 return MakeParseError(PARSE_ERROR_NOT_ENOUGH_DATA);
             }
             // Send 400 back.
-            butil::IOBuf bad_req;
+            butil::IOBuf bad_resp;
```
chenzhangyi, line 1149: The variable name should be bad_req, which indicates that the response content is 'Bad Request', rather than a 'bad' response.

jl2005 (commented):
Already fixed

src/brpc/policy/http_rpc_protocol.cpp
wasphin: Not a member.

chenzhangyi (approved):

chenzhangyi:
Merged, thanks for your contribution.

PR #1651, opened 2021-12-21
Fix rpc view `defer_close_second flag` typo

ehds:
`defer_close_seconds` should be `defer_close_second`.


wwbmmm:
@ehds Thanks for you contribution!

PR #1667, opened 2022-01-10
Add two cmake option for build

x2c3z4:
This is useful when integrating brpc to existing cmake project.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,7 @@ option(WITH_DEBUG_SYMBOLS "With debug symbols" ON)
 option(WITH_THRIFT "With thrift framed protocol supported" OFF)
 option(BUILD_UNIT_TESTS "Whether to build unit tests" OFF)
 option(DOWNLOAD_GTEST "Download and build a fresh copy of googletest. Requires Internet access." ON)
+option(BUILD_TOOLS "Whether to build tools" ON)
 
 # Enable MACOSX_RPATH. Run "cmake --help-policy CMP0042" for policy details.
 if(POLICY CMP0042)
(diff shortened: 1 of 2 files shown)


x2c3z4:
CI has down, please fix it.

CMakeLists.txt
```diff
@@ -25,6 +25,8 @@ option(WITH_DEBUG_SYMBOLS "With debug symbols" ON)
 option(WITH_THRIFT "With thrift framed protocol supported" OFF)
 option(BUILD_UNIT_TESTS "Whether to build unit tests" OFF)
 option(DOWNLOAD_GTEST "Download and build a fresh copy of googletest. Requires Internet access." ON)
+option(BUILD_TOOLS "Whether to build tools" ON)
+option(WITH_SHARED_LIB "With dynamic lib" ON)
```
lorinlee, line 29: Hi @x2c3z4 , Thanks for your contribution, could you please explain the reason for adding WITH_SHARED_LIB, thanks!

wwbmmm: Maybe it is better to be:
option(BUILD_SHARED_LIB "Whether to build shared lib" ON)

x2c3z4: > Hi @x2c3z4 , Thanks for your contribution, could you please explain the reason for adding WITH_SHARED_LIB, thanks!

I want to static link the brpc to my project, and my project has also static linked with leveldb/gflags/glog, so if I link the brpc shared lib, there are some errors that I should handle, so add a switch to control the shared lib is a good option.

x2c3z4: > Maybe it is better to be: option(BUILD_SHARED_LIB "Whether to build shared lib" ON)
OK.

zyearn:
Hi @x2c3z4, We had a similar PR(https://github.com/apache/incubator-brpc/pull/1797) which has been merged to solve the same problem. I'm going to close this PR. Thanks for the contribution anyway!

PR #1669, opened 2022-01-12
Fix C++ compile issue and clang compile issue

x2c3z4:
See commit messages.

### DIFF
--- a/src/brpc/redis.h
+++ b/src/brpc/redis.h
@@ -107,21 +107,21 @@ public:
     bool SerializeTo(butil::IOBuf* buf) const;
 
     // Protobuf methods.
-    RedisRequest* New() const;
-    void CopyFrom(const ::google::protobuf::Message& from);
-    void MergeFrom(const ::google::protobuf::Message& from);
+    RedisRequest* New() const override;
+    void CopyFrom(const ::google::protobuf::Message& from) override;
+    void MergeFrom(const ::google::protobuf::Message& from) override;
     void CopyFrom(const RedisRequest& from);
     void MergeFrom(const RedisRequest& from);
-    void Clear();
-    bool IsInitialized() const;
+    void Clear() override;
+    bool IsInitialized() const override;
   
-    int ByteSize() const;
+    int ByteSize() const override;
     bool MergePartialFromCodedStream(
-        ::google::protobuf::io::CodedInputStream* input);
+        ::google::protobuf::io::CodedInputStream* input) override;
     void SerializeWithCachedSizes(
-        ::google::protobuf::io::CodedOutputStream* output) const;
-    ::google::protobuf::uint8* SerializeWithCachedSizesToArray(::google::protobuf::uint8* output) const;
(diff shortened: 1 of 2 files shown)


src/bthread/mutex.h
```diff
@@ -48,15 +48,15 @@ class Mutex {
...
-            throw std::system_error(std::error_code(ec, std::system_category()), "Mutex constructor failed");
+            CHECK(false) << "Mutex constructor failed";
         }
     }
     ~Mutex() { CHECK_EQ(0, bthread_mutex_destroy(&_mutex)); }
     native_handler_type native_handler() { return &_mutex; }
     void lock() {
         int ec = bthread_mutex_lock(&_mutex);
         if (ec != 0) {
-            throw std::system_error(std::error_code(ec, std::system_category()), "Mutex lock failed");
```
lorinlee, line 59: Hi @x2c3z4 , Thanks for your contribution! But I have a small question, why remove the exception here, it was been added by #645

src/brpc/redis.h
```diff
@@ -107,21 +107,21 @@ class RedisRequest : public ::google::protobuf::Message {
     bool SerializeTo(butil::IOBuf* buf) const;
 
     // Protobuf methods.
-    RedisRequest* New() const;
-    void CopyFrom(const ::google::protobuf::Message& from);
-    void MergeFrom(const ::google::protobuf::Message& from);
+    RedisRequest* New() const override;
```
wwbmmm, line 110: Why "override" is needed only in "redis.h" while not in "thrift_message.h" or "nshead_message.h" or other headers files?

wwbmmm:
This branch has conflicts with master now.

src/brpc/redis.h
x2c3z4: Look like the redis.h is included by some very common header, and my project includes the "redis.h" indirectly.

src/bthread/mutex.h
x2c3z4: Google C++ code style suggests that don't use the c++ exception, and I think a lot of projects have disabled the c++ exception including mine. All the brpc code is worked with c++ no-exception, except this code snippet.

If we don't accept this error, just abort it.
As I know, if the mutex has been freed or is a bad memory address, the pthread_mutex_lock will return EINVAL(22 Invalid argument). CHECK(false) is better than c++ exception.

src/brpc/redis.h
wwbmmm: I see "redis.h" is included in "server.h".
It is ok to add "override" to "redis.h".

src/bthread/mutex.h
JimChengLin: @x2c3z4 Google Style is good but I guess there is no way to ensure no one else is using this behavior. Removing the exception may break others codebase.

x2c3z4: What codebase will it break? I don't understand why needs to throw exception here.

PR #1680, opened 2022-01-22
fix(input_messenger) client side retry policy

ehds:
*Client Ptotocol Rule*: Ptotocol in client-side is fixed except `baidu_std`, because of `streaming_rpc` need create by `baidu_std` ptotocol. 

But in current implementation， there is a chance that client will accept message but not match with current protocol.

Case:

Client issue two RPC names `A` (use `hulu` protocol) and `B`(use` baidu_std` protocol) to server in the same connection, If A is sent before B,  client will set preferred protocol to `baidu_std`,  server will accept these two request and response. Then if response of B received by client  before A, B will be parsed success, because it is matched with current protocol(`baidu_std`). When parse response of A by current preferred ptotocol(`baidu_std`) will fail, according to *Client Ptotocol Rule*, client think this response is `streaming_rpc`, and try other ptotocols, then response of B will be parsed ok by `hulu` ptotocol.

That's the point: Shall  we just retry `streaming_rpc` when parse failed if current protocol is `baidu_std`? instead of trying all other protocols, or other ptotocol mixed with `baidu_std` will be accepted successfully.

### DIFF
--- a/src/brpc/input_messenger.cpp
+++ b/src/brpc/input_messenger.cpp
@@ -67,31 +67,42 @@ ParseResult InputMessenger::CutInputMessage(
     // selection or by client.
     if (preferred >= 0 && preferred <= max_index
             && _handlers[preferred].parse != NULL) {
-        ParseResult result =
-            _handlers[preferred].parse(&m->_read_buf, m, read_eof, _handlers[preferred].arg);
-        if (result.is_ok() ||
-            result.error() == PARSE_ERROR_NOT_ENOUGH_DATA) {
-            *index = preferred;
-            return result;
-        } else if (result.error() != PARSE_ERROR_TRY_OTHERS) {
-            // Critical error, return directly.
-            LOG_IF(ERROR, result.error() == PARSE_ERROR_TOO_BIG_DATA)
-                << "A message from " << m->remote_side()
-                << "(protocol=" << _handlers[preferred].name
-                << ") is bigger than " << FLAGS_max_body_size
-                << " bytes, the connection will be closed."
-                " Set max_body_size to allow bigger messages";
-            return result;
-        }
-        if (m->CreatedByConnect() &&
-            // baidu_std may fall to streaming_rpc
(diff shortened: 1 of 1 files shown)


wwbmmm (approved):

wwbmmm:
@ehds Thanks for you contribution!

src/brpc/input_messenger.cpp
```diff
@@ -67,31 +67,42 @@ ParseResult InputMessenger::CutInputMessage(
...
-                       << " by " << _handlers[preferred].name 
-                       << " at client-side";
-            return MakeParseError(PARSE_ERROR_ABSOLUTELY_WRONG);
-        }
+        int cur_index = preferred;
+        do {
+            ParseResult result =
+                _handlers[cur_index].parse(&m->_read_buf, m, read_eof, _handlers[cur_index].arg);
+            if (result.is_ok() ||
+                result.error() == PARSE_ERROR_NOT_ENOUGH_DATA) {
```
lorinlee, line 75: 这里如果cur_index先是PROTOCOL_BAIDU_STD，后重试PROTOCOL_STREAMING_RPC成功，需要set_preferred_index吧

src/brpc/input_messenger.cpp
```diff
@@ -67,31 +67,42 @@ ParseResult InputMessenger::CutInputMessage(
...
+                    LOG(ERROR) << "Fail to parse response from " << m->remote_side()
+                        << " by " << _handlers[preferred].name 
+                        << " at client-side";
+                    return MakeParseError(PARSE_ERROR_ABSOLUTELY_WRONG);
+                }
+            } else {
+                // Try other protocols.
+                break;
+            }
+        } while (1);
```
lorinlee, line 105: 这里用while(true)吧，while(1)其实引入了implicit conversion

src/brpc/input_messenger.cpp
ehds: 是的，否则 PROTOCOL_STREAMING_RPC 每次都会重试两次。

PR #1682, opened 2022-01-24
Fix build warning, ByteSize() is deprecated, use ByteSizeLong() instead

yangzhg:
Fix build warning, protobuf ByteSize() is deprecated, use ByteSizeLong() instead


src/brpc/policy/baidu_rpc_protocol.cpp
```diff
@@ -69,7 +69,7 @@ inline void PackRpcHeader(char* rpc_header, int meta_size, int payload_size) {
 
 static void SerializeRpcHeaderAndMeta(
     butil::IOBuf* out, const RpcMeta& meta, int payload_size) {
-    const int meta_size = meta.ByteSize();
+    const int meta_size = meta.ByteSizeLong();
```
wwbmmm, line 72: Thanks for your contribution!
ByteSizeLong() is not available in old version protobuf, for compatibility with old version protobuf, you should test protobuf version here.

wwbmmm (changes requested):

PR #1719, opened 2022-03-14
[Compile] avoid compile error after gcc11 with --std=c++20

hiberabyss:


hiberabyss:
Could anyone have a look of this?

cdjingit:
thanks for your fix!

PR #1736, opened 2022-04-11
[UPDATE] fix c struct compile error

wolfdan666:
solve this error. make it the same as bthread.h


jamesge:
under which compiler(and which version) is this error produced?

wolfdan666:
> under which compiler(and which version) is this error produced?

gcc version 7.5.0 (GCC)

PR #1762, opened 2022-05-13
Check append return code

### LINKED ISSUE #1761: What's the behavior of SendRpcResponse in memory constrained env?
It seems `SendRpcResponse` contains heap allocations (`butil::IOBuf::append`) which returns `-1` if malloc fails. This error code is silently dropped when building response. Do we have defined behavior when response is partially built and sent back to client?

ref:

https://github.com/apache/incubator-brpc/blob/master/src/brpc/policy/baidu_rpc_protocol.cpp#L222-L229

btw, can we mark `SendRpcResponse` as **noexcept** ?

wwbmmm:
Fix #1761

### DIFF
--- a/src/brpc/policy/baidu_rpc_protocol.cpp
+++ b/src/brpc/policy/baidu_rpc_protocol.cpp
@@ -80,11 +80,11 @@ static void SerializeRpcHeaderAndMeta(
         ::google::protobuf::io::CodedOutputStream coded_out(&arr_out);
         meta.SerializeWithCachedSizes(&coded_out); // not calling ByteSize again
         CHECK(!coded_out.HadError());
-        out->append(header_and_meta, sizeof(header_and_meta));
+        CHECK_EQ(0, out->append(header_and_meta, sizeof(header_and_meta)));
     } else {
         char header[12];
         PackRpcHeader(header, meta_size, payload_size);
-        out->append(header, sizeof(header));
+        CHECK_EQ(0, out->append(header, sizeof(header)));
         butil::IOBufAsZeroCopyOutputStream buf_stream(out);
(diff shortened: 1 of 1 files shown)


src/brpc/policy/baidu_rpc_protocol.cpp
```diff
@@ -80,11 +80,11 @@ static void SerializeRpcHeaderAndMeta(
         ::google::protobuf::io::CodedOutputStream coded_out(&arr_out);
         meta.SerializeWithCachedSizes(&coded_out); // not calling ByteSize again
         CHECK(!coded_out.HadError());
-        out->append(header_and_meta, sizeof(header_and_meta));
+        CHECK_EQ(0, out->append(header_and_meta, sizeof(header_and_meta)));
```
amosbird, line 83: IIUC this is just a assertion and will only be effective in debug build. Perhaps we should change the return type of `SerializeRpcHeaderAndMeta` into `bool` and handle it properly, similar to what we did for `SerializeAsCompressedData`.

wwbmmm: // The CHECK(condition) macro is active in both debug and release builds and
// effectively performs a LOG(FATAL) which terminates the process and
// generates a crashdump unless a debugger is attached.

https://github.com/apache/incubator-brpc/blob/master/src/butil/logging.h#L122

amosbird: Ok... It's interesting to see a release assertion macro) However, this error is recoverable. Resorting to `std::terminate` seems to be an overkill.

serverglen: LGTM

PR #1777, opened 2022-06-06
FIX BUG(Issue #1776): Switching pthread causes coredump when bthread local variable is destructed

zhaodongzhi:
FIX BUG(Issue #1776): Switching pthread causes coredump when bthread local variable is destructed

If the bthread local destructor specified by bthread_key_create calls bthread_mutex_lock
it may cause TaskGroup::task_runner->return_keytable to switch pthreads after execution
causing coredump

### DIFF
--- a/src/bthread/task_group.cpp
+++ b/src/bthread/task_group.cpp
@@ -297,9 +297,6 @@ void TaskGroup::task_runner(intptr_t skip_remained) {
             thread_return = e.value();
         }
 
-        // Group is probably changed
-        g = tls_task_group;
-
         // TODO: Save thread_return
         (void)thread_return;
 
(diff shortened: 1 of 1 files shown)


wwbmmm:
Thanks for reporting the issue and submitting this PR!
I think it's better to add some comments to explain the code, or add some test cases, to prevent future modification from breaking this case.

wwbmmm:
> > Thanks for reporting the issue and submitting this PR! I think it's better to add some comments to explain the code, or add some test cases, to prevent future modification from breaking this case.
> 
> OK! I will add some test cases and then update this pr

@zhaodongzhi Any update of this PR?

chenBright:
This bug has been fixed in #2558 .

PR #1804, opened 2022-06-17
support handle on_received_messages failure of streaming rpc

### LINKED ISSUE #1803: Request for support streaming RPC data receive failure handling
The `StreamInputHandler` on_received_messages function has return value, but  no one will handle the return value. I think it is better to make it possible for handling the failure

yangzhg:
closed #1803 

support handle `on_received_messages` failure of streaming rpc, user can overload `on_failure` of `StreamInputHandler`  to handle `on_received_messages` failure,  the default is do nothing

### DIFF
--- a/src/brpc/stream.cpp
+++ b/src/brpc/stream.cpp
@@ -462,8 +462,12 @@ public:
     ~MessageBatcher() { flush(); }
     void flush() {
         if (_size > 0 && _s->_options.handler != NULL) {
-            _s->_options.handler->on_received_messages(
+            int ret = _s->_options.handler->on_received_messages(
                     _s->id(), _storage, _size);
+            if (BAIDU_UNLIKELY(ret != 0)) {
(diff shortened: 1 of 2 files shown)


src/brpc/stream.cpp
```diff
@@ -462,8 +462,12 @@ class MessageBatcher {
     ~MessageBatcher() { flush(); }
     void flush() {
         if (_size > 0 && _s->_options.handler != NULL) {
-            _s->_options.handler->on_received_messages(
+            int ret = _s->_options.handler->on_received_messages(
                     _s->id(), _storage, _size);
+            if (BAIDU_UNLIKELY(ret != 0)) {
+                LOG(WARNING) << "Fail to receive message for stream: " << _s->id();
+                _s->_options.handler->on_failure(_s->id());
```
wwbmmm, line 469: I think it's unnecessary to call `on_failure` here. Since user knows what will return from `on_received_messages`, they can call `on_failure` in `on_received_messages`.

PR #1896, opened 2022-08-19
fix: flamegraph initialization

zbzbw:
Currently the <script> tag inside returned data from `flamegraph.pl` is dropped by jQuery, and `init` won't work (so does features like zoom in/out: https://github.com/apache/incubator-brpc/pull/864#issuecomment-1149521212)
This PR fixed the issue mentioned above.


zyearn:
Could you write something more about how you tested the code? For example, what's the difference in UI before and after?

PR #1898, opened 2022-08-20
Fix the slow test in brpc_socket_unittest.cpp

zyearn:
Fix the slow test https://app.travis-ci.com/github/apache/incubator-brpc/jobs/580274145


wwbmmm:
CI失败了
brpc_socket_unittest.cpp:750: Failure
Expected: (butil::gettimeofday_us()) < (start_time + 1100000L), actual: 1661020823868252 vs 1661020823867190

PR #1904, opened 2022-08-24
delete _dl_sym

pangyoki:
ATT


wwbmmm:
Why delete _dl_sym?

pangyoki:
> Why delete _dl_sym?

_dl_sym can't be used in ubuntu22.10. It will get error `GLIBC_PRIVATE _dl_sym not_found`.

wwbmmm:
This error have been fixed by https://github.com/apache/incubator-brpc/pull/1784
_dl_sym is now weak symbol, it will not cause undefined reference.

PR #1925, opened 2022-09-14
[document] Add vcpkg instruction step

JackBoosY:
`brpc` is available as a port in [vcpkg](https://github.com/microsoft/vcpkg), a C++ library manager that simplifies installation for `brpc` and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build `brpc`, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/brpc/portfile.cmake). We try to keep the library maintained as close as possible to the original library. :)

### DIFF
--- a/docs/en/getting_started.md
+++ b/docs/en/getting_started.md
@@ -107,6 +107,18 @@ Examples link brpc statically, if you need to link the shared version, remove `C
 $ mkdir build && cd build && cmake -DBUILD_UNIT_TESTS=ON .. && make && make test
 ```
 
+### Compile brpc with vcpkg
+
+[vcpkg](https://github.com/microsoft/vcpkg) is a package manager that supports all platforms,
+you can use vcpkg to build llvm with the following step:
+
+```shell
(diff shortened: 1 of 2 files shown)


docs/en/getting_started.md
```diff
@@ -107,6 +107,18 @@ Examples link brpc statically, if you need to link the shared version, remove `C
 $ mkdir build && cd build && cmake -DBUILD_UNIT_TESTS=ON .. && make && make test
 ```
 
+### Compile brpc with vcpkg
+
+[vcpkg](https://github.com/microsoft/vcpkg) is a package manager that supports all platforms,
+you can use vcpkg to build llvm with the following step:
```
day253, line 113: typo?
llvm should be brpc?

JackBoosY: Sorry, typo.

Issue #1640, opened 2021-12-16
更新brpc在各个公司的落地场景


tanzhongyi003:
https://github.com/apache/incubator-brpc/blob/master/community/cases.md

针对这个文件，提交PR，然后找我review即可。

Divyansh200102:
@wasphin could you please take a look at #2540

PR #2003, opened 2022-11-18
src/butil: correct assign atomic pointer member to be NULL

changchengx:
1. memset assign value to var without considering var type.
2. memset can't guarantee atomic operation

Signed-off-by: Liu, Changcheng <changcheng.liu@aliyun.com>


wwbmmm:
@changchengx Do you have any problem using memset ? (eg. compile error or unexpected behavior at runtime)

changchengx:
@wwbmmm No. From semantic perspective, memset doesn't consider variable type at all and guarantee atomic operation.

serverglen:
Could you please submit the application scenario of brpc in your company?
See #1640

PR #2009, opened 2022-11-22
Early return when size == 0

372046933:

Issue Number:

Problem Summary:

Changed:

Side effects:
- Performance effects(性能影响):

---
### Check List:

As pointed out by @Tuvie https://github.com/apache/incubator-brpc/issues/1995#issuecomment-1323723025, when `IOBuf::append_user_data` is called with size 0, we can do early return. This avoids unnecessary block memory allocation in the later.

### DIFF
--- a/src/butil/iobuf.cpp
+++ b/src/butil/iobuf.cpp
@@ -1211,18 +1211,16 @@ int IOBuf::append_user_data(void* data, size_t size, void (*deleter)(void*)) {
         LOG(FATAL) << "data_size=" << size << " is too large";
         return -1;
     }
+    if (!deleter) {
+        deleter = ::free;
+    }
     if (!size) {
-        if (deleter) {
-            deleter(data);
-        }
+        deleter(data);
         return 0;
     }
     char* mem = (char*)malloc(sizeof(IOBuf::Block) + sizeof(UserDataExtension));
     if (mem == NULL) {
         return -1;
-    }
-    if (deleter == NULL) {
-        deleter = ::free;
     }
     IOBuf::Block* b = new (mem) IOBuf::Block((char*)data, size, deleter);
     const IOBuf::BlockRef r = { 0, b->cap, b };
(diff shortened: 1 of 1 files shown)


src/butil/iobuf.cpp
```diff
@@ -1211,6 +1211,9 @@ int IOBuf::append_user_data(void* data, size_t size, void (*deleter)(void*)) {
         LOG(FATAL) << "data_size=" << size << " is too large";
         return -1;
     }
+    if (!size) {
+        return 0;
```
wwbmmm, line 1215: Maybe we need to call deleter here?

Tuvie: I still think size==0 should return -1 rather than 0. If it is failed, it means IOBuf does not take this Block at all. It is the caller's responsibility to delete the buffer.

372046933: @Tuvie Makes sense. Does return -1 break backward compatibility? 
Also return 0 && call deleter means that IOBuf takes the responsibility to free the buff. This is consistent with the behavior before this PR.

wwbmmm:
LGTM

src/butil/iobuf.cpp
372046933: @wwbmmm Could you have a look and decide which is better?

wwbmmm: I think call deleter and return 0 is ok.

372046933: Thanks. Maybe we can merge now 😊

zyearn: Shouldn't we move the deleter null check and assignment below before this change? Otherwise the data may be leaked if user doesn't pass deleter.

372046933: @zyearn Yes, before this change,  when size equals 0, `append_user_data` relies on the default `free` to release user memory if deleter not specified. We should maintain the semantics here. CC @wwbmmm

wwbmmm: LGTM

PR #2019, opened 2022-11-26
ssl错误码问题

yyweii:
> ERR_get_error() returns the earliest error code from the thread's error queue and removes the entry. This function can be called repeatedly until there are no more error codes to return.

据文档描述，ERR_get_error是会修改线程的错误码队列的，所以多次调用该函数，返回的值并不是同一个。
#1096 可能和这个问题有关。我司线上也有大量的“Fail to write into ssl_fd=, Success”的日志。这个Success可能是第二次调用ERR_get_error的返回值。

> SSL_get_error() returns a result code (suitable for the C "switch" statement) for a preceding call to SSL_connect(), SSL_accept(), SSL_do_handshake(), SSL_read_ex(), SSL_read(), SSL_peek_ex(), SSL_peek(), SSL_shutdown(), SSL_write_ex() or SSL_write() on ssl.

> In addition to ssl and ret, SSL_get_error() inspects the current thread's OpenSSL error queue. Thus, SSL_get_error() must be used in the same thread that performed the TLS/SSL I/O operation, and no other OpenSSL function calls should appear in between. The current thread's error queue must be empty before the TLS/SSL I/O operation is attempted, or SSL_get_error() will not work reliably.

在调用SSL_do_handshake、SSL_read、SSL_write之前最好是先调用ERR_clear_error，这样才能保证这些函数之后调用的SSL_get_error的返回值正确。

### DIFF
--- a/src/brpc/details/ssl_helper.cpp
+++ b/src/brpc/details/ssl_helper.cpp
@@ -350,7 +350,7 @@ static int LoadCertificate(SSL_CTX* ctx,
     if (err != 0 && (ERR_GET_LIB(err) != ERR_LIB_PEM
                      || ERR_GET_REASON(err) != PEM_R_NO_START_LINE)) {
         LOG(ERROR) << "Fail to read chain certificate in "
-                   << certificate << ": " << SSLError(ERR_get_error());
+                   << certificate << ": " << SSLError(err);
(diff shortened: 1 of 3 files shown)


wwbmmm:
LGTM

PR #2020, opened 2022-11-27
Fix MacOS warning: bool literal returned from 'main' [-Wmain]

zyearn:

Problem Summary:
https://github.com/apache/incubator-brpc/actions/runs/3559564054/jobs/5979002361


serverglen:
LGTM

PR #2180, opened 2023-03-27
Fix core when ssl is enabled

warriorpaw:
OpenSSL does not guarantee that any of its objects can be used concurrently by multiple threads:
https://github.com/openssl/openssl/issues/2165
https://github.com/openssl/openssl/issues/20446

Related issue with this PR: #2166

And this PR is not the best solution for this core issue.
The best way is put SSL obj's read/write operate in same event loop but this changes too much of the current code structure, i'm too lazy to do it.

### DIFF
--- a/src/brpc/socket.cpp
+++ b/src/brpc/socket.cpp
@@ -1505,6 +1505,7 @@ X509* Socket::GetPeerCertificate() const {
     if (ssl_state() != SSL_CONNECTED) {
         return NULL;
     }
+    BAIDU_SCOPED_LOCK(_ssl_session_mutex);
     return SSL_get_peer_certificate(_ssl_session);
 }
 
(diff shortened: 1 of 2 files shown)


wwbmmm:
Does this affect performance?

jingshi-ant:
> Does this affect performance?

about 20% slower in my 100w data set test. (mainly because of ssl enable. if remove lock and enable ssl, the speed is almost the same with lock added.)

warriorpaw:
> Does this affect performance?

Brpc already piped write requests by linked list and only one bthread calls ssl_write.

In data reading, brpc use epoll in edge triggered mode and an atomic variable to make sure there is only one bthread is calling ssl_read.

Adding a mutex here will introduce some competition, but it should be very slight. Further this mutex will avoided the competition of using underlay socket fd. So, the impact of this change on performance should be almost none.

wwbmmm:
LGTM

warriorpaw:
@jamesge @Huixxi @wwbmmm @chenBright 
Please take another look if this change can be merged, thanks!

chenBright:
LGTM

PR #2301, opened 2023-07-04
fix ignore SIGPIPE signal concurrent issue with other library

lrita:

Issue Number:

Problem Summary:
There is a crash issue that at beginning many library concurrently ignore SIGPIPE to assert failed.

none
Changed:
none
Side effects:
- Performance effects(性能影响):
none
none


wwbmmm:
LGTM

PR #2326, opened 2023-07-24
fix `huge` rdma_recv_block_type no effect issue.

legionxiong:
It has no effect for `huge` type of recv_block_size
    which is as large as 2^21, because the `block_size`
    variable is defined as type `uint16_t`, its maximum
    value is 2^16. So, change the type of `block_size`
    to `uint32_t` with relevant updates.

Issue Number:

Problem Summary:

Changed:

Side effects:
- Performance effects(性能影响):

---
### Check List:

### DIFF
--- a/src/brpc/rdma/rdma_endpoint.cpp
+++ b/src/brpc/rdma/rdma_endpoint.cpp
@@ -71,20 +71,20 @@ static const size_t RESERVED_WR_NUM = 3;
 // message length (2B)
 // hello version (2B)
 // impl version (2B): 0 means should use tcp
-// block size (2B)
+// block size (4B)
 // sq size (2B)
 // rq size (2B)
 // GID (16B)
 // QP number (4B)
 static const char* MAGIC_STR = "RDMA";
 static const size_t MAGIC_STR_LEN = 4;
-static const size_t HELLO_MSG_LEN_MIN = 38;
+static const size_t HELLO_MSG_LEN_MIN = 40;
 static const size_t HELLO_MSG_LEN_MAX = 4096;
 static const size_t ACK_MSG_LEN = 4;
-static uint16_t g_rdma_hello_msg_len = 38;  // In Byte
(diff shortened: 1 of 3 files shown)


src/brpc/rdma/rdma_endpoint.cpp
```diff
@@ -71,20 +71,20 @@ static const size_t RESERVED_WR_NUM = 3;
...
 // QP number (4B)
 static const char* MAGIC_STR = "RDMA";
 static const size_t MAGIC_STR_LEN = 4;
-static const size_t HELLO_MSG_LEN_MIN = 38;
+static const size_t HELLO_MSG_LEN_MIN = 40;
 static const size_t HELLO_MSG_LEN_MAX = 4096;
 static const size_t ACK_MSG_LEN = 4;
-static uint16_t g_rdma_hello_msg_len = 38;  // In Byte
+static uint16_t g_rdma_hello_msg_len = 40;  // In Byte
 static uint16_t g_rdma_hello_version = 1;
```
Tuvie, line 85: Since we modify the hello message format, we should change the hello version.

legionxiong: Done.

test/brpc_rdma_unittest.cpp
```diff
@@ -203,7 +203,7 @@ TEST_F(RdmaTest, client_hello_msg_invalid_magic_str) {
     Socket* s = GetSocketFromServer(0);
     ASSERT_EQ(rdma::RdmaEndpoint::UNINIT, s->_rdma_ep->_state);
 
-    uint8_t data[38];
+    uint8_t data[40];
```
wwbmmm, line 206: use named constant instead of 40

legionxiong: Done.

wwbmmm:
LGTM

Issue #2286, opened 2023-06-21
Grpc 向 Brpc迁移遇到的问题


982945902:
**Describe the bug (描述bug)**
client,server都采用h2:grpc协议，client请求server直连一切正常, 但是通过nginx转发会有一半出错, 请教下原因, 报错如下

upstream sent invalid http2 table index: 63 while reading response header from upstream
upstream sent invalid header while reading response header from upstream
错误码502 bad gate way

**To Reproduce (复现方法)**
必现
**Expected behavior (期望行为)**
正常通过

**Versions (各种版本)**
OS:ubuntu1~22.04
Compiler:gcc version 11.3.0 (Ubuntu 11.3.0-1ubuntu1~22.04)
brpc:brpc-1.4.0
protobuf:protobuf-3.19.4

**Additional context/screenshots (更多上下文/截图)**
![image](https://github.com/apache/brpc/assets/49055103/42d89d0d-96f7-44c7-9dc9-1c5dd7c7afa8)

jiangdongzi:
同遇到

jiangdongzi:
https://github.com/apache/brpc/issues/2176

982945902:
> 同遇到

你可以看下我fork 的分支， 暂时规避了这个问题 https://github.com/982945902/brpc/tree/a1

982945902:
https://trac.nginx.org/nginx/ticket/1538

PR #2373, opened 2023-09-06
fix h2 protocol abnormal  case when SETTINGS_HEADER_TABLE_SIZE set to zero

982945902:

when use h2:grpc protocol pass nginx,  if nginx not support dynamic HPACK,  it will set SETTINGS_HEADER_TABLE_SIZE value to 0，now brpc not address this case.

https://trac.nginx.org/nginx/ticket/1538
rfc reference: https://datatracker.ietf.org/doc/html/rfc9113 4.3.1

Issue Number:

https://github.com/apache/brpc/issues/2286
https://github.com/apache/brpc/issues/2176

Problem Summary:
ignore a h2 protocol param set case

Changed:
add a if check when encode h2 header

Side effects:
- Performance effects(性能影响):
It hardly makes a difference

completely compatible
---
### Check List:

### DIFF
--- a/src/brpc/policy/http2_rpc_protocol.cpp
+++ b/src/brpc/policy/http2_rpc_protocol.cpp
@@ -1569,6 +1569,10 @@ H2UnsentRequest::AppendAndDestroySelf(butil::IOBuf* out, Socket* socket) {
     HPackOptions options;
     options.encode_name = FLAGS_h2_hpack_encode_name;
     options.encode_value = FLAGS_h2_hpack_encode_value;
+    if (ctx->remote_settings().header_table_size == 0) {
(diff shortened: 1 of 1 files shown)


982945902:
Put it another way because of dynamic table size limitation, HTTP/2.0 uses dynamic tables to store the most recently used request header fields for compression and reuse. If the dynamic table reaches a size limit, the new request Header Field is sent using the Literal Header Field without Indexing encoding to avoid overflows.

982945902:
https://datatracker.ietf.org/doc/html/rfc7541#section-4.2 about Maximum Table Size

wwbmmm:
LGTM

PR #2387, opened 2023-09-19
use AppendHeader for http2

lhsoft:

Issue Number:

Problem Summary:

Changed:

Side effects:
- Performance effects(性能影响):

---
### Check List:

### DIFF
--- a/test/brpc_http_rpc_protocol_unittest.cpp
+++ b/test/brpc_http_rpc_protocol_unittest.cpp
@@ -1544,7 +1544,7 @@ TEST_F(HttpTest, http2_header_after_data) {
     ASSERT_EQ(res_header.content_type(), "application/proto");
     // Check overlapped header is overwritten by the latter.
     const std::string* user_defined1 = res_header.GetHeader("user-defined1");
-    ASSERT_EQ(*user_defined1, "overwrite-a");
+    ASSERT_EQ(*user_defined1, "a,overwrite-a");
     const std::string* user_defined2 = res_header.GetHeader("user-defined2");
     ASSERT_EQ(*user_defined2, "b");
 }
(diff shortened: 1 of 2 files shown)


test/brpc_http_rpc_protocol_unittest.cpp
```diff
@@ -1544,7 +1544,7 @@ TEST_F(HttpTest, http2_header_after_data) {
     ASSERT_EQ(res_header.content_type(), "application/proto");
     // Check overlapped header is overwritten by the latter.
     const std::string* user_defined1 = res_header.GetHeader("user-defined1");
-    ASSERT_EQ(*user_defined1, "overwrite-a");
+    ASSERT_EQ(*user_defined1, "a,overwrite-a");
```
wwbmmm, line 1547: The original test case was written as "overwrite-a", was it expected to be overwritten? Changing the behavior here may not conform to the previous design.

lhsoft: The test case is `http2_header_after_data`.  
For http2,  according to the rfc : [ https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2](url), it shouldn't be expected to be overwritten

wwbmmm: Do you know why it was designed to overwrite before? @zyearn

zyearn: RFC doesn't mention overwritten. It says: Multiple message-header fields with the same field-name MAY be present in a message if and only if the entire field-value for that header field is defined as a comma-separated list [i.e., #(values)].

In brpc implementation, if the incoming header is in the format of "key:value1,value2,value3", brpc won't overwrite cases like this. But combining headers seems reasonable otherwise the previous header would be lost.

zyearn: LGTM

zyearn (approved):

PR #2402, opened 2023-10-07
Update logo in readme

Sea-n:
> (template text deleted)

Original Logo in README:

It might be better if we adjust for dark mode.

New logo:
| Light Mode | Dark Mode |
| :--: | :--: |
|  |  |

Logo source is modified from https://brpc.apache.org/:

Not sure should we keep the "Apache" here


serverglen:
LGTM

wwbmmm:
It's better to keep the original `logo.png` as dark mode and `logo-light.png` as light mode

Sea-n:
Oh, I get confused at that time. 😖

How about now? changed the dark mode image to `logo-white.png` and keep the black one (light mode) as `logo.png`.

PR #2422, opened 2023-10-23
[FIX] fix compile on ubuntu18.04 + use anonymous namespace for hardco…

Simonhancrew:
### Problem Summary:

1. comile error on ubuntu18.04, `BRPC/src/brpc/redis_command.cpp:411:29: error: ‘numeric_limits’ is not a member of ‘std’`
2. CTX_WIDTH should in anonymous namespace, only used in this cpp file.

### Changed:

1. compile on ubuntu18.04, redis_commands.cpp use numeric_limits<>, while missing <limits> header, cause compile error
2. use anonymous namespace for hardcode var, dont emit this to brpc namespace

Side effects:

+ no side effect

### DIFF
--- a/src/brpc/redis_command.cpp
+++ b/src/brpc/redis_command.cpp
@@ -15,14 +15,19 @@
 // specific language governing permissions and limitations
 // under the License.
 
+#include <limits>
 
 #include "butil/logging.h"
 #include "brpc/log.h"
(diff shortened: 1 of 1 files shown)


Simonhancrew:
@wwbmmm 烦请老哥take a look, 谢谢

PR #2494, opened 2024-01-04
tests: use global CMAKE_CPP_FLAGS

OdyX:

Problem Summary: when passing `-DCMAKE_CPP_FLAGS="-D_FORTIFY_SOURCE=2"` (for example, in Debian packaging), this doesn't get propagated in the `tests/` directory

Changed: pass the command-line `CMAKE_CPP_FLAGS` into tests/ compilation…

Side effects:
- None


chenBright:
Thanks for the contribution.

PR #2655, opened 2024-05-31
[feat] latency scale factor

superhail:

Issue Number:

Problem Summary: brpc's latency dispaly is displayed as `us`, while in our prod env, `ms` is more meaningful.

Changed: provide a gflag to change latency's scale factor

Side effects:
- Performance effects(性能影响): Little or none

- Breaking backward compatibility(向后兼容性): No

---
### Check List:

### DIFF
--- a/src/brpc/server.cpp
+++ b/src/brpc/server.cpp
@@ -123,8 +123,6 @@ const int INITIAL_CERT_MAP = 64;
 // compilation units is undefined.
 const int s_ncore = sysconf(_SC_NPROCESSORS_ONLN);
 
-DEFINE_uint64(latency_scale_factor, 1, "latency scale factor, used by method status, etc., latency_us = latency * latency_scale_factor");
-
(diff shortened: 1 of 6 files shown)


src/brpc/server.cpp
```diff
@@ -123,6 +123,8 @@ const int INITIAL_CERT_MAP = 64;
 // compilation units is undefined.
 const int s_ncore = sysconf(_SC_NPROCESSORS_ONLN);
 
+DEFINE_uint64(latency_scale_factor, 1, "latency scale factor, used by method status, etc., latency_us = latency * latency_scale_factor");
```
wwbmmm, line 126: I think this gflag can be directly defined in `latency_recorder.cpp`, so that it doesn't need to be passed through so many layers of parameters.

superhail: ok

superhail: @wwbmmm done

PR #2798, opened 2024-10-28
Fix memory leak of ArenaRpcPBMessageFactory

chenBright:

Issue Number:

Problem Summary:

Changed:

Side effects:
- Performance effects(性能影响):

---
### Check List:
- Please follow [Contributor Covenant Code of Conduct](https://github.com/apache/brpc/blob/master/CODE_OF_CONDUCT.md).(请遵循贡献者准则).

### DIFF
--- a/src/brpc/rpc_pb_message_factory.h
+++ b/src/brpc/rpc_pb_message_factory.h
@@ -113,6 +113,7 @@ public:
         auto arena_messages = static_cast<ArenaRpcPBMessages*>(messages);
         arena_messages->request = NULL;
         arena_messages->response = NULL;
+        arena_messages->arena.Reset();
         butil::return_object(arena_messages);
     }
 


sikey647:
@chenBright  请问下，这里使用 arena.Reset() 是直接释放内存；还是说内存空间保留，只是清理空间上的内容。

chenBright:
[析构后释放内](https://github.com/protocolbuffers/protobuf/blob/1f5df18680a5084fc9b4e4150b22f0b388d678dd/src/google/protobuf/arena.h#L305-L310)

```c++
  // Frees all storage allocated by this arena after calling destructors
  // registered with OwnDestructor() and freeing objects registered with Own().
  // Any objects allocated on this arena are unusable after this call. It also
  // returns the total space used by the arena which is the sums of the sizes
  // of the allocated blocks. This method is not thread-safe.
  uint64_t Reset() { return impl_.Reset(); }
```

Issue #2825, opened 2024-11-22
brpc的brpc_metrics的输出格式与prometheus的标准不符，尤其MultiDimension<bvar::LatencyRecorder>方式


xinhua5:
**Is your feature request related to a problem? (你需要的功能是否与某个问题有关?)**

使用MultiDimension<bvar::LatencyRecorder> 自定义了监控埋点，输出之后看到的结果是这样的：
> \# TYPE server_latency{db_id="112",shard_id="2",method="fetch",stage="next",quantile="80"} gauge
> server_latency{db_id="112",shard_id="2",method="fetch",stage="next",quantile="80"} 4076
> \# HELP server_latency{db_id="112",shard_id="2",method="fetch",stage="next",quantile="90"}
> \# TYPE server_latency{db_id="112",shard_id="2",method="fetch",stage="next",quantile="90"} gauge
> server_latency{db_id="112",shard_id="2",method="fetch",stage="next",quantile="90"} 5185
> \# HELP server_latency{db_id="112",shard_id="2",method="fetch",stage="next",quantile="99"}
> \# TYPE server_latency{db_id="112",shard_id="2",method="fetch",stage="next",quantile="99"} gauge
> server_latency{db_id="112",shard_id="2",method="fetch",stage="next",quantile="99"} 5185
> \# HELP server_latency{db_id="112",shard_id="2",method="fetch",stage="next",quantile="999"}
> \# TYPE server_latency{db_id="112",shard_id="2",method="fetch",stage="next",quantile="999"} gauge
> server_latency{db_id="112",shard_id="2",method="fetch",stage="next",quantile="999"} 5185
> \# HELP server_latency{db_id="112",shard_id="2",method="fetch",stage="next",quantile="9999"}
> \# TYPE server_latency{db_id="112",shard_id="2",method="fetch",stage="next",quantile="9999"} gauge
> server_latency{db_id="112",shard_id="2",method="fetch",stage="next",quantile="9999"} 5185

**Describe the solution you'd like (描述你期望的解决方法)**

prometheus期望的格式是这样的：
> \# HELP server_latency Latency of zbase server operations  
> \# TYPE server_latency gauge  
> server_latency{db_id="112",shard_id="1",method="fetch",stage="next",quantile="99"} 2816  
> server_latency{db_id="112",shard_id="1",method="fetch",stage="next",quantile="999"} 2816  
> server_latency{db_id="112",shard_id="1",method="fetch",stage="next",quantile="9999"} 2816  
[...]

PR #2870, opened 2025-01-13
Fix/metric emission tracking #2825

snehavats1404:

Issue Number: #2825

Problem Summary:
The issue pertains to incorrectly tracking and emitting metric names when dumping Prometheus metrics. Some metrics, especially those related to latency, were not being properly emitted due to incorrect handling of the `emitted_metrics` set.

Changed:
- Fixed the logic for tracking emitted metrics by using a set to ensure that duplicate metrics are not emitted.
- Ensured that all necessary latency percentiles and summary metrics are properly output, with correct handling of the `_latency_*` and `_count` suffixes.
- Added a mechanism to track previously emitted metric names to prevent duplication in Prometheus output.

Side effects:
- **Performance effects**: The introduction of a set to track emitted metrics could slightly impact performance, but this change ensures more efficient handling of duplicate metrics.
- **Breaking backward compatibility**: No breaking changes; existing behavior should remain intact except for the fix to prevent duplicate metric emission.

### DIFF
--- a/src/brpc/builtin/prometheus_metrics_service.cpp
+++ b/src/brpc/builtin/prometheus_metrics_service.cpp
@@ -80,6 +80,7 @@ private:
     butil::IOBufBuilder* _os;
     const std::string _server_prefix;
     std::map<std::string, SummaryItems> _m;
+    std::set<std::string> emitted_metrics;//added to track emitted metric names
 };
 
 butil::StringPiece GetMetricsName(const std::string& name) {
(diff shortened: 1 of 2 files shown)


chenBright:
If the bvar of mbvar is not continuously input into the dump, is there a problem?

snehavats1404:
I don't feel there must be any issue in not adding any unit testcase, if there is any please let me know.

chenBright:
> I don't feel there must be any issue in not adding any unit testcase, if there is any please let me know.

We hope that PR will be equipped with unit tests as much as possible to ensure correctness.

chenBright:
closed this as completed in #2899 .

PR #2959, opened 2025-04-25
use standard alg to optimze comparator

zhangqiongyu:

Issue Number:

Problem Summary:

Changed: Use standard library functions to optimize comparison operators, cache path objects, reduce the number of pointer indirect addressing operations, and streamline code.

Side effects:
- Performance effects: no

---
### Check List:

### DIFF
--- a/src/brpc/restful.cpp
+++ b/src/brpc/restful.cpp
@@ -314,29 +314,22 @@ void RestfulMap::ClearMethods() {
 struct CompareItemInPathList {
     bool operator()(const RestfulMethodProperty* e1,
                     const RestfulMethodProperty* e2) const {
-        const int rc1 = e1->path.prefix.compare(e2->path.prefix);
+        const RestfulMethodPath& path1 = e1->path;
+        const RestfulMethodPath& path2 = e2->path;
+        const int rc1 = path1.prefix.compare(path2.prefix);
         if (rc1 != 0) {
             return rc1 < 0;
         }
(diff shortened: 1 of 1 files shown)


wwbmmm:
LGTM

chenBright (approved):
LGMT

src/brpc/restful.cpp
```diff
@@ -314,29 +314,22 @@ void RestfulMap::ClearMethods() {
...
-               it2 != e2->path.postfix.rend()) {
-            if (*it1 != *it2) {
-                return (*it1 < *it2);
-            }
-            ++it1;
-            ++it2;
-        }
-        return (it1 == e1->path.postfix.rend())
-            > (it2 == e2->path.postfix.rend());
+        const bool postfix_result = std::lexicographical_compare(
```
wasphin, line 329: [这里](https://en.cppreference.com/w/cpp/algorithm/lexicographical_compare) 看依赖标准 C++17/C++20，如果确实要变需要同步更新相关编译说明了。

chenBright: I saw in the document([C++ Standard Library Defect Report List](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2008/n2495.html)) that it seems that c++98 has introduced std::lexicographical_compare.

wasphin: You are right, `constexpr since C++20`, not the function itself.

PR #2964, opened 2025-05-07
Support channel level application health check options

zhoukangsheng:
**Current Application-Layer Health Check Mechanism**
The current application-layer health check mechanism is configured via the -health_check_path flag. If this flag is left empty, the health check is disabled. However, since gflag is a global variable, once set, it enables the application-layer health check for all downstream services accessed by the service.

For complex services that may interact with dozens or even hundreds of downstream services (some of which might not be brpc-based and could instead be Go/Java services), not all downstream services necessarily support health check requests. Even for those that do, the health check request paths (URLs) may vary across services. Therefore, it is necessary to support channel-level configuration for application-layer health check parameters.

Issue Number:

Problem Summary:

no

Changed:
**About This PR**
This PR maintains backward compatibility with the existing gflag-based configuration for health check paths and timeouts while introducing support for setting health check parameters via members of ChannelOptions (hc_option). Parameters in ChannelOptions (hc_option) take precedence over the global gflag settings. If a parameter is not specified in ChannelOptions, the corresponding gflag value will be used as the fallback.

Side effects:
- Performance effects:  no
- Breaking backward compatibility:  no
---
### Check List:

### DIFF
--- a/src/brpc/channel.cpp
+++ b/src/brpc/channel.cpp
@@ -42,6 +42,13 @@ namespace brpc {
 
 DECLARE_bool(enable_rpcz);
 DECLARE_bool(usercode_in_pthread);
+DEFINE_string(health_check_path, "", "Http path of health check call."
+    "By default health check succeeds if the server is connectable."
+    "If this flag is set, health check is not completed until a http "
+    "call to the path succeeds within -health_check_timeout_ms(to make "
+    "sure the server functions well).");
+DEFINE_int32(health_check_timeout_ms, 500, "The timeout for both establishing "
+    "the connection and the http call to -health_check_path over the connection");
 
 ChannelOptions::ChannelOptions()
     : connect_timeout_ms(200)
@@ -58,6 +65,7 @@ ChannelOptions::ChannelOptions()
     , backup_request_policy(NULL)
     , retry_policy(NULL)
     , ns_filter(NULL)
+    , health_check_timeout_ms(-1)
 {}
 
 ChannelSSLOptions* ChannelOptions::mutable_ssl_options() {
@@ -172,6 +180,13 @@ int Channel::InitChannelOptions(const ChannelOptions* options) {
     if (NULL == protocol || !protocol->support_client()) {
         LOG(ERROR) << "Channel does not support the protocol";
         return -1;
+    }
+
+    if (_options.health_check_path.empty()) {
+        _options.health_check_path = FLAGS_health_check_path;
+    }
+    if (_options.health_check_timeout_ms < 0) {
+        _options.health_check_timeout_ms = FLAGS_health_check_timeout_ms;
     }
 
     if (_options.use_rdma) {
@@ -348,8 +363,13 @@ int Channel::InitSingle(const butil::EndPoint& server_addr_and_port,
     if (CreateSocketSSLContext(_options, &ssl_ctx) != 0) {
         return -1;
     }
+    HealthCheckOption hc_option;
+    hc_option.health_check_path = _options.health_check_path;
+    if (_options.health_check_timeout_ms > 0) {
+        hc_option.health_check_timeout_ms = _options.health_check_timeout_ms;
+    }
     if (SocketMapInsert(SocketMapKey(server_addr_and_port, sig),
-                        &_server_id, ssl_ctx, _options.use_rdma) != 0) {
+                        &_server_id, ssl_ctx, _options.use_rdma, hc_option) != 0) {
         LOG(ERROR) << "Fail to insert into SocketMap";
         return -1;
     }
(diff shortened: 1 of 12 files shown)


src/brpc/details/naming_service_thread.h
```diff
@@ -45,11 +45,14 @@ struct GetNamingServiceThreadOptions {
...
         : succeed_without_server(false)
         , log_succeed_without_server(true)
-        , use_rdma(false) {}
+        , use_rdma(false) 
+        , health_check_timeout_ms(500) {}
     
     bool succeed_without_server;
     bool log_succeed_without_server;
     bool use_rdma;
+    int32_t health_check_timeout_ms;
```
wwbmmm, line 54: Can change these two members to a single HealthCheckOption member

src/brpc/channel.h
```diff
@@ -143,6 +143,18 @@ struct ChannelOptions {
     // Default: ""
     std::string connection_group;
 
+    // Set the health check path according to the channel granularity. 
+    // Its priority is higher than FLAGS_health_check_path
+    // When it is not set, FLAGS_health_check_path will take effect
+    // Default: ""
+    std::string health_check_path;
```
wwbmmm, line 150: Can change these two members to a single HealthCheckOption member

src/brpc/socket.h
```diff
@@ -266,6 +266,8 @@ struct SocketOptions {
     // one thread at any time.
     void (*on_edge_triggered_events)(Socket*){NULL};
     int health_check_interval_s{-1};
+    std::string health_check_path;
```
wwbmmm, line 269: Can change these two members to a single HealthCheckOption member

src/brpc/socket_map.h
```diff
@@ -44,6 +44,11 @@ inline bool operator!=(const ChannelSignature& s1, const ChannelSignature& s2) {
     return !(s1 == s2);
 }
 
+struct HealthCheckOption {
+    std::string health_check_path;
+    int32_t health_check_timeout_ms = 500;
```
wwbmmm, line 49: better initialize like this: `int32_t health_check_timeout_ms{500};`

src/brpc/channel.cpp
```diff
@@ -388,6 +408,10 @@ int Channel::Init(const char* ns_url,
     ns_opt.log_succeed_without_server = _options.log_succeed_without_server;
     ns_opt.use_rdma = _options.use_rdma;
     ns_opt.channel_signature = ComputeChannelSignature(_options);
```
wwbmmm, line 410: maybe these hc options should be included in ComputeChannelSignature

wwbmmm:
Thank you for contributing! This is a useful feature.

zhoukangsheng:
> Thank you for contributing! This is a useful feature.
> I have modified the title to English. Since this is an Apache project and we'd better using Engish in PR discussions.

Thank you for your code review very much. I have modified it according to your suggestion. Please check if there is anything else that needs to be modified.

src/brpc/channel.h
```diff
@@ -36,6 +36,7 @@
 #include "brpc/retry_policy.h"
 #include "brpc/backup_request_policy.h"
 #include "brpc/naming_service_filter.h"
+#include "brpc/socket.h"
```
wwbmmm, line 39: I think the `HealthCheckOption`  is better to be defined in a separate header file, like `ChannelSSLOptions` defined in `ssl_options.h`. So that you can avoid importing the unrelated classes and headers in `socket.h` here.

src/brpc/socket.h
```diff
@@ -980,6 +992,8 @@ friend void DereferenceSocket(Socket*);
     int _tcp_user_timeout_ms;
 
     HttpMethod _http_request_method;
+    std::string _health_check_path;
```
wwbmmm, line 995: These can be replaced into `HealthCheckOption` too.

src/brpc/details/health_check.cpp
```diff
@@ -65,6 +57,8 @@ class OnAppHealthCheckDone : public google::protobuf::Closure {
     SocketId id;
     int64_t interval_s;
     int64_t last_check_time_ms;
+    int32_t health_check_timeout_ms;
```
wwbmmm, line 60: These can be replaced into HealthCheckOption too.

src/brpc/channel.h
zhoukangsheng: done

zhoukangsheng:
ut
The implementation of this UT appears to fail randomly. Increasing the number of executions of SelectServer can, to some extent, make the probability of IP selection closer to the proportion of the set weight to the total weight.

wwbmmm:
LGTM

PR #2965, opened 2025-05-07
bugfix: extend block pool runtime

yanglimingcn:

Issue Number:

Problem Summary:

rdma block pool runtime ExtendBlockPool has some error:
1. a region may has many buckets, extend a region not hold all bucket locks, will case race condition, disable runtime extend a region when buckets greater than 1.
2. when new region created, link the region to idle_list case link list break.
3. add feature, in scenarios where users need to manually specify memory regions (e.g., using hugepages or custom memory pools), when FLAGS_rdma_memory_pool_user_specified_memory is true, user is  responsibility of extending memory blocks , this ensuring flexibility for advanced use cases.

Changed:

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/brpc/rdma/block_pool.cpp
+++ b/src/brpc/rdma/block_pool.cpp
@@ -36,9 +36,12 @@ DEFINE_int32(rdma_memory_pool_initial_size_mb, 1024,
              "Initial size of memory pool for RDMA (MB)");
 DEFINE_int32(rdma_memory_pool_increase_size_mb, 1024,
              "Increased size of memory pool for RDMA (MB)");
-DEFINE_int32(rdma_memory_pool_max_regions, 4, "Max number of regions");
+DEFINE_int32(rdma_memory_pool_max_regions, 1, "Max number of regions");
 DEFINE_int32(rdma_memory_pool_buckets, 4, "Number of buckets to reduce race");
 DEFINE_int32(rdma_memory_pool_tls_cache_num, 128, "Number of cached block in tls");
+DEFINE_bool(rdma_memory_pool_user_specified_memory, false,
+            "If true, the user must call UserExtendBlockPool() to extend "
+            "memory. bRPC will not handle memory extension.");
 
 static RegisterCallback g_cb = NULL;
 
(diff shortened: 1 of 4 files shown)


yanglimingcn:
@wwbmmm @chenBright  please review this bugfix

wwbmmm:
LGTM

src/brpc/rdma/block_pool.cpp
```diff
@@ -178,23 +163,79 @@ static void* ExtendBlockPool(size_t region_size, int block_type) {
...
+        PLOG_EVERY_SECOND(ERROR) << "Memory not enough";
+        return NULL;
+    }
+
+    return ExtendBlockPoolImpl(region_base, region_size, block_type);
+}
+
+void* ExtendBlockPoolByUser(void* region_base, size_t region_size,
+                            int block_type) {
+    if (FLAGS_rdma_memory_pool_user_specified_memory == false) {
```
chenBright, line 205: `!FLAGS_rdma_memory_pool_user_specified_memory`？

yanglimingcn: FLAGS_rdma_memory_pool_user_specified_memory == false, means not allowed to ExtentBlockPoolByUser

AnDiXL:
> ### What problem does this PR solve?
> Issue Number:
> 
> Problem Summary:
> 
> rdma block pool runtime ExtendBlockPool has some error:
> 
> 1. a region may has many buckets, extend a region not hold all bucket locks, will case race condition, disable runtime extend a region when buckets greater than 1.
> 2. when new region created, link the region to idle_list case link list break.
> 3. add feature, in scenarios where users need to manually specify memory regions (e.g., using hugepages or custom memory pools), when FLAGS_rdma_memory_pool_user_specified_memory is true, user is  responsibility of extending memory blocks , this ensuring flexibility for advanced use cases.
> 
> ### What is changed and the side effects?
> Changed:
> 
> Side effects:
> 
> * Performance effects:
> * Breaking backward compatibility:
> 
> ### Check List:
> * Please make sure your changes are compilable.
> * When providing us with a new feature, it is best to add related tests.
> * Please follow [Contributor Covenant Code of Conduct](https://github.com/apache/brpc/blob/master/CODE_OF_CONDUCT.md).

“a region may has many buckets, extend a region not hold all bucket locks, will case race condition, disable runtime extend a region when buckets greater than 1.”

Why not modify it so that a lock is added to the bucket when expanding the region? In this way, there would be no need to limit the number of buckets. 
For example, in the following form:

```c++
    for (size_t i = 0; i < g_buckets; ++i) {
        node[i]->start = (void*)(region->start + i * (region_size / g_buckets));
        node[i]->len = region_size / g_buckets;

        BAIDU_SCOPED_LOCK(*g_info->lock[block_type][i]);
        node[i]->next = g_info->idle_list[block_type][i];
        g_info->idle_list[block_type][i] = node[i];
        g_info->idle_size[block_type][i] += node[i]->len;
    }
```

PR #2967, opened 2025-05-12
baidu_std support checksum

yanglimingcn:

Issue Number: #2159

Problem Summary:

Implementing Checksum Functionality at the bRPC Level:

General Requirement: Similar to the compression feature, implement a unified solution at the bRPC level.
Better Performance: Implementing checksum at the user level requires users to serialize protobuf or other data packets and then calculate the checksum. Implementing it at the bRPC level allows direct use of the serialized results for checksum calculation.
Better Compatibility: Implementing checksum at the user level means users work with protobuf data packets. If the protocol changes and the client and server protocols become inconsistent, checksum calculations based on protobuf will fail, preventing users from leveraging protobuf's compatibility features.

Changed:

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/brpc/checksum.h
+++ b/src/brpc/checksum.h
@@ -1,0 +1,63 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#ifndef BRPC_CHECKSUM_H
+#define BRPC_CHECKSUM_H
+
+#include "brpc/controller.h"
+#include "brpc/options.pb.h"  // ChecksumType
+#include "butil/iobuf.h"      // butil::IOBuf
+
+namespace brpc {
+
+struct ChecksumIn {
+    const butil::IOBuf* buf;
+    Controller* cntl;
+};
+
+struct ChecksumHandler {
+    // checksum `buf'.
+    // Returns checksum value
+    void (*Compute)(const ChecksumIn& in);
+
+    // verify buf checksum
+    // Rerturn true on success, false otherwise
+    bool (*Verify)(const ChecksumIn& in);
+
+    // Name of the checksum algorithm, must be string constant.
+    const char* name;
+};
+
+// [NOT thread-safe] Register `handler' using key=`type'
+// Returns 0 on success, -1 otherwise
+int RegisterChecksumHandler(ChecksumType type, ChecksumHandler handler);
+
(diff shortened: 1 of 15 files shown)


yanglimingcn:
image
这个协议的checksum_value字段目前定义为uint32，可能后续扩展性不太好，大家有什么建议。

yanglimingcn:
@wwbmmm @chenBright 辛苦看看，给点建议。

wwbmmm:
>  这个PR中协议的checksum_value字段目前定义为uint32，可能后续扩展性不太好，大家有什么建议。

maybe define checksum_value as bytes in proto, as std::string in code?

yanglimingcn:
> > 这个PR中协议的checksum_value字段目前定义为uint32，可能后续扩展性不太好，大家有什么建议。
> 
> maybe define checksum_value as bytes in proto, as std::string in code?

done

wwbmmm:
LGTM

src/brpc/policy/crc32c_checksum.cpp
```diff
@@ -0,0 +1,59 @@
...
+    butil::IOBufAsZeroCopyInputStream wrapper(buf);
+    const void* data;
+    int size;
+    uint32_t crc = 0;
+    while (wrapper.Next(&data, &size)) {
+        crc = butil::crc32c::Extend(crc, static_cast<const char*>(data), size);
+    }
+    auto& val = ControllerPrivateAccessor(const_cast<Controller*>(&cntl))
+                    .checksum_value();
+    auto expected = *reinterpret_cast<const uint32_t*>(val.data());
```
chenBright, line 52: Do I need to check the length of val? If the length is less than sizeof(uint32_t), there will be a memory out-of-bounds problem.

src/brpc/checksum.h
```diff
@@ -0,0 +1,62 @@
...
+namespace brpc {
+
+struct ChecksumHandler {
+    // checksum `buf'.
+    // Returns checksum value
+    void (*Compute)(const butil::IOBuf& buf, Controller* cntl);
+
+    // verify buf checksum
+    // Rerturn true on success, false otherwise
+    bool (*Verify)(const butil::IOBuf& buf, const Controller& cntl);
```
chenBright, line 36: Do we need to design parameters to be more extensible? Like #2946, the parameters may be expanded in the future.

src/brpc/policy/crc32c_checksum.cpp
yanglimingcn: val is assigned by a uint32_t, so it is length
will always 4 bytes

src/brpc/checksum.h
chenBright: For example, use struct instead of all parameters:

```c++
struct In {
    const butil::IOBuf& buf; 
    Controller* cntl
};
void (*Compute)(const In& in);
bool (*Verify)(const In& in);
```

src/brpc/policy/crc32c_checksum.cpp
chenBright: I think if the payload is likely to be modified, the val string is also likely to be modified, so it is safer to check the length of val.

src/brpc/checksum.h
```diff
@@ -0,0 +1,65 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#ifndef BRPC_CHECKSUM_H
+#define BRPC_CHECKSUM_H
+
+#include <google/protobuf/message.h>  // Message
```
chenBright: Redundant header?

Issue #2983, opened 2025-06-06
bthread task tracer 存在死锁问题


CodeZHXS:
**Describe the bug**
在 `TaskTracer::TraceImpl` 的实现中，使用 `m->traced = true` 表示当前 bthread 正在回溯。
正常的流程是，在回溯完毕后，会将这个值设置成 `false`，随后唤醒条件变量。
但是，如果遇到了不正确的状态，例如 `TASK_STATUS_UNKNOWN`、`TASK_STATUS_CREATED`、`TASK_STATUS_FIRST_READY`，这个值没有被设置回来。这会导致 `TaskTracer::WaitForTracing` 等待的线程永远不会被唤醒。
如下图所示：
![Image](https://github.com/user-attachments/assets/81878060-1175-448d-a0d8-aecbb269b0f2)

![Image](https://github.com/user-attachments/assets/559f0dfd-ed0e-4a80-a01b-9cb2ff9dbb23)

**To Reproduce**
我的分支：https://github.com/CodeZHXS/brpc/tree/task_tracer_bug_report
我在 1.12.1 版本之上进行了少量的改动，以便让这个问题可以稳定复现。可以直接 clone 我的分支复现。

具体来说，我会在第一次触发 `TaskTracer::TraceImpl` 时视为一个错误的状态，随后 `TaskTracer::TraceImpl` 会立刻返回。

![Image](https://github.com/user-attachments/assets/81867b6f-8112-4d2b-a1d7-e6e21a318a9e)

我使用了 example/bthread_tag_echo_c++ 这个示例程序复现，步骤如下：
```bash
cd example/bthread_tag_echo_c++/build
cmake -B build -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DWITH_BTHREAD_TRACER=ON
cmake --build build -j30
cd build
./echo_server -task_group_ntags 3 -tag1 0 -tag2 1 -bthread_concurrency 20 -bthread_min_concurrency 8 -event_dispatcher_num 1
```

随后命令行会打印 server 的其中一个 bthread_id，将其复制到浏览器中打印调用堆栈，会得到一个错误提示：
![Image](https://github.com/user-attachments/assets/079c6d0b-9bb6-4825-8b83-ee66545631ea)

然后，server 彻底卡住，使用 gdb 调试可以看到卡在 `TaskTracer::WaitForTracing` 中。

**Expected behavior**
即使状态不对，返回之后也不应该影响这个 bthread 的运行。

**Versions**
[...]

PR #2988, opened 2025-06-09
Fix bthread task tracer deadlock bug (issue #2983)

CodeZHXS:

Issue Number: #2983 

Problem Summary: `TaskTracer::TraceImpl`, did not set the relevant variables(`m->traced`) on all return paths, which would cause a deadlock.

use `BRPC_SCOPE_EXIT` to ensure `m->traced==false` on all return paths.

Changed: src/bthread/task_tracer.cpp `TaskTracer::TraceImpl()` function

Side effects:
- Performance effects: No

- Breaking backward compatibility: No

---


chenBright (approved):
LGTM

PR #2994, opened 2025-06-12
Consistent hashing support server tag

egolearner:

Issue Number:2993

Problem Summary:

Changed:
1. 在判断Node是否相同时，依次比较hash/server_addr/tag
2. 在生成hash时，包含tag。

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/brpc/policy/consistent_hashing_load_balancer.cpp
+++ b/src/brpc/policy/consistent_hashing_load_balancer.cpp
@@ -33,6 +33,8 @@ namespace policy {
 // TODO: or 160?
 DEFINE_int32(chash_num_replicas, 100, 
              "default number of replicas per server in chash");
+DEFINE_bool(consistent_hashing_enable_server_tag, false, 
+             "if consistent hashing enable server with tag");
 
 // Defined in hasher.cpp.
 const char* GetHashName(HashFunc hasher);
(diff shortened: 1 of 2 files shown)


egolearner:
fix #2993

src/brpc/policy/consistent_hashing_load_balancer.cpp
```diff
@@ -71,8 +71,8 @@ bool DefaultReplicaPolicy::Build(ServerId server,
     replicas->clear();
     for (size_t i = 0; i < num_replicas; ++i) {
         char host[256];
-        int len = snprintf(host, sizeof(host), "%s-%lu",
-                           endpoint2str(ptr->remote_side()).c_str(), i);
+        int len = snprintf(host, sizeof(host), "%s-%lu-%s",
```
chenBright: How to handle the situation where the length exceeds 256?

egolearner:
@chenBright @wwbmmm please take a look

wwbmmm:
Thank you for your contribution! I suggest to add a flag for backward compatibility: If the flag is on, add server tag for hash computing, if the flag is off, remain the original logic. Since consistent-hashing is always used in senarios that the servers are stateful, changing hashing logic may cause the request routed to new server that does not contain previous data for that request.

src/brpc/policy/consistent_hashing_load_balancer.cpp
egolearner: 1. with snprintf，no buffer overflow problem.
2. If total length exceeds 256
    1. if prefix of tag is different, we will generate a different hash probably
    2. if prefix is the same, we will generate the same hash, which should already be handled as hash conflict.

In short, no big deal if total length exceeds 256.

egolearner:
> Thank you for your contribution! I suggest to add a flag for backward compatibility: If the flag is on, add server tag for hash computing, if the flag is off, remain the original logic. Since consistent-hashing is always used in senarios that the servers are stateful, changing hashing logic may cause the request routed to new server that does not contain previous data for that request.

done

egolearner:
@chenBright @wwbmmm please take a look， thx

chenBright (approved):
LGTM

PR #3002, opened 2025-06-19
Fix the bug that the grpc protocol is incompatible with authentication(issue:#3001)

Paragrf:
…y could not be parsed by the client to get the correct error code after authentication failed

Issue Number: #3001 

Problem Summary:

Changed:

- grpc协议和http协议当前共用同一个鉴权函数，所以在鉴权函数中增加了对http协议类型的判断，如果是http2.0版本，按grpc格式返回错误

Side effects:

BossHi-20250619160635

- Performance effects:

---
### Check List:

### DIFF
--- a/src/brpc/policy/http_rpc_protocol.cpp
+++ b/src/brpc/policy/http_rpc_protocol.cpp
@@ -1331,7 +1331,26 @@ ParseResult ParseHttpMessage(butil::IOBuf *source, Socket *socket,
     }
 }
 
-static void SendUnauthorizedResponse(const std::string& user_error_text, Socket* socket) {
+static void SendUnauthorizedResponse(const std::string& user_error_text, Socket* socket, const InputMessageBase* msg) {
+    HttpContext* http_request = (HttpContext*)msg;
+    const bool is_http2 = http_request->header().is_http2();
+    if (is_http2) {
+        // for grpc client
+        const H2StreamContext* h2_sctx = static_cast<const H2StreamContext*>(msg);
+        brpc::Controller* cntl = new brpc::Controller();
+        cntl->http_response().set_status_code(200);
+        cntl->http_response().set_content_type("application/grpc");
+        cntl->SetFailed(ERPCAUTH, "%s", user_error_text.empty() ? "Fail to authenticate" : user_error_text.c_str());
+
+        SocketMessagePtr<H2UnsentResponse> h2_response(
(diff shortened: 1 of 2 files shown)


Paragrf:
Close #3001

Paragrf:
@chenBright  please take a look ：）

chenBright:
Please add unit test for this case in test/brpc_server_unittest.cpp.

Paragrf:
> Please add unit test for this case in test/brpc_server_unittest.cpp.

@chenBright done

Paragrf:
@chenBright  @wwbmmm  please take a look， thx

src/brpc/policy/http_rpc_protocol.cpp
```diff
@@ -1331,7 +1331,26 @@ ParseResult ParseHttpMessage(butil::IOBuf *source, Socket *socket,
...
 }
 
-static void SendUnauthorizedResponse(const std::string& user_error_text, Socket* socket) {
+static void SendUnauthorizedResponse(const std::string& user_error_text, Socket* socket, const InputMessageBase* msg) {
+    HttpContext* http_request = (HttpContext*)msg;
+    const bool is_http2 = http_request->header().is_http2();
+    if (is_http2) {
+        // for grpc client
+        const H2StreamContext* h2_sctx = static_cast<const H2StreamContext*>(msg);
+        brpc::Controller* cntl = new brpc::Controller();
```
chenBright: I think no need to new Controller.

Paragrf: Sorry, I didn't understand you. I notice that line 1411 in ProcessHttpRequest is also handled in this way. Why is there no need for a new Controller here?

Paragrf: @chenBright  A cntl object is indeed needed to set the error code and message for authentication failures. Are you suggesting that we should avoid using a pointer here, and instead simply declare it as "Controller cntl = Controller()";, thereby also eliminating the need to call "delete cntl"

chenBright: Yeah, `Controller cntl;` already meets the need.

Paragrf: @chenBright  DONE

Paragrf:
@chenBright  hello, can this PR be merged

PR #3008, opened 2025-06-26
optimize the fallback code style

zhangqiongyu:

Issue Number:

Problem Summary:

Changed:
Engineering equivalent implementation, when I need to use multiple versions of ibverbs.so in actual applications, it is easier to expand.
Side effects:
- Performance effects: no

---
### Check List:


TousakaRin:
Does LD_LIBRARY_PATH meet your needs?

zhangqiongyu:
> Does LD_LIBRARY_PATH meet your needs?

yes，not only LD_LIBRARY_PATH  but also soft link and renaming so can meet my needs. I just want to make this snippet more extensible.

PR #3009, opened 2025-06-28
Bugfix: bthread_worker_usage could exceed bthread_worker_count

### LINKED ISSUE #1231: bthread_worker_usage超出bthread_worker_count
**Describe the bug (描述bug)**
应用场景是dmkit当发起一个调用unit的请求时，bthread_worker_usage有时会超出bthread_worker_count很多，
当bthread_worker_count设置为9的时候，bthread_worker_usage最大会跳到100多，
当bthread_worker_count设置为48的时候，bthread_worker_usage最大会跳到1200多，
然后就会一直timeout，无论请求多少次。
此时如果过一会不管他，再次请求就好了。。。

**To Reproduce (复现方法)**
偶尔的访问调用会触发 bthread_worker_usage 的爆炸式增长，正常情况下是不会增长太多的，bthread_worker_usage也就长个1左右

**Expected behavior (期望行为)**
bthread_worker_usage波动正常，不会出现跳跃

**Versions (各种版本)**
OS: centos 7
Compiler: 
brpc: commit 2ae7f04c
protobuf: 2.5.0

WeChatbfc8219ad99a001851121a190aacd2ff

### LINKED ISSUE #2640: bthread_worker_usage这个统计值为何有时候会超时bthread
**Describe the bug (描述bug)**
在brpc的统计页面上，很容易就发现bthread_worker_usage的值
超过bthread_worker_count,理论上根据代码，bthread_worker_usage
的值不会超过TaskGroup的个数，因为代码
double TaskControl::get_cumulated_worker_time() {
    int64_t cputime_ns = 0;
    BAIDU_SCOPED_LOCK(_modify_group_mutex);
    const size_t ngroup = _ngroup.load(butil::memory_order_relaxed);
    for (size_t i = 0; i < ngroup; ++i) {
        if (_groups[i]) {
            cputime_ns += _groups[i]->_cumulated_cputime_ns;
        }
    }
    return cputime_ns / 1000000000.0;
}

这个函数是一秒钟统计一次，如果定时器完全准的话，一定不会超过ngroup的个数
因此这个值是否强依赖定时器，因而不一定完全准确，是否应该将
 cputime_ns / 1000000000.0;
更改为 cputime_ns / （real_elapse_time）;其中real_elapse_time是这个函数两次
调用的时间差。

**To Reproduce (复现方法)**

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
brpc: 1.0.0
protobuf:

image

### LINKED ISSUE #2972: vars 监控中bthread_worker_usage > bthread_worker_count 是正常的么
如下图所示 bthread_worker_usage > bthread_worker_count , 正在使用的不应该<=上限么，这里是统计口径问题么？

Image

chenBright:

Issue Number: resolve #1231 , resolve #2640 , resolve #2972

Problem Summary:

Changed:

`bthread_worker_usage` should consist of `_cumulated_cputime_ns` and the CPU time of the task currently run by the worker.

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/bthread/task_control.cpp
+++ b/src/bthread/task_control.cpp
@@ -185,6 +185,7 @@ TaskControl::TaskControl()
     , _nbthreads("bthread_count")
     , _priority_queues(FLAGS_task_group_ntags)
     , _pl(FLAGS_task_group_ntags)
+    , _last_get_cumulated_time_ns(0)
 {}
 
 int TaskControl::init(int concurrency) {
@@ -524,12 +525,29 @@ void TaskControl::print_rq_sizes(std::ostream& os) {
 
 double TaskControl::get_cumulated_worker_time() {
     int64_t cputime_ns = 0;
+    int64_t now = butil::cpuwide_time_ns();
     BAIDU_SCOPED_LOCK(_modify_group_mutex);
     for_each_task_group([&](TaskGroup* g) {
         if (g) {
-            cputime_ns += g->_cumulated_cputime_ns;
+            // With the acquire-release atomic operation, the CPU time of the bthread is
+            // only calculated once through `_cumulated_cputime_ns' or `_last_run_ns'.
+            cputime_ns += g->_cumulated_cputime_ns.load(butil::memory_order_acquire);
+            // The bthread is still running on the worker,
+            // so we need to add the elapsed time since it started.
+            // In extreme cases, before getting `_last_run_ns_in_tc',
+            // `_last_run_ns_in_tc' may have been updated multiple times,
+            // and `cputime_ns' will miss some cpu time, which is ok.
+            int64_t last_run_ns = g->_last_run_ns;
+            if (last_run_ns > _last_get_cumulated_time_ns) {
+                g->_last_run_ns_in_tc = last_run_ns;
+                cputime_ns += now - last_run_ns;
+            } else if (last_run_ns == g->_last_run_ns_in_tc) {
+                // The bthread is still running on the same worker.
+                cputime_ns += now - last_run_ns;
+            }
         }
     });
+    _last_get_cumulated_time_ns = now;
     return cputime_ns / 1000000000.0;
 }
 
--- a/src/bthread/task_control.cpp
+++ b/src/bthread/task_control.cpp
@@ -533,7 +533,7 @@ double TaskControl::get_cumulated_worker_time() {
     int64_t cputime_ns = 0;
     BAIDU_SCOPED_LOCK(_modify_group_mutex);
     for_each_task_group([&](TaskGroup* g) {
-        cputime_ns += get_cumulated_worker_time(g);
+        cputime_ns += g->cumulated_cputime_ns();
     });
     return cputime_ns / 1000000000.0;
 }
@@ -544,38 +544,9 @@ double TaskControl::get_cumulated_worker_time(bthread_tag_t tag) {
     const size_t ngroup = tag_ngroup(tag).load(butil::memory_order_relaxed);
     auto& groups = tag_group(tag);
     for (size_t i = 0; i < ngroup; ++i) {
-        cputime_ns += get_cumulated_worker_time(groups[i]);
+        cputime_ns += groups[i]->cumulated_cputime_ns();
     }
     return cputime_ns / 1000000000.0;
-}
-
-double TaskControl::get_cumulated_worker_time(TaskGroup* g) {
-    if (NULL == g) {
-        return 0.0;
-    }
-
-#if __x86_64__ || __ARM_NEON
-#ifdef __x86_64__
-    __m128i cpu_time_stat = _mm_load_si128(reinterpret_cast<__m128i*>(&g->_cpu_time_stat));
-#else // __ARM_NEON
-    int64x2_t cpu_time_stat = vld1q_s64(reinterpret_cast<const int64_t*>(&g->_cpu_time_stat));
-#endif // __x86_64__
-    int64_t last_run_ns = cpu_time_stat[0];
-    int64_t cputime_ns = cpu_time_stat[1];
-#else // __x86_64__ || __ARM_NEON
-    int64_t last_run_ns = 0;
-    int64_t cputime_ns = 0;
-    {
-        BAIDU_SCOPED_LOCK(g->_cpu_time_stat_mutex);
-        last_run_ns = g->_cpu_time_stat.last_run_ns;
-        cputime_ns = g->_cpu_time_stat.cumulated_cputime_ns;
-    }
-#endif // __x86_64__ || __ARM_NEON
-    // Add the elapsed time of running bthread.
-    if (last_run_ns > 0) {
-        cputime_ns += butil::cpuwide_time_ns() - last_run_ns;
-    }
-    return cputime_ns;
 }
 
 int64_t TaskControl::get_cumulated_switch_count() {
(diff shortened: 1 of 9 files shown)


src/bthread/task_control.cpp
```diff
@@ -524,12 +525,29 @@ void TaskControl::print_rq_sizes(std::ostream& os) {
...
+            // `_last_run_ns_in_tc' may have been updated multiple times,
+            // and `cputime_ns' will miss some cpu time, which is ok.
+            int64_t last_run_ns = g->_last_run_ns;
+            if (last_run_ns > _last_get_cumulated_time_ns) {
+                g->_last_run_ns_in_tc = last_run_ns;
+                cputime_ns += now - last_run_ns;
+            } else if (last_run_ns == g->_last_run_ns_in_tc) {
+                // The bthread is still running on the same worker.
+                cputime_ns += now - last_run_ns;
+            }
```
gitccl: This change may introduce a subtle race condition when sampling cumulative CPU time.
Suppose `bthread_worker_usage` starts sampling at T=0s, and a task also starts execution at the same time. The task keeps running continuously for 3s. According to the current implementation, `get_cumulated_worker_time()` will return approximately 0, 1, and 2 at T=0s, T=1s, and T=2s respectively — which is expected.

However, at T=3s, a race can occur as follows:
- The sampling thread first reads `_cumulated_cputime_ns`, which is still 0.
- Then, the worker thread updates `_last_run_ns` and `_cumulated_cputime_ns` to 3e9.
- The sampling thread then proceeds to read `_last_run_ns` = 3e9.
- Since `last_run_ns` > `_last_get_cumulated_time_ns`, it goes into the branch:
```
cputime_ns += now - last_run_ns;
```
But `now` was captured earlier and is slightly less than `last_run_ns`, resulting in a **negative delta**.
As a result, the sampling thread misses the 3 seconds of accumulated CPU time, causes `get_cumulated_worker_time()` at T=3s to return a smaller value than at T=2s(**maybe even return a negative value**), finally leads to a negative value in `bthread_worker_usage`.

To prevent this issue, `_cumulated_cputime_ns` and `_last_run_ns` should be read together under a consistent view. One possible approach without using a lock maybe to read `_last_run_ns` **before and after** reading `_cumulated_cputime_ns`, and verify that both `_last_run_ns` reads return the same value.

chenBright: The problem of negative delta ​​was previously considered, but was missed during implementation.

chenBright: > read `_last_run_ns` **before and after** reading `_cumulated_cputime_ns`, and verify that both `_last_run_ns` reads return the same value.

I guess your intention is to try to read `_cumulated_cputime_ns` again when the `_last_run_ns` read twice are different. However, this approach cannot guarantee that the value of `_cumulated_cputime_ns` read the second time is 3e9, and it may still read 0.

chenBright: https://rigtorp.se/isatomic/
> On the latest CPU microarchitectures (Skylake and Zen 2) AVX/AVX2 128b/256b aligned loads and stores are atomic even though Intel and AMD officially doesn’t guarantee this. 

Perhaps we can use the implicit feature of modern CPUs that 128-bit aligned load/store operations are atomic to implement atomic operations on combination of `_cumulated_cputime_ns` and `_last_run_ns`.

x86_64: `_mm_load_si128`, `_mm_store_si128`
Arm and later: `vld1q_s64`

For scenarios that do not support 128-bit atomic operations, use the lock to guarantee this. Only the worker thread and sample thread acquire the lock, and the sample thread acquires locks once a second. There is no lock contention between worker threads. So lock contention can be ignored.

gitccl: > > read `_last_run_ns` **before and after** reading `_cumulated_cputime_ns`, and verify that both `_last_run_ns` reads return the same value.
> 
> I guess your intention is to try to read `_cumulated_cputime_ns` again when the `_last_run_ns` read twice are different. However, this approach cannot guarantee that the value of `_cumulated_cputime_ns` read the second time is 3e9, and it may still read 0.

Yeah, you're right, it's still possible to read 0...

chenBright:
@wwbmmm @yanglimingcn Please take a look.

chenBright:
@wwbmmm @yanglimingcn Friendly ping!

src/bthread/task_control.cpp
```diff
@@ -526,26 +533,51 @@ double TaskControl::get_cumulated_worker_time() {
...
     }
     return cputime_ns / 1000000000.0;
 }
 
+double TaskControl::get_cumulated_worker_time(TaskGroup* g) {
+    if (NULL == g) {
+        return 0.0;
+    }
+
+#if __x86_64__ || __ARM_NEON
```
wwbmmm: Maybe wrap these code in a data structure like 'AtomicInteger128' for better maintainability

src/bthread/task_group.h
```diff
@@ -203,6 +206,14 @@ class TaskGroup {
 private:
 friend class TaskControl;
 
+    struct CPUTimeStat {
+        // Last scheduling time.
+        // If this value is negative,
+        // it means that it is the main task.
+        int64_t last_run_ns;
```
wwbmmm: This name is confusing. Maybe change to 'last_run_ns_and_state', and offer reader and writer function like 

```cpp
last_run_ns()
is_main_task()
set_last_run_ns(last_run_ns, is_main_task)
```

src/bthread/task_group.cpp
```diff
@@ -292,7 +282,7 @@ int TaskGroup::init(size_t runqueue_capacity) {
     _cur_meta = m;
     _main_tid = m->tid;
     _main_stack = stk;
-    _last_run_ns = butil::cpuwide_time_ns();
+    _cpu_time_stat.last_run_ns = -m->cpuwide_start_ns;
```
yanglimingcn: why this is -m->cpuwide_start_ns ?

src/bthread/task_control.cpp
chenBright: done

src/bthread/task_group.cpp
chenBright: The higher bit for task type, main task is 1, otherwise 0. Worker usage does not calculate CPU time of the main task.

src/bthread/task_group.h
chenBright: done

chenBright:
```c++
void* test(void*) {
    while (!brpc::IsAskedToQuit()) {
        sleep(2);
        bthread_yield();
    }
}

for (int i = 0; i < 3; ++i) {
    bthread_t tid;
    bthread_start_background(&tid, NULL, test, NULL);
}
```

image


Start 3 bthreads and loop sleep for 2s. The value of bthread_worker_usage is correct, and the extra 1 comes from the epoll bthread.

src/bthread/task_control.cpp
```diff
@@ -33,6 +33,13 @@
 #include "bthread/timer_thread.h"         // global_timer_thread
 #include <gflags/gflags.h>
 #include "bthread/log.h"
+#ifdef __x86_64__
+#include <x86intrin.h>
+#endif // __x86_64__
+
+#ifdef __ARM_NEON
+#include <arm_neon.h>
+#endif // __ARM_NEON
```
wwbmmm: This code can be removed

chenBright: done

wwbmmm:
LGTM

PR #3018, opened 2025-07-06
Make unit tests runnable on macOS

gitccl:

Issue Number:

Problem Summary:
When running brpc unit tests binaries (e.g., test_butil) on macOS, the program exits with an error from gflags:
```
Ignoring RegisterValidateFunction() for flag pointer 0x10961d842: no flag found at that address
```

After investigating, I found that the error occurs at the point where a validator is registered for a flag:
```
DEFINE_bool(crash_on_fatal_log, false,
            "Crash process when a FATAL log is printed");
BUTIL_VALIDATE_GFLAG(crash_on_fatal_log, butil::PassValidate);
```

Using lldb to debug the test binary, I discovered that the `DEFINE_bool` and `BUTIL_VALIDATE_GFLAG` access different instances of gflags:
- `DEFINE_bool` accesses the gflags instance in `test_butil`
- `BUTIL_VALIDATE_GFLAG` accesses the gflags instance in `libbrpc.dbg.dylib`

On macOS, the test binary links against:
- `libbrpc.dbg.dylib` (which already statically links to `libgflags.a`)
- `libgflags.a`

Because macOS uses a two-level namespace model, symbols from separate static libraries are not unified across dynamic libraries and executables. As a result, there are two isolated instances of gflags — one in the test binary, and one in the dylib. Consequently, the flag is registered in one instance, but the validator is registered in the other, leading to registration failure.

To resolve this, we should avoid linking `libgflags.a` (and other static dependencies like protobuf) into the test binary if they are already linked inside `libbrpc.dbg.dylib`. This ensures that all gflags-related macros operate on a single shared instance, and the validator registration works as expected.

Changed:

Side effects:
- Performance effects:

---
### Check List:


chenBright (approved):
LGTM

PR #3026, opened 2025-07-13
Support generics for MultiDimension APIs

chenBright:

Issue Number:

Problem Summary:

The current mbvar get/set operation APIs all require the construction of std::list<std::string>, which requires the allocation of memory and the copying of strings, resulting in noticeable performance overhead.

Changed:

The bvar get/set operation API supports more data structures through generics, thus achieving zero copy. Using `Container<const char* / std::string_view / butil::StringPieces>` can avoid string copying, where Container can be a container such as `std::vector`, `std::set`, `std::map`, etc. Furthermore, `std::array` and `absl::InlinedVector` can be used to avoid dynamic heap memory allocation.

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/docs/cn/mbvar_c++.md
+++ b/docs/cn/mbvar_c++.md
@@ -331,11 +331,13 @@ size_t mbvar_list_exposed(std::vector<std::string>* names) {
 
 多维度统计的实现，主要提供bvar的获取、列举等功能。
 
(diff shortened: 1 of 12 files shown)


wwbmmm:
LGTM

PR #3032, opened 2025-07-19
Bugfix: SignalTrace mode has memory access problem

### LINKED ISSUE #3031: Signal Trace has memory access problem
**Describe the bug**
In `TaskTracer::SignalTrace`, we put a  `Result*` in `SignalSync`. But if `SignalTrace` timeout, `Result` will be delete. This may cause memory access problem in the `SignalHandler` thread. I think it will be good if we store pcs by `std::vector<void *>`

CodeZHXS:

Issue Number: Fix #3031

Problem Summary: SignalTrace mode has memory access problem

Changed: use `Result` instead of `Result *`

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/bthread/task_tracer.cpp
+++ b/src/bthread/task_tracer.cpp
@@ -32,7 +32,6 @@
 
 namespace bthread {
 
-DEFINE_bool(enable_fast_unwind, true, "Whether to enable fast unwind");
 DEFINE_uint32(signal_trace_timeout_ms, 50, "Timeout for signal trace in ms");
 BUTIL_VALIDATE_GFLAG(signal_trace_timeout_ms, butil::PositiveInteger<uint32_t>);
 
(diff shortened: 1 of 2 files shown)


src/bthread/task_tracer.cpp
```diff
@@ -356,15 +356,18 @@ void TaskTracer::SignalHandler(int, siginfo_t* info, void* context) {
...
     }
-    Result* result = signal_sync->result;
+    void *pcs[Result::MAX_TRACE_NUM];
     // Skip the first frame, which is the signal handler itself.
-    result->frame_count = absl::DefaultStackUnwinder(&result->ips[0], NULL,
-                                                     arraysize(result->ips),
-                                                     1, context, NULL);
+    int depth = absl::DefaultStackUnwinder(pcs, NULL, arraysize(pcs), 1,
+                                           context, NULL);
+    signal_sync->pcs.reserve(depth);
```
chenBright: Allocating memory in a signal handler is dangerous. Maybe we should make the type of result in SignalSync Result instead of Result*, just like before.

CodeZHXS: > Allocating memory in a signal handler is dangerous. Maybe we should make the type of result in SignalSync Result instead of Result*, just like before.

fix

src/bthread/task_tracer.h
```diff
@@ -87,12 +87,12 @@ class TaskTracer {
 
     // For signal trace.
     struct SignalSync : public butil::SharedObject {
-        explicit SignalSync(Result* result) : result(result) {}
+        explicit SignalSync() {}
```
chenBright: No need for SignalSync().

src/bthread/task_tracer.cpp
```diff
@@ -356,15 +356,14 @@ void TaskTracer::SignalHandler(int, siginfo_t* info, void* context) {
...
         return;
     }
-    Result* result = signal_sync->result;
     // Skip the first frame, which is the signal handler itself.
-    result->frame_count = absl::DefaultStackUnwinder(&result->ips[0], NULL,
-                                                     arraysize(result->ips),
-                                                     1, context, NULL);
+    signal_sync->result.frame_count = absl::DefaultStackUnwinder(signal_sync->result.ips, NULL, 
+                                    arraysize(signal_sync->result.ips), 1,
+                                           context, NULL);
```
chenBright: Format please.

chenBright (approved):
LGTM

PR #3036, opened 2025-07-23
replace embeded snappy by latest offical version

gulu-goolu:
使用了官方最新的 snappy 替换了 butil 下面的 third_party 中的版本

Issue Number:
https://github.com/apache/brpc/issues/3014

Problem Summary:

Changed:

Side effects:
- Performance effects:

---
### Check List:


chenBright (approved):
I think it is a direct replacement for bRPC's built-in snappy, rather than a third-party dependency. Otherwise, make and cmake will fail to compile.

PR #3037, opened 2025-07-23
add flags for set worker thread name

yanglimingcn:

Issue Number: None

Problem Summary:

In some scenarios, services may prefer to keep worker thread names unchanged, as external monitoring and alerting systems already rely on custom-defined thread names. To accommodate this, we've introduced a gflag to control whether bRPC should modify worker thread names.

When thread name modification is disabled through this configuration, users can alternatively set worker thread names within the **run_worker_startfn** function.

Changed:

Side effects:
- Performance effects:

---
### Check List:


wwbmmm:
LGTM

PR #3047, opened 2025-07-31
when rdma event dispatcher unschedule is true, start ProcessEvent background

yanglimingcn:
when rdma event dispatcher unschedule is true, start ProcessEvent background

Issue Number:

Problem Summary:

In certain scenarios, such as when using SPDK, it is required that the bthread executing SPDK must run on a single thread—that is, in single-threaded polling mode. This means the bthread of the event dispatcher must continuously occupy one worker thread and must not preempt other bthreads. This ensures no scheduling occurs, preventing SPDK from being scheduled onto other worker threads.

No

Changed:

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/brpc/rdma/rdma_helper.h
+++ b/src/brpc/rdma/rdma_helper.h
@@ -91,7 +91,16 @@ bool SupportedByRdma(std::string protocol);
 
 }  // namespace rdma
 }  // namespace brpc
+#else
+namespace brpc {
+namespace rdma {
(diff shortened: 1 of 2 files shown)


wwbmmm:
The code LGTM, but need some explaination in the PR description

yanglimingcn:
> The code LGTM, but need some explaination in the PR description

done

PR #3061, opened 2025-08-11
`Fix: Preserve HTTP headers in ParallelChannel sub-calls`

XiaoYingGee:
## Problem
When using ParallelChannel, HTTP headers set on the parent controller (e.g., Authorization headers) are lost during sub-calls because they are not propagated to the sub-controllers.

## Solution
- Propagate all HTTP headers from parent controller to each sub-controller
- Preserve application-set headers (like Authorization) on each sub-call
- Maintain backward compatibility

### DIFF
--- a/src/brpc/parallel_channel.cpp
+++ b/src/brpc/parallel_channel.cpp
@@ -139,6 +139,14 @@ public:
             new (d->sub_done(i)) SubDone;
             d->sub_done(i)->cntl.ApplyClientSettings(settings);
             d->sub_done(i)->cntl.allow_done_to_run_in_place();
+
+            // Propagate all HTTP headers from parent controller to sub-controllers.
+            // This preserves application-set headers (e.g., Authorization) on each sub-call.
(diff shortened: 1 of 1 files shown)


src/brpc/parallel_channel.cpp
```diff
@@ -139,6 +139,18 @@ class ParallelChannelDone : public google::protobuf::Closure {
             new (d->sub_done(i)) SubDone;
             d->sub_done(i)->cntl.ApplyClientSettings(settings);
             d->sub_done(i)->cntl.allow_done_to_run_in_place();
+
+            // Propagate all HTTP headers from parent controller to sub-controllers.
+            // This preserves application-set headers (e.g., Authorization) on each sub-call.
```
wwbmmm, line 144: Does it need to check if sub-controller headers has exists key before AppendHeader or SetHeader? Avoid overwrite existing header.

chenBright:
Closed this as completed in #3213

PR #3075, opened 2025-08-27
Refactor: Use static_cast for safer downcasting in brpc::Span

zhangqiongyu:

Issue Number: no

Problem Summary:
Hello maintainers,

This PR makes two contributions:

##### 1. Code Quality Improvement:

It replaces the C-style cast in Span class with static_cast.

##### 2. Discussion on Robustness (Important):

During my usage of bRPC, I extended the functions of Collected and encountered a scenario where the container being sorted by SpanEarlier contained bvar::Collected* pointers that were not Span instances. This led to a crash due to the unsafe downcast.

The current implementation implicitly assumes that only Span objects will ever be passed to this comparator. While this is efficient, it's not robust against incorrect usage.

Changed:
Using static_cast instead of a C-style cast is a modern C++ best practice. It improves code quality by:

Type Safety: It provides compile-time checks for the validity of the cast within the type hierarchy.
Clarity: It makes the programmer's intent (a static downcast) explicit.
Maintainability: It's easier to search for and audit C++-style casts.
This change has no impact on runtime performance or behavior, as it preserves the original assumption that the bvar::Collected* pointers are guaranteed to be Span instances.

Side effects:
- Performance effects: no

---
### Check List:

### DIFF
--- a/src/brpc/span.cpp
+++ b/src/brpc/span.cpp
@@ -130,7 +130,7 @@ Span* Span::CreateClientSpan(const std::string& full_method_name,
     span->_tls_next = NULL;
     span->_full_method_name = full_method_name;
     span->_info.clear();
-    Span* parent = (Span*)bthread::tls_bls.rpcz_parent_span;
+    Span* parent = static_cast<Span*>(bthread::tls_bls.rpcz_parent_span);
     if (parent) {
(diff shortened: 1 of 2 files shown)


wwbmmm:
LGTM

yanglimingcn:
LGTM

PR #3076, opened 2025-08-27
rdma runtime extend block restriction

yanglimingcn:

Issue Number:

Problem Summary:

When both rdma_memory_pool_max_regions and rdma_memory_pool_buckets are greater than 1, dynamic memory expansion may cause concurrent modification issues in the memory linked list due to lock contention problems. To address this, we increase the region_num count for each block_type. Dynamic memory expansion is only permitted when both of the following conditions are met:

rdma_memory_pool_buckets equals 1
g_info->region_num[block_type] is less than 1

Changed:

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/brpc/input_messenger.cpp
+++ b/src/brpc/input_messenger.cpp
@@ -331,7 +331,12 @@ int InputMessenger::ProcessNewMessage(
                     "destroyed when authentication failed";
             }
         }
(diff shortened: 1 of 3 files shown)


wwbmmm:
LGTM

Issue #3088, opened 2025-09-11
Support to check all living bthread to help debug


ZhengweiZhu:
**Is your feature request related to a problem?**
brpc kindly provides a bthreads_service to check a **specified** thread by curl ip:port/bthreads/<bthread_id>. The problem is that we have no idea what the <bthread_id> is as it is generated by code, which makes this service useless.

**Describe the solution you'd like**
I suggest to provide a method to check **all** living bthreads (bthreads created and not finished yet) and we can check by curl ip:port/bthreads/all or ip:port/bthreads/all?st=1, which would be very helpful for bthread debugging like bthread deadlock.

I have implemented the method and would like to create a pr if it is ok. @chenBright 
**Describe alternatives you've considered**

**Additional context/screenshots**

chenBright:
Nice job!

By the way, in our practice, the bthread id is obtained through logs.

PR #3096, opened 2025-09-17
Add support for checking all living bthreads

ZhengweiZhu:
User can check all living bthreads by `curl ip:port/bthreads/all` or `curl ip:port/bthreads/all?st=1` to show bthread stack trace. This is an enhancement of the original /bthreads service which provides a method to check a specified bthread by designated bthread id, as user has no idea what the bthread id is.

Condisering the performance cost brought by recording the bthread id on bthread startup and finish, currently this function is only enabled when BRPC_BTHREAD_TRACER is defined.

brpc kindly provides a bthreads_service to check a specified thread by curl ip:port/bthreads/<bthread_id>. The problem is that we have no idea what the <bthread_id> is as it is generated by code, which makes this service useless.

Issue Number:
#3088 

The sample output with stack trace (note that the bthread name is shown due to another commit not included in this pr yet) :
```
# curl ip:port/bthreads/all?st=1
bthread=4294972672 :
stop=0
interrupted=0
about_to_quit=0
fn=0x55b2104e6d40
arg=0
attr={stack_type=3 flags=0 specified tag=-1 name=GlobalUpdate keytable_pool=0}
has_tls=0
uptime_ns=190567011013
cputime_ns=4864568
nswitch=191
status=6
traced=0
worker_tid=64
bthread call stack:
#0 0x55b210601235 bthread::TaskGroup::sched()
#1 0x55b21060139d bthread::TaskGroup::usleep()
#2 0x55b2105f7c68 bthread_usleep
#3 0x55b2104e72f5 brpc::GlobalUpdate()
#4 0x55b210601b52 bthread::TaskGroup::task_runner()
#5 0x55b21061baf1 bthread_make_fcontext

bthread=4294972673 :
stop=0
interrupted=0
about_to_quit=0
fn=0x55b21047fe00
arg=0x55b2d3d0ba00
attr={stack_type=3 flags=0 specified tag=-1 name=RunWatchConnections keytable_pool=0}
has_tls=0
uptime_ns=190569412002
cputime_ns=2741573
nswitch=191
status=6
traced=0
worker_tid=63
bthread call stack:
#0 0x55b210601235 bthread::TaskGroup::sched()
#1 0x55b21060139d bthread::TaskGroup::usleep()
#2 0x55b2105f7c68 bthread_usleep
#3 0x55b21047fade brpc::SocketMap::WatchConnections()
#4 0x55b21047fe09 brpc::SocketMap::RunWatchConnections()
#5 0x55b210601b52 bthread::TaskGroup::task_runner()
#6 0x55b21061baf1 bthread_make_fcontext

bthread=4294972674 :
stop=0
interrupted=0
about_to_quit=0
fn=0x55b2104e09f0
arg=0x55b213820248
attr={stack_type=3 flags=64 specified tag=0 name=EventDispatcher::RunThis keytable_pool=0}
has_tls=0
uptime_ns=190569994951
cputime_ns=188933862493
nswitch=92666
status=5
traced=0
worker_tid=62
bthread call stack:
#0 0x55b2104e0729 brpc::EventDispatcher::Run()
#1 0x55b2104e09f9 brpc::EventDispatcher::RunThis()
#2 0x55b210601b52 bthread::TaskGroup::task_runner()
#3 0x55b21061baf1 bthread_make_fcontext

bthread=4294972675 :
stop=0
interrupted=0
about_to_quit=0
fn=0x55b2104e09f0
arg=0x55b2138202a0
attr={stack_type=3 flags=64 specified tag=1 name=EventDispatcher::RunThis keytable_pool=0}
has_tls=0
uptime_ns=190571575082
cputime_ns=0
nswitch=0
status=5
traced=0
worker_tid=73
bthread call stack:
#0 0x55b2104e0729 brpc::EventDispatcher::Run()
#1 0x55b2104e09f9 brpc::EventDispatcher::RunThis()
#2 0x55b210601b52 bthread::TaskGroup::task_runner()
#3 0x55b21061baf1 bthread_make_fcontext

bthread=4294972676 :
stop=0
interrupted=0
about_to_quit=0
fn=0x55b2104a8bc0
arg=0x7ffda6297480
attr={stack_type=3 flags=0 specified tag=0 name=UpdateDerivedVars keytable_pool=0}
has_tls=0
uptime_ns=186566458068
cputime_ns=31630252
nswitch=187
status=6
traced=0
worker_tid=63
bthread call stack:
#0 0x55b210601235 bthread::TaskGroup::sched()
#1 0x55b21060139d bthread::TaskGroup::usleep()
#2 0x55b2105f7c68 bthread_usleep
#3 0x55b2104a9a15 brpc::Server::UpdateDerivedVars()
#4 0x55b210601b52 bthread::TaskGroup::task_runner()
#5 0x55b21061baf1 bthread_make_fcontext

bthread=4294972677 :
stop=0
interrupted=0
about_to_quit=0
fn=0x55b2104a8bc0
arg=0x7ffda629a340
attr={stack_type=3 flags=0 specified tag=0 name=UpdateDerivedVars keytable_pool=0}
has_tls=0
uptime_ns=186567623212
cputime_ns=33209308
nswitch=187
status=6
traced=0
worker_tid=64
bthread call stack:
#0 0x55b210601235 bthread::TaskGroup::sched()
#1 0x55b21060139d bthread::TaskGroup::usleep()
#2 0x55b2105f7c68 bthread_usleep
#3 0x55b2104a9a15 brpc::Server::UpdateDerivedVars()
#4 0x55b210601b52 bthread::TaskGroup::task_runner()
#5 0x55b21061baf1 bthread_make_fcontext

bthread=592705498944 :
stop=0
interrupted=0
about_to_quit=0
fn=0x55b210489c00
arg=0x55b3366ccf00
attr={stack_type=3 flags=0 specified tag=0 name=ProcessEvent keytable_pool=0x55b30773c780}
has_tls=0
uptime_ns=13930923
cputime_ns=0
nswitch=0
status=5
traced=0
worker_tid=67
bthread call stack:
No frame
Error message:

Can not trace self=592705498944
```

Problem Summary:

Changed:
when BRPC_BTHREAD_TRACER is defined: 1. add method to record bthread id on bthread startup and finish. 2. add /bthreads/all and /bthreads/all?st=1 service for user

Side effects:
- Performance effects:
The primary effects on performance is brought by recording the bthread id on bthread startup and finish, which needs lock to protect the data structure (like std::set).
To test the performance cost I use one rdma_performance_client and rdma_performance_server in each machine, with TCP mode enabled:   
image

This commit use multiple spin lock by sharding to minimize performance effects.

- Breaking backward compatibility: N/A

---
### Check List:

### DIFF
--- a/src/brpc/builtin/bthreads_service.cpp
+++ b/src/brpc/builtin/bthreads_service.cpp
@@ -23,7 +23,10 @@
 #include "brpc/builtin/bthreads_service.h"
 
 namespace bthread {
-extern void print_task(std::ostream& os, bthread_t tid);
+extern void print_task(std::ostream& os, bthread_t tid, bool enable_trace);
+#ifdef BRPC_BTHREAD_TRACER
+extern void print_living_task(std::ostream& os, bool enable_trace);
+#endif // BRPC_BTHREAD_TRACER
 }
 
 
@@ -38,30 +41,42 @@ void BthreadsService::default_method(::google::protobuf::RpcController* cntl_bas
     cntl->http_response().set_content_type("text/plain");
     butil::IOBufBuilder os;
     const std::string& constraint = cntl->http_request().unresolved_path();
-    
     if (constraint.empty()) {
 #ifdef BRPC_BTHREAD_TRACER
-        os << "Use /bthreads/<bthread_id> or /bthreads/<bthread_id>?st=1 for stack trace";
+        // when BRPC_BTHREAD_TRACER enabled, we can trace a specified bthread and
+        // check all living bthread
+        os << "Use /bthreads/<bthread_id> or /bthreads/<bthread_id>?st=1 for stack trace\n";
+        os << "To check all living bthread, use /bthreads/<all> or /bthreads/<all>?st=1 for stack trace\n";
 #else
-        os << "Use /bthreads/<bthread_id>";
+        os << "Use /bthreads/<bthread_id>\n";
 #endif // BRPC_BTHREAD_TRACER
     } else {
-        char* endptr = NULL;
-        bthread_t tid = strtoull(constraint.c_str(), &endptr, 10);
-        if (*endptr == '\0' || *endptr == '/' || *endptr == '?') {
-            ::bthread::print_task(os, tid);
-
+        bool enable_trace = false;
 #ifdef BRPC_BTHREAD_TRACER
             const std::string* st = cntl->http_request().uri().GetQuery("st");
             if (NULL != st && *st == "1") {
-                os << "\nbthread call stack:\n";
-                ::bthread::stack_trace(os, tid);
+                enable_trace = true;
             }
 #endif // BRPC_BTHREAD_TRACER
+        char* endptr = NULL;
+        bthread_t tid = strtoull(constraint.c_str(), &endptr, 10);
+        if (*endptr == '\0' || *endptr == '/' || *endptr == '?') {
+            ::bthread::print_task(os, tid, enable_trace);
+        }
+#ifdef BRPC_BTHREAD_TRACER
+        else if (constraint != "all" && constraint != "all?st=1") {
+            cntl->SetFailed(ENOMETHOD, "path=%s is not a bthread id or all, or all?st=1\n",
+                            constraint.c_str());
         } else {
-            cntl->SetFailed(ENOMETHOD, "path=%s is not a bthread id",
+            ::bthread::print_living_task(os, enable_trace);
+        }
+#else
+        else {
+            cntl->SetFailed(ENOMETHOD, "path=%s is not a bthread id\n",
                             constraint.c_str());
         }
+#endif // BRPC_BTHREAD_TRACER
+
     }
     os.move_to(cntl->response_attachment());
 }
--- a/src/bthread/bthread.cpp
+++ b/src/bthread/bthread.cpp
@@ -32,6 +32,7 @@
 #include "bthread/bthread.h"
 
 namespace bthread {
+extern void print_task(std::ostream& os, bthread_t tid, bool enable_trace);
 
 static bool validate_bthread_concurrency(const char*, int32_t val) {
     // bthread_setconcurrency sets the flag on success path which should
@@ -191,6 +192,23 @@ std::string stack_trace(bthread_t tid) {
         return "TaskControl has not been created";
     }
     return c->stack_trace(tid);
+}
+
+// Print all living (started and not finished) bthread
+void print_living_task(std::ostream& os, bool enable_trace) {
+    TaskControl* c = get_task_control();
+    if (NULL == c) {
+        os << "TaskControl has not been created";
+        return;
+    }
+    std::set<bthread_t> tids = c->get_living_bthreads();
+    if (tids.empty()) {
+        os << "No living bthreads\n";
+        return;
+    }
+    for (auto tid : tids) {
+        print_task(os, tid, enable_trace);
+    }
 }
 #endif // BRPC_BTHREAD_TRACER
 
--- a/src/bthread/task_control.h
+++ b/src/bthread/task_control.h
@@ -28,6 +28,7 @@
 #include <signal.h>
 #include <stddef.h>                             // size_t
 #include <vector>
+#include <set>
 #include <array>
 #include <memory>
 #include "butil/atomicops.h"                     // butil::atomic
@@ -39,6 +40,71 @@
 
 DECLARE_int32(task_group_ntags);
 namespace bthread {
+
+#ifdef BRPC_BTHREAD_TRACER
+class ShardedBthreadSet {
+public:
+    explicit ShardedBthreadSet(size_t shard_count = 16)
+        : _shard_count(shard_count), _shards(shard_count) {}
+
+    void add(bthread_t id) {
+        get_shard(id).add(id);
+    }
+
+    void remove(bthread_t id) {
+        get_shard(id).remove(id);
+    }
+
+    std::set<bthread_t> get_all() {
+        std::set<bthread_t> result;
+        for (auto& shard : _shards) {
+            shard.copy_to(result);
+        }
+        return result;
+    }
+
+private:
+    class Shard {
+    public:
+        Shard() {
+            pthread_spin_init(&_spinlock, PTHREAD_PROCESS_PRIVATE);
+        }
+
+        ~Shard() {
+            pthread_spin_destroy(&_spinlock);
+        }
+
+        Shard(const Shard&) = delete;
+        Shard& operator=(const Shard&) = delete;
+
+        void add(bthread_t id) {
+            BAIDU_SCOPED_LOCK(_spinlock);
+            _bthread_ids.insert(id);
+        }
+
+        void remove(bthread_t id) {
+            BAIDU_SCOPED_LOCK(_spinlock);
+            _bthread_ids.erase(id);
+        }
+
+        void copy_to(std::set<bthread_t>& target) {
+            BAIDU_SCOPED_LOCK(_spinlock);
+            target.insert(_bthread_ids.begin(), _bthread_ids.end());
+        }
+
+    private:
+        pthread_spinlock_t _spinlock;
+        std::set<bthread_t> _bthread_ids;
+    };
+
+    Shard& get_shard(bthread_t id) {
+        return _shards[id % _shard_count];
+    }
+
+    const size_t _shard_count;
+    std::vector<Shard> _shards;
+};
+#endif // BRPC_BTHREAD_TRACER
 
 class TaskGroup;
 
@@ -95,6 +161,9 @@ public:
     // A stacktrace of bthread can be helpful in debugging.
     void stack_trace(std::ostream& os, bthread_t tid);
     std::string stack_trace(bthread_t tid);
+    std::set<bthread_t> get_living_bthreads() {
+        return _living_bthreads.get_all();
+    }
 #endif // BRPC_BTHREAD_TRACER
 
     void push_priority_queue(bthread_tag_t tag, bthread_t tid) {
@@ -129,6 +198,16 @@ private:
     bvar::LatencyRecorder* create_exposed_pending_time();
     bvar::Adder<int64_t>& tag_nworkers(bthread_tag_t tag);
     bvar::Adder<int64_t>& tag_nbthreads(bthread_tag_t tag);
+
+#ifdef BRPC_BTHREAD_TRACER
+    void record_bthread_start(bthread_t tid) {
+        _living_bthreads.add(tid);
+    }
+
+    void record_bthread_finish(bthread_t tid) {
+        _living_bthreads.remove(tid);
+    }
+#endif // BRPC_BTHREAD_TRACER
 
     std::vector<butil::atomic<size_t>> _tagged_ngroup;
     std::vector<TaggedGroups> _tagged_groups;
@@ -165,6 +244,7 @@ private:
 
 #ifdef BRPC_BTHREAD_TRACER
     TaskTracer _task_tracer;
+    ShardedBthreadSet _living_bthreads;
 #endif // BRPC_BTHREAD_TRACER
 
 };
--- a/src/bthread/task_group.cpp
+++ b/src/bthread/task_group.cpp
@@ -36,6 +36,7 @@
 #include "bthread/task_control.h"
 #include "bthread/task_group.h"
 #include "bthread/timer_thread.h"
+#include "bthread/bthread.h"
 
 #ifdef __x86_64__
 #include <x86intrin.h>
@@ -442,6 +443,7 @@ void TaskGroup::task_runner(intptr_t skip_remained) {
             g->_control->_task_tracer.WaitForTracing(m);
         }
         g->_control->_task_tracer.set_status(TASK_STATUS_UNKNOWN, m);
+        g->_control->record_bthread_finish(m->tid);
 #endif // BRPC_BTHREAD_TRACER
 
         g->_control->_nbthreads << -1;
@@ -507,6 +509,7 @@ int TaskGroup::start_foreground(TaskGroup** pg,
     g->_control->tag_nbthreads(g->tag()) << 1;
 #ifdef BRPC_BTHREAD_TRACER
     g->_control->_task_tracer.set_status(TASK_STATUS_CREATED, m);
+    g->_control->record_bthread_start(*th);
 #endif // BRPC_BTHREAD_TRACER
     if (g->is_current_pthread_task()) {
         // never create foreground task in pthread.
@@ -570,6 +573,7 @@ int TaskGroup::start_background(bthread_t* __restrict th,
     _control->tag_nbthreads(tag()) << 1;
 #ifdef BRPC_BTHREAD_TRACER
     _control->_task_tracer.set_status(TASK_STATUS_CREATED, m);
+    _control->record_bthread_start(*th);
 #endif // BRPC_BTHREAD_TRACER
     if (REMOTE) {
         ready_to_run_remote(m, (using_attr.flags & BTHREAD_NOSIGNAL));
@@ -1088,10 +1092,10 @@ void TaskGroup::yield(TaskGroup** pg) {
     sched(pg);
 }
 
-void print_task(std::ostream& os, bthread_t tid) {
+void print_task(std::ostream& os, bthread_t tid, bool enable_trace) {
     TaskMeta* const m = TaskGroup::address_meta(tid);
     if (m == NULL) {
-        os << "bthread=" << tid << " : never existed";
+        os << "bthread=" << tid << " : never existed\n";
         return;
     }
     const uint32_t given_ver = get_version(tid);
@@ -1127,7 +1131,7 @@ void print_task(std::ostream& os, bthread_t tid) {
         }
     }
     if (!matched) {
-        os << "bthread=" << tid << " : not exist now";
+        os << "bthread=" << tid << " : not exist now\n";
     } else {
         os << "bthread=" << tid << " :\nstop=" << stop
            << "\ninterrupted=" << interrupted
@@ -1136,6 +1140,7 @@ void print_task(std::ostream& os, bthread_t tid) {
            << "\narg=" << (void*)arg
            << "\nattr={stack_type=" << attr.stack_type
            << " flags=" << attr.flags
+           << " specified tag=" << attr.tag
            << " keytable_pool=" << attr.keytable_pool
            << "}\nhas_tls=" << has_tls
            << "\nuptime_ns=" << butil::cpuwide_time_ns() - cpuwide_start_ns
@@ -1145,8 +1150,15 @@ void print_task(std::ostream& os, bthread_t tid) {
            << "\nstatus=" << status
            << "\ntraced=" << traced
            << "\nworker_tid=" << worker_tid;
-#else
-           ;
+        if (enable_trace) {
+            os << "\nbthread call stack:\n";
+            stack_trace(os, tid);
+            os << "\n";
+        } else {
+            os << "\n\n";
+        }
+ #else
+           << "\n\n";
            (void)status;(void)traced;(void)worker_tid;
 #endif // BRPC_BTHREAD_TRACER
     }
--- a/src/brpc/builtin/index_service.cpp
+++ b/src/brpc/builtin/index_service.cpp
@@ -158,7 +158,11 @@ void IndexService::default_method(::google::protobuf::RpcController* controller,
        << Path("/health", html_addr) << " : Test healthy" << NL
        << Path("/vlog", html_addr) << " : List all VLOG callsites" << NL
        << Path("/sockets", html_addr) << " : Check status of a Socket" << NL
+#ifdef BRPC_BTHREAD_TRACER
+       << Path("/bthreads", html_addr) << " : Check status of a bthread or all living bthreads" << NL
+#else
        << Path("/bthreads", html_addr) << " : Check status of a bthread" << NL
+#endif // BRPC_BTHREAD_TRACER
        << Path("/ids", html_addr) << " : Check status of a bthread_id" << NL
        << Path("/protobufs", html_addr) << " : List all protobuf services and messages" << NL
        << Path("/list", html_addr) << " : json signature of methods" << NL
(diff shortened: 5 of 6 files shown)


ZhengweiZhu:
This pr will be useful to help diagnose bthread problem like bthread deadlock, which is hard or unrealistic by using gdb only. 
However condisering the performance cost brought by lock,  temporarily I just enable this feature when BRPC_BTHREAD_TRACER is defined. **Do u guys have any suggestion to minimize the performance cost so let it be applicable for all situations.**  If so, I plan to add method to set bthread name and user can check all living bthreads with more detail, reducing the need to enable BRPC_BTHREAD_TRACER for stack trace.

wwbmmm:
Maybe you can try to get all bthread_id from the `butil::ResourcePool<bthread::TaskMeta>`? (needs to add some interface to butil::ResourcePool)

chenBright:
> Maybe you can try to get all bthread_id from the `butil::ResourcePool<bthread::TaskMeta>`? (needs to add some interface to butil::ResourcePool)

I agree with the `butil::ResourcePool<bthread::TaskMeta>` approach.

yanglimingcn:
yes, I think list all bthread_id is better。

ZhengweiZhu:
> Maybe you can try to get all bthread_id from the `butil::ResourcePool<bthread::TaskMeta>`? (needs to add some interface to butil::ResourcePool)

@wwbmmm @chenBright   Yes I have thought of this approach but with no result.  


Firstly, when we create a bthread, a ResourceId and TaskMeta instance is acquired from butil::ResourcePool by this way:
`  
butil::ResourceId<TaskMeta> slot;
TaskMeta* m = butil::get_resource(&slot);
`
If we know all the ResourceId<TaskMeta> slot then we can get all living bthread id by address_resource(slot)->tid. Inside the ResourcePool implementation, the slot is acquired by the following order: 1. thread local FreeChunk 2. global FreeChunk 3. thread local Block 4. global Block (from BlockGroup). The problem is the ResourcePool does not record the slot **in use** and it only knows the slot after it is not in use and returned to ResourcePool, which does not satisfy our requirement here.

Secondly, if we add some interface in ResourcePool to record the slot when get_resource and consider removing it when return_resource, certainly we might use thread local LocalPool to record for the sake of performance effect. The problem is the get_resource and return_resource may run in different pthread (different tls) as bthread switches to another worker. What's more, if we plan to get all living bthread id by summarizing **all**  thread local records, the tricky part is the ResourcePool interface can be called from worker and non worker, making it even impossible to summarize.

Any ideas?

chenBright (commented):
I think you can traverse the global `butil::ResourceId<TaskMeta>::_block_groups` and get all tids.

ZhengweiZhu:
> I think you can traverse the global `butil::ResourceId<TaskMeta>::_block_groups` and get all tids.

That's huge. And they represent all bthread id which ever existed，not living one？

chenBright:
> > I think you can traverse the global `butil::ResourceId<TaskMeta>::_block_groups` and get all tids.
> 
> That's huge. And they represent all bthread id which ever existed，not living one？

Maybe you can use TaskStatus to determine whether a bthread is alive.

chenBright:
I think we can add a function to display all the live bthread ids and names, and click on the link to display the corresponding bthread details.

wwbmmm:
> > I think you can traverse the global `butil::ResourceId<TaskMeta>::_block_groups` and get all tids.
> 
> That's huge. And they represent all bthread id which ever existed，not living one？

No. After a bthread exit, its TaskMeta will be return to the ResourcePool, and be reused by new bthread. When you traverse the ResourcePool, you only need to traverse those slots in use, you don't need to traverse those in the free list.

ZhengweiZhu:
> > > I think you can traverse the global `butil::ResourceId<TaskMeta>::_block_groups` and get all tids.
> > 
> > 
> > That's huge. And they represent all bthread id which ever existed，not living one？
> 
> No. After a bthread exit, its TaskMeta will be return to the ResourcePool, and be reused by new bthread. When you traverse the ResourcePool, you only need to traverse those slots in use, you don't need to traverse those in the free list.



How to judge if the slot is in use? @chenBright  The TaskStatus is only used when TaskTracer is enabled, but I want this pr to be applicable even when TaskTracer is not enabled. 

If I can judge if the slot is in use, then I need to traverse in the same way as `describe_resources` does?

ZhengweiZhu:
> I think we can add a function to display all the live bthread ids and names, and click on the link to display the corresponding bthread details.

If the TaskTracer is enabled, we can easily see call trace of all the living bthread and debug the deadlock problem, just like the same way as gdb/gstack does. And in normal situation, the living bthreads won't be many. Is it necessary then?

chenBright:
> > I think we can add a function to display all the live bthread ids and names, and click on the link to display the corresponding bthread details.
> 
> If the TaskTracer is enabled, we can easily see call trace of all the living bthread and debug the deadlock problem, just like the same way as gdb/gstack does. And in normal situation, the living bthreads won't be many. Is it necessary then?

Yes. Because it allows users to view the call stack of a specified bthread, not all of them.

chenBright:
> The TaskStatus is only used when TaskTracer is enabled, but I want this pr to be applicable even when TaskTracer is not enabled.

Perhaps the default support for TaskStatus can meet your needs.

ZhengweiZhu:
> TaskStatus

ZhengweiZhu:
> > The TaskStatus is only used when TaskTracer is enabled, but I want this pr to be applicable even when TaskTracer is not enabled.
> 
> Perhaps the default support for TaskStatus can meet your needs.

Currently the TaskStatus is only set when TaskTracer is enabled. Do u mean I need to change the code to set some status   when TaskTracer is not enabled? Like set status to TASK_STATUS_CREATED when create a bthread and set it status to TASK_STATUS_UNKNOWN when that bthread is destroyed, and a bthread can be judged  alive when its status is not TASK_STATUS_UNKNOWN?

ZhengweiZhu:
> > > I think we can add a function to display all the live bthread ids and names, and click on the link to display the corresponding bthread details.
> > 
> > 
> > If the TaskTracer is enabled, we can easily see call trace of all the living bthread and debug the deadlock problem, just like the same way as gdb/gstack does. And in normal situation, the living bthreads won't be many. Is it necessary then?
> 
> Yes. Because it allows users to view the call stack of a specified bthread, not all of them.

I  can add link to display the corresponding bthread details. And still display all living bthreads by default. The bthread name will not be shown and I can make another pr to add support naming bthread (and even execution_queue). Is that ok?

chenBright:
> > > The TaskStatus is only used when TaskTracer is enabled, but I want this pr to be applicable even when TaskTracer is not enabled.
> > 
> > 
> > Perhaps the default support for TaskStatus can meet your needs.
> 
> Currently the TaskStatus is only set when TaskTracer is enabled. Do u mean I need to change the code to set some status when TaskTracer is not enabled? Like set status to TASK_STATUS_CREATED when create a bthread and set it status to TASK_STATUS_UNKNOWN when that bthread is destroyed, and a bthread can be judged alive when its status is not TASK_STATUS_UNKNOWN?

Yes.

chenBright:
> > > > I think we can add a function to display all the live bthread ids and names, and click on the link to display the corresponding bthread details.
> > > 
> > > 
> > > If the TaskTracer is enabled, we can easily see call trace of all the living bthread and debug the deadlock problem, just like the same way as gdb/gstack does. And in normal situation, the living bthreads won't be many. Is it necessary then?
> > 
> > 
> > Yes. Because it allows users to view the call stack of a specified bthread, not all of them.
> 
> I can add link to display the corresponding bthread details. And still display all living bthreads by default. The bthread name will not be shown and I can make another pr to add support naming bthread (and even execution_queue). Is that ok?

No problem.

ZhengweiZhu:
@wwbmmm @chenBright  PTAL. The pr is updated and applicable no matter if BRPC_BTHREAD_TRACER is defined, with no performance side effect. Thanks for the suggestion.  As for the idea to ` add link to display the corresponding bthread details`, it's nice to have and probably be implemented later.

src/brpc/builtin/bthreads_service.cpp
```diff
@@ -23,7 +23,9 @@
 #include "brpc/builtin/bthreads_service.h"
 
 namespace bthread {
-extern void print_task(std::ostream& os, bthread_t tid);
+extern void print_task(std::ostream& os, bthread_t tid, bool enable_trace,
+    bool ignore_not_matched = false);
+extern void print_living_task(std::ostream& os, bool enable_trace);
```
wwbmmm: task -> tasks

src/brpc/builtin/bthreads_service.cpp
```diff
@@ -38,30 +40,34 @@ void BthreadsService::default_method(::google::protobuf::RpcController* cntl_bas
     cntl->http_response().set_content_type("text/plain");
     butil::IOBufBuilder os;
     const std::string& constraint = cntl->http_request().unresolved_path();
-    
     if (constraint.empty()) {
 #ifdef BRPC_BTHREAD_TRACER
-        os << "Use /bthreads/<bthread_id> or /bthreads/<bthread_id>?st=1 for stack trace";
+        os << "Use /bthreads/<bthread_id> or /bthreads/<bthread_id>?st=1 for stack trace\n";
+        os << "To check all living bthread, use /bthreads/<all> or /bthreads/<all>?st=1 for stack trace\n";
```
wwbmmm: `<all>` -> all

src/brpc/builtin/bthreads_service.cpp
```diff
@@ -38,30 +40,34 @@ void BthreadsService::default_method(::google::protobuf::RpcController* cntl_bas
...
 #else
-        os << "Use /bthreads/<bthread_id>";
+        os << "Use /bthreads/<bthread_id>\n";
+        os << "To check all living bthread, use /bthreads/<all>\n";
```
wwbmmm: `<all>` -> all

src/butil/resource_pool_inl.h
```diff
@@ -307,6 +308,28 @@ class BAIDU_CACHELINE_ALIGNMENT ResourcePool {
         const size_t n = ResourcePoolFreeChunkMaxItem<T>::value();
         return n < FREE_CHUNK_NITEM ? n : FREE_CHUNK_NITEM;
     }
+
+    std::list<T*> list_resources() const {
```
wwbmmm: list -> vector

src/bthread/task_group.cpp
```diff
@@ -1127,7 +1135,9 @@ void print_task(std::ostream& os, bthread_t tid) {
         }
     }
     if (!matched) {
-        os << "bthread=" << tid << " : not exist now";
+        if (!ignore_not_matched) {
```
wwbmmm, line 1132: I don't understand the function of ignore_not_matched here. I think when the tid and version_butex not match, it indicates that the bthread is not living.

src/bthread/task_meta.h
```diff
@@ -107,7 +107,8 @@ struct TaskMeta {
     // DO NOT use this field directly, use tls_bls instead.
     LocalStorage local_storage{};
 
-    // Only used when TaskTracer is enabled.
+    // Mainly used when TaskTracer is enabled. Otherwise used to just indicate
+    // if a bthread is alive.
```
wwbmmm: I think the tid and version_butex can be used to judge whether the bthread is alive, so you don't need to use this status.

src/bthread/task_control.h
```diff
@@ -28,6 +28,7 @@
 #include <signal.h>
 #include <stddef.h>                             // size_t
 #include <vector>
+#include <set>
```
chenBright: unused?

src/bthread/task_group.cpp
```diff
@@ -1145,8 +1156,15 @@ void print_task(std::ostream& os, bthread_t tid) {
...
            << "\nworker_tid=" << worker_tid;
-#else
-           ;
+        if (enable_trace) {
+            os << "\nbthread call stack:\n";
+            stack_trace(os, tid);
+            os << "\n\n";
+        } else {
+            os << "\n\n";
+        }
```
chenBright, line 1156:
```c++
if (enable_trace) {
    os << "\nbthread call stack:\n";
    stack_trace(os, tid);
}
os << "\n\n";
```

src/bthread/task_group.cpp
ZhengweiZhu: > I don't understand the function of ignore_not_matched here. I think when the tid and version_butex not match, it indicates that the bthread is not living.

Yes, when the tid and version_butex not matches, it indicates the bthread is not living and should not be printed. Without this ignore_not_matched judgement, it will print `bthread=xxx : not exist now`, which is not what I want.

src/butil/resource_pool_inl.h
ZhengweiZhu: > list -> vector

As the vector will grow inefficiently so I prefer list. is it ok?

src/bthread/task_control.h
ZhengweiZhu: yes. I will fix it.

src/bthread/task_meta.h
ZhengweiZhu: If so, the code can be simplified. @chenBright  Do u agree?

src/butil/resource_pool_inl.h
wwbmmm: std::list use linked list and will do a lot of malloc/free calls. I am not sure whether std::list is faster than std::vector here. Maybe you can do some performance testing.
This is a small issue. I think both std::list and std::vector are acceptable.

src/bthread/task_meta.h
chenBright: I think this is better. If the bthread is not alive, tid is old and version_butexis new.

src/butil/resource_pool_inl.h
chenBright: The number of bthreads is limited by vm.max_map_count, so it should be finite. Even if the number of active bthreads is too large, the bottleneck will not be std::vector, but somewhere else. Therefore, I think using std::vector is better.

ZhengweiZhu:
> Maybe you can try to get all bthread_id from the `butil::ResourcePool<bthread::TaskMeta>`? (needs to add some interface to butil::ResourcePool)

There's a trick here. As each task group creates a TaskMeta object internally to run main task, these TaskMeta will also be traversed. I will do a filter to not show those TaskMeta as they are opaque to user.

src/bthread/task_control.cpp
```diff
@@ -587,4 +587,21 @@ bvar::LatencyRecorder* TaskControl::create_exposed_pending_time() {
...
 
+std::vector<bthread_t> TaskControl::get_all_bthreads() {
+    std::vector<bthread_t> all_bthread_ids;
+    std::vector<TaskMeta*> all_tasks = butil::list_resources<TaskMeta>();
+    all_bthread_ids.reserve(all_tasks.size());
+    for (TaskMeta* m : all_tasks) {
+        // filter out those bthreads created by bthread_start* functions,
+        // i.e. not those created internally to run main task as they are
+        // opaque to user
+        if (m && !m->main_task) {
```
chenBright: https://github.com/apache/brpc/blob/07083331e1496746f4b1cfa54d0b6364f321be34/src/bthread/task_group.cpp#L310
https://github.com/apache/brpc/blob/07083331e1496746f4b1cfa54d0b6364f321be34/src/bthread/task_group.cpp#L50-L51

Can we use `m->attr.stack_type == BTHREAD_STACKTYPE_PTHREAD` to judge?

src/bthread/task_group.cpp
```diff
@@ -1088,10 +1098,11 @@ void TaskGroup::yield(TaskGroup** pg) {
     sched(pg);
 }
 
-void print_task(std::ostream& os, bthread_t tid) {
+void print_task(std::ostream& os, bthread_t tid, bool enable_trace,
+    bool ignore_not_matched = false) {
```
chenBright:
```c++
void print_task(std::ostream& os, bthread_t tid, bool enable_trace,
                bool ignore_not_matched = false) {
```
Alignment?

src/bthread/bthread.cpp
```diff
@@ -32,6 +32,8 @@
 #include "bthread/bthread.h"
 
 namespace bthread {
+extern void print_task(std::ostream& os, bthread_t tid, bool enable_trace,
+    bool ignore_not_matched = false);
```
chenBright: Alignment?

src/bthread/task_control.cpp
```diff
@@ -587,4 +587,21 @@ bvar::LatencyRecorder* TaskControl::create_exposed_pending_time() {
...
+            // the judgement of living bthread is deferred to `print_task`
+            // to avoid extra judgement
+            all_bthread_ids.push_back(m->tid);
```
chenBright: Support a `ResourcePool::for_each<>(std::function callback)` function to iterate over all TaskMeta and filter out the living bthreads. This way the vector will not be too large and there will be no need to copy between vectors. Isn't this better?

src/bthread/task_control.cpp
ZhengweiZhu: It works to use m->fn != NULL to judge

ZhengweiZhu:
The comments have all been resolved. But there's one problem left, that's when the bthread status is set to TASK_STATUS_READY, the actual status is set to TASK_STATUS_FIRST_READY.
https://github.com/apache/brpc/blob/07083331e1496746f4b1cfa54d0b6364f321be34/src/bthread/task_tracer.cpp#L150-L154

so the following judgement will never meet? 
https://github.com/apache/brpc/blob/07083331e1496746f4b1cfa54d0b6364f321be34/src/bthread/task_tracer.cpp#L257-L259

According to my test, there seems to always exist a bthread in TASK_STATUS_FIRST_READY status and not traceable.  This bthread has flag 320 which means "BTHREAD_NEVER_QUIT | BTHREAD_GLOBAL_PRIORITY" , which seems to be EventDispatcher? Is this expected?
image

@chenBright

src/bthread/task_control.cpp
```diff
@@ -587,4 +587,29 @@ bvar::LatencyRecorder* TaskControl::create_exposed_pending_time() {
     return pt;
 }
 
+std::vector<bthread_t> TaskControl::get_living_bthreads() {
+    std::vector<bthread_t> living_bthread_ids;
+    std::vector<TaskMeta*> living_tasks = butil::list_resources<TaskMeta>(
```
wwbmmm: Since `butil::list_resources` accepts a lambda function, I think it don't need to return `vector<TaskMeta*>` any more. you can capture the local variable `living_bthread_ids`  by reference and directly use `living_bthread_ids.push_back(m->tid)` in the lambda function.

chenBright:
> The comments have all been resolved. But there's one problem left, that's when the bthread status is set to TASK_STATUS_READY, the actual status is set to TASK_STATUS_FIRST_READY.
> 
> https://github.com/apache/brpc/blob/07083331e1496746f4b1cfa54d0b6364f321be34/src/bthread/task_tracer.cpp#L150-L154
> 
> so the following judgement will never meet?
> 
> https://github.com/apache/brpc/blob/07083331e1496746f4b1cfa54d0b6364f321be34/src/bthread/task_tracer.cpp#L257-L259
> 
> According to my test, there seems to always exist a bthread in TASK_STATUS_FIRST_READY status and not traceable. This bthread has flag 320 which means "BTHREAD_NEVER_QUIT | BTHREAD_GLOBAL_PRIORITY" , which seems to be EventDispatcher? Is this expected? image
> 
> @chenBright

It should be `if (TASK_STATUS_READY == s && NULL == m->stack)`, please fix it.

src/bthread/task_control.cpp
ZhengweiZhu: > Since `butil::list_resources` accepts a lambda function, I think it don't need to return `vector<TaskMeta*>` any more. you can capture the local variable `living_bthread_ids` by reference and directly use `living_bthread_ids.push_back(m->tid)` in the lambda function.

Yes. It's two different interface. list_resources indicates get all resources in resource pool meeting the filter condition, however your proposed interface is actually for_each_resource(lamda) to traverse each resource and apply some operation. Both way works here if the aim is to get living bthreads. I use list_resources interface just like the other interfaces like get_resource, address_resource etc.  Which one shall I expose? Or both ?

ZhengweiZhu:
> > The comments have all been resolved. But there's one problem left, that's when the bthread status is set to TASK_STATUS_READY, the actual status is set to TASK_STATUS_FIRST_READY.
> > https://github.com/apache/brpc/blob/07083331e1496746f4b1cfa54d0b6364f321be34/src/bthread/task_tracer.cpp#L150-L154
> > 
> > so the following judgement will never meet?
> > https://github.com/apache/brpc/blob/07083331e1496746f4b1cfa54d0b6364f321be34/src/bthread/task_tracer.cpp#L257-L259
> > 
> > According to my test, there seems to always exist a bthread in TASK_STATUS_FIRST_READY status and not traceable. This bthread has flag 320 which means "BTHREAD_NEVER_QUIT | BTHREAD_GLOBAL_PRIORITY" , which seems to be EventDispatcher? Is this expected? image
> > @chenBright
> 
> It should be `if (TASK_STATUS_READY == s && NULL == m->stack)`, please fix it.

Fixed!  BTW fix another _enable_priority_queue not initialized bug. 😂

src/bthread/task_control.cpp
wwbmmm: If you have `for_each_resource`, you don't need to expose the `list_resources`.
I think `for_each_resource` is more efficient, since it don't need to maintain a temp vector.

src/bthread/task_control.cpp
```diff
@@ -192,6 +192,7 @@ TaskControl::TaskControl()
     , _signal_per_second(&_cumulated_signal_count)
     , _status(print_rq_sizes_in_the_tc, this)
     , _nbthreads("bthread_count")
+    , _enable_priority_queue(false)
```
chenBright: `, _enable_priority_queue(FLAGS_enable_bthread_priority_queue)`

src/bthread/task_control.cpp
ZhengweiZhu: done

ZhengweiZhu:
@wwbmmm @chenBright  PTAL

wwbmmm:
LGTM

src/bthread/task_control.cpp
ZhengweiZhu: done

ZhengweiZhu:
Anything to do for the merge?

PR #3105, opened 2025-09-23
Fix hanging during SSL_write for large packets

old-bear:

Issue Number:

Problem Summary:
When sending large packet under SSL, brpc could stuck as data still resides in BIO buffer.

Changed: Disable BIO for SSL for 2 reasons:
1. The current implementation makes it very hard to handle EAGAIN from both SSL_write and BIO_flush.
2. BIO is optional and kind of outdate for modern TCP, which already has buffer mechanism.

Side effects:
- Performance effects: Should be no impact

- Breaking backward compatibility: No

---
### Check List:

### DIFF
--- a/src/butil/iobuf.cpp
+++ b/src/butil/iobuf.cpp
@@ -929,16 +929,17 @@ ssize_t IOBuf::cut_multiple_into_SSL_channel(SSL* ssl, IOBuf* const* pieces,
     }
 
 #ifndef USE_MESALINK
-    // Flush remaining data inside the BIO buffer layer
-    BIO* wbio = SSL_get_wbio(ssl);
-    if (BIO_wpending(wbio) > 0) {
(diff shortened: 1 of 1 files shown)


wwbmmm:
LGTM

chenBright:
Is BIO_flush still needed?

https://github.com/apache/brpc/blob/7229c3608f8cb98b24a0a2e7f99bc01d357d9312/src/butil/iobuf.cpp#L934-L939

old-bear:
> Is BIO_flush still needed?
> 
> https://github.com/apache/brpc/blob/7229c3608f8cb98b24a0a2e7f99bc01d357d9312/src/butil/iobuf.cpp#L934-L939

No. BIO_wpending will return 0, so this code will not be reached. For clarity, I have commented out the code and left a note.

PR #3115, opened 2025-10-12
Bugfix: TaskTracer deadlocks due to ABA problem

chenBright:

Issue Number:

Problem Summary:
https://github.com/apache/brpc/issues/2983#issuecomment-3388228204
https://github.com/apache/brpc/issues/2983#issuecomment-3394061697

Changed:

1. Not to support the tracing of bthread in TASK_STATUS_JUMPING status and just return an error msg like below, instead of waiting for jumping.
2. Taskmeta has its lock for waiting for tracing, to avoid starvation of bthread jumping caused by new tracing.
3. WaitForTracing acquire the lock, indicating that tracing is complete, so no condition variable is needed.

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/bthread/task_meta.h
+++ b/src/bthread/task_meta.h
@@ -112,6 +112,8 @@ struct TaskMeta {
     TaskStatus status{TASK_STATUS_UNKNOWN};
     // Whether bthread is traced？
     bool traced{false};
+    // [Not Reset] guarantee tracing completion before jumping.
+    pthread_mutex_t trace_lock{};
     // Worker thread id.
     pthread_t worker_tid{};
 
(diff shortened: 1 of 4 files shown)


chenBright:
CC @ZhengweiZhu @wwbmmm

wwbmmm:
LGTM

ZhengweiZhu:
LGTM

ZhengweiZhu:
I have an irrelevant doubt. The cur_ending parameter seems not used?
`static void sched_to(TaskGroup** pg, TaskMeta* next_meta, bool cur_ending);`

chenBright:
> I have an irrelevant doubt. The cur_ending parameter seems not used? `static void sched_to(TaskGroup** pg, TaskMeta* next_meta, bool cur_ending);`

cur_ending was used before, but is no longer used. You can delete it.

PR #3117, opened 2025-10-13
rdma polling mode all messages run in new bthreads

yanglimingcn:

Issue Number:

Problem Summary:

 In RDMA polling mode, all messages must be executed in a new bthread and
 not in the bthread where the polling bthread is located, because the
 method for processing messages may call synchronization primitives,
 causing the polling bthread to be scheduled out.

Changed:

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/brpc/input_messenger.cpp
+++ b/src/brpc/input_messenger.cpp
@@ -331,12 +331,7 @@ int InputMessenger::ProcessNewMessage(
                     "destroyed when authentication failed";
             }
         }
-#if BRPC_WITH_RDMA
-        if (!m->is_read_progressive() && !rdma::FLAGS_rdma_use_polling)
-#else
-        if (!m->is_read_progressive())
-#endif
-        {
+        if (!m->is_read_progressive()) {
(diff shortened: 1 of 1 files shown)


wwbmmm:
If the method for processing messages don't have synchronization primitives, running in the polling bthread is ok?
Will this PR affect performance?

yanglimingcn:
> If the method for processing messages don't have synchronization primitives, running in the polling bthread is ok? Will this PR affect performance?

yes，for example the method just change a variable. this just use in rdma polling mode and it is unavoidable. we don't know user's logic.
If the user confirms that synchronization primitives will not be called, set rdma_disable_bthread flag to true, all message process will be run in the polling thread.
summary:
1、in non-polling mode, socket will process in a new bthread(bthread_start_urgent) and last message will process in this bthread, other message will start new bthreads.
2、in polling-mode, socket will process in polling bthread, and all message will start bthreads.

chenBright (approved):
LGTM

PR #3124, opened 2025-10-23
feat: make bthread stack trace signal number configurable

### LINKED ISSUE #3118: Initialize TaskTracer failed due to SIGURG signal handler already registered
**Describe the bug**
When BRPC_BTHREAD_TRACER is enabled and the app is linked with cgo library (such as linked with etcd client library which is written in golang), the app will fail to start up as TaskTracer finds SIGURG signal handler is already registered (LINE 346).
https://github.com/apache/brpc/blob/4eeb5e260af165291f95f30f0fa64f9ff67ba192/src/bthread/task_tracer.cpp#L334-L350

The root cause is the SIGURG signal is unfortunately used internally as sigPreempt for non-cooperative preemption by golang. 

Image

```
file: https://go.dev/src/runtime/signal_unix.go

    43  // sigPreempt is the signal used for non-cooperative preemption.
    44  //
    45  // There's no good way to choose this signal, but there are some
    46  // heuristics:
    47  //
    48  // 1. It should be a signal that's passed-through by debuggers by
    49  // default. On Linux, this is SIGALRM, SIGURG, SIGCHLD, SIGIO,
    50  // SIGVTALRM, SIGPROF, and SIGWINCH, plus some glibc-internal signals.
    51  //
    52  // 2. It shouldn't be used internally by libc in mixed Go/C binaries
    53  // because libc may assume it's the only thing that can handle these
    54  // signals. For example SIGCANCEL or SIGSETXID.
    55  //
    56  // 3. It should be a signal that can happen spuriously without
    57  // consequences. For example, SIGALRM is a bad choice because the
    58  // signal handler can't tell if it was caused by the real process
    59  // alarm or not (arguably this means the signal is broken, but I
    60  // digress). SIGUSR1 and SIGUSR2 are also bad because those are often
    61  // used in meaningful ways by applications.
    62  //
    63  // 4. We need to deal with platforms without real-time signals (like
    64  // macOS), so those are out.
    65  //
    66  // We use SIGURG because it meets all of these criteria, is extremely
    67  // unlikely to be used by an application for its "real" meaning (both
    68  // because out-of-band data is basically unused and because SIGURG
    69  // doesn't report which socket has the condition, making it pretty
    70  // useless), and even if it is, the application has to be ready for
    71  // spurious SIGURG. SIGIO wouldn't be a bad choice either, but is more
    72  // likely to be used for real.
```

Any idea to fix? Maybe choose another signal number for stack trace as we can't change golang's implementation. @chenBright 

**To Reproduce**

**Expected behavior**

**Versions**
OS:
Compiler:
brpc:
protobuf:

ZhengweiZhu:

Issue Number: resolve #3118

Problem Summary:

Changed:

Side effects:
- Performance effects:
NO
NO
---
### Check List:

### DIFF
--- a/src/bthread/task_tracer.cpp
+++ b/src/bthread/task_tracer.cpp
@@ -35,6 +35,10 @@ namespace bthread {
 
 DEFINE_uint32(signal_trace_timeout_ms, 50, "Timeout for signal trace in ms");
 BUTIL_VALIDATE_GFLAG(signal_trace_timeout_ms, butil::PositiveInteger<uint32_t>);
+// Note that SIGURG handler may be registered by some library such as cgo
+// so let the signal number be configurable
+DEFINE_int32(signal_number_for_trace, SIGURG,
+             "signal number used for stack trace, default to SIGURG");
 
 extern BAIDU_THREAD_LOCAL TaskMeta* pthread_fake_meta;
 
@@ -315,17 +319,19 @@ TaskTracer::Result TaskTracer::TraceByLibunwind(unw_cursor_t& cursor) {
 
 bool TaskTracer::RegisterSignalHandler() {
     // Set up the signal handler.
+    _signal_num = FLAGS_signal_number_for_trace;
     struct sigaction old_sa{};
     struct sigaction sa{};
     sa.sa_sigaction = SignalHandler;
     sa.sa_flags = SA_SIGINFO;
     sigfillset(&sa.sa_mask);
-    if (sigaction(SIGURG, &sa, &old_sa) != 0) {
+    if (sigaction(_signal_num, &sa, &old_sa) != 0) {
         PLOG(ERROR) << "Failed to sigaction";
         return false;
     }
     if (NULL != old_sa.sa_handler || NULL != old_sa.sa_sigaction) {
-        LOG(ERROR) << "Signal handler of SIGURG is already registered";
+        LOG(ERROR) << "Signal handler of signal number "
+                   << _signal_num << " is already registered";
         return false;
     }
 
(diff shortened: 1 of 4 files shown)


ZhengweiZhu:
@chenBright Take a look~

src/bthread/task_tracer.h
```diff
@@ -115,6 +118,7 @@ class TaskTracer {
     std::vector<butil::intrusive_ptr<SignalSync>> _inuse_signal_syncs;
 
     bvar::LatencyRecorder _trace_time{"bthread_trace_time"};
+    int _signal_num{SIGURG};
```
wwbmmm, line 121: I think you can directly use FLAGS_signal_number_for_trace everywhere, don't need to add this class member.

src/bthread/task_tracer.cpp
```diff
@@ -35,6 +35,10 @@ namespace bthread {
 
 DEFINE_uint32(signal_trace_timeout_ms, 50, "Timeout for signal trace in ms");
 BUTIL_VALIDATE_GFLAG(signal_trace_timeout_ms, butil::PositiveInteger<uint32_t>);
+// Note that SIGURG handler may be registered by some library such as cgo
+// so let the signal number be configurable
+DEFINE_int32(signal_number_for_trace, SIGURG,
+             "signal number used for stack trace, default to SIGURG");
```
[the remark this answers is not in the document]
ZhengweiZhu: Add validator will make it configurable by user, which is not expected.

src/bthread/task_tracer.h
ZhengweiZhu: Both way works. This is to guarantee the signal number is immutable.  Should I change?

chenBright: > This is to guarantee the signal number is immutable.

I prefer this approach.

wwbmmm: > Both way works. This is to guarantee the signal number is immutable. Should I change?

That make sense. You don't need to change.

ZhengweiZhu:
ping~

wwbmmm:
LGTM

PR #3137, opened 2025-11-04
add pending signaled wrs checking

yanglimingcn:

Issue Number: #3132

Problem Summary:

Changed:

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/brpc/controller.cpp
+++ b/src/brpc/controller.cpp
@@ -232,6 +232,7 @@ void Controller::ResetNonPods() {
     }
     delete _remote_stream_settings;
     _thrift_method_name.clear();
+    _checksum_value.clear();
     _after_rpc_resp_fn = nullptr;
 
     CHECK(_unfinished_call == NULL);


src/brpc/rdma/rdma_endpoint.cpp
```diff
@@ -972,10 +1007,12 @@ ssize_t RdmaEndpoint::HandleCompletion(ibv_wc& wc) {
 
             // Update window
             uint32_t wnd_thresh = _local_window_capacity / 8;
-            if (_window_size.fetch_add(acks, butil::memory_order_relaxed) >= wnd_thresh
-                    || acks >= wnd_thresh) {
+            if (_pending_signaled_wrs.load(butil::memory_order_relaxed) <= 1 &&
```
chenBright: When `_pending_signaled_wrs` is greater than 1, `_window_size` is not updated, and acks will be lost.

src/brpc/rdma/rdma_endpoint.cpp
```diff
@@ -918,6 +950,7 @@ int RdmaEndpoint::SendImm(uint32_t imm) {
 
     ibv_send_wr wr;
     memset(&wr, 0, sizeof(wr));
+    wr.wr_id = GenerateWrId();
```
chenBright: Would it be sufficient to set `wr_id` to a fixed value for `Imm`?

yanglimingcn: Maybe, just use a fixed value

src/brpc/rdma/rdma_endpoint.cpp
```diff
@@ -889,6 +896,10 @@ ssize_t RdmaEndpoint::CutFromIOBufList(butil::IOBuf** from, size_t ndata) {
             return -1;
         }
 
+        if (signaled) {
```
yanglimingcn, line 898: this line, I'm not sure if send cqe can generate energy after this atomic operation.

src/brpc/rdma/rdma_endpoint.cpp
```diff
@@ -995,6 +1004,17 @@ ssize_t RdmaEndpoint::HandleCompletion(ibv_wc& wc) {
                      << wc.opcode;
         return -1;
     }
+    if (pending_signaled_wrs <= 0) {
```
chenBright, line 1007: It seems that _window_size may still be larger than the actual sq size.

chenBright (commented):
I think we should use two windows: one representing the local sending window size and the other representing the remote receiving window size. Sending should only occur if both window values ​​are greater than 0.

See #3145 . We encountered the same problem, and the above method solved it.

src/brpc/rdma/rdma_endpoint.cpp
yanglimingcn: Could you give an example to describe it?

yanglimingcn:
> I think we should use two windows: one representing the local sending window size and the other representing the remote receiving window size. Sending should only occur if both window values ​​are greater than 0.
> 
> See #3145 . We encountered the same problem, and the above method solved it.

Yes, I think your PR is better.

src/brpc/rdma/rdma_endpoint.cpp
chenBright: After the client sends _local_window_capacity / 4 + 1 requests, it receives an ACK and a send CQE from the server. At this point, _window_size increases by _local_window_capacity / 4 + 1, and sq size increases by _local_window_capacity / 4.

The issue of inconsistency between _window_size and sq size still exists.

PR #3140, opened 2025-11-06
Fix span lifecycle with smart pointers to prevent use-after-free in async RPC callbacks 

### LINKED ISSUE #3068: span.cpp中destroy()方法存在内存重复释放问题，会导致服务崩掉
**Describe the bug**
服务开启enable_rpcz后，上线跑一段时间就会内存暴涨然后core掉，经排查定位发现在destroy()中存在内存重复释放问题。

```
void Span::destroy() {
    EndAsParent();
    Span* p = _next_client;
    while (p) {
        if (p == this) {
            LOG(ERROR) << "Span next_client points to self, span_id=" << p->_span_id;
        }
        Span* p_next = p->_next_client;
        p->_info.clear();
        butil::return_object(p);
        p = p_next;
    }
    _info.clear();
    butil::return_object(this);
}
```
如上代码，ERROR日志是我debug时加的，经过验证在服务core掉时确实打印了该行日志。
说明span链表中有span异常指向了自身。

这个问题的根源是什么？新版本是否有解决该问题？

**To Reproduce**

**Expected behavior**

**Versions**
OS: centos7.6
Compiler: gcc4.8.5
brpc: 0.9.0
protobuf:

lh2debug:

Issue Number: resolve #3068 

Problem Summary:

Span lifecycle management defect in distributed storage system:

Problem Scenario:
- Server maintains a parent span for client Write requests
- Two child spans are created for append_entries RPCs to followers
- When the first follower responds, the parent span is prematurely destroyed along with all child spans
- When the second follower responds, it attempts to access the already-freed child span, causing use-after-free

Root Cause:
a. Premature deallocation: Parent span destroyed while child spans still in use
b. Dangling pointer: Response callback accesses freed span objects

#### Reference counting relationship
- Parent span holds strong references to child spans, while child spans hold weak references to parent span.
- RPC Done holds a strong reference to the parent span (i.e., server span), ensuring the reference count is released only after trace recording is completed when RPC Done executes.
- SpanContainer holds a strong reference to the parent span, ensuring the reference count is released only after the background collector thread completes dumping the trace to the database.
- Controller holds weak references to parent span/child spans, which does not affect the lifecycle of related spans while avoiding access to dangling pointers.
```mermaid
graph LR
    subgraph "Parent Span (WriteChunk)"
        PS["Parent Span<br/>ref_count = 2<br/>trace_id: 12345<br/>span_id: 67890"]
    end
    
    subgraph "强引用持有者 (shared_ptr)"
        RD["RPC Done Callback<br/>SendRpcResponse<br/>shared_ptr<Span>"]
        SC["SpanContainer<br/>后台collector<br/>shared_ptr<Span>"]
    end
    
    subgraph "弱引用持有者 (weak_ptr)"
        PC["Controller<br/>_span<br/>weak_ptr<Span>"]
        CS1["Child Span 1<br/>_local_parent<br/>weak_ptr<Span>"]
        CS2["Child Span 2<br/>_local_parent<br/>weak_ptr<Span>"]
    end
    
    subgraph "Child Spans"
        CS1_DETAIL["Child Span 1<br/>ref_count = 1<br/>append_entries to Follower1<br/>trace_id: 12345<br/>span_id: 67891"]
        CS2_DETAIL["Child Span 2<br/>ref_count = 1<br/>append_entries to Follower2<br/>trace_id: 12345<br/>span_id: 67892"]
    end
    
    %% Parent Span的强引用
    RD -->|"+1 ref_count"| PS
    SC -->|"+1 ref_count"| PS
    
    %% Parent Span的弱引用
    PC -.->|"不增加ref_count"| PS
    CS1 -.->|"不增加ref_count"| PS
    CS2 -.->|"不增加ref_count"| PS
    
    %% Parent Span持有Child Spans
    PS -->|"_client_list<br/>shared_ptr<br/>+1 ref_count"| CS1_DETAIL
    PS -->|"_client_list<br/>shared_ptr<br/>+1 ref_count"| CS2_DETAIL
    
    %% Child Spans的弱引用
    CS1_DETAIL -.->|"_local_parent<br/>weak_ptr"| PS
    CS2_DETAIL -.->|"_local_parent<br/>weak_ptr"| PS
    
    %% Controller对Child Spans的弱引用
    C1["Controller 1<br/>weak_ptr"] -.-> CS1_DETAIL
    C2["Controller 2<br/>weak_ptr"] -.-> CS2_DETAIL
    
    %% 样式
    classDef parentSpan fill:#e1f5fe,stroke:#01579b,stroke-width:4px
    classDef childSpan fill:#f3e5f5,stroke:#4a148c,stroke-width:3px
    classDef strongRef fill:#e8f5e8,stroke:#2e7d32,stroke-width:3px
    classDef weakRef fill:#fff3e0,stroke:#e65100,stroke-width:2px
    
    class PS parentSpan
    class CS1_DETAIL,CS2_DETAIL childSpan
    class RD,SC strongRef
    class PC,C1,C2 weakRef
```

Changed:

Side effects: NO
- Performance effects: NO

- Breaking backward compatibility: NO

---
### Check List:

### DIFF
--- a/src/brpc/builtin/rpcz_service.cpp
+++ b/src/brpc/builtin/rpcz_service.cpp
@@ -211,7 +211,15 @@ static void PrintAnnotations(
             PrintElapse(os, anno_time, last_time);
             os << ' ';
             if (span) {
-                os << '[' << span_type_str << ' ' << SPAN_ID_STR << '=' << Hex(span->span_id()) << "] ";
+                const char* short_type = "SPAN";
+                if (span->type() == SPAN_TYPE_SERVER) {
+                    short_type = "Server";
+                } else if (span->type() == SPAN_TYPE_CLIENT) {
+                    short_type = "Client";
+                } else if (span->type() == SPAN_TYPE_BTHREAD) {
+                    short_type = "Bthread";
+                }
+                os << '[' << short_type << " SPAN#" << Hex(span->span_id()) << "] ";
             }
             os << WebEscape(a);
             if (a.empty() || butil::back_char(a) != '\n') {
@@ -292,11 +300,11 @@ static void PrintClientSpan(
 
     if (PrintAnnotationsAndRealTimeSpan(os, span.sent_real_us(),
                                         last_time, extr, num_extr, &span)) {
-        os << " [ClientSpan " << SPAN_ID_STR << '=' << Hex(span.span_id()) << "] Requested(" << span.request_size() << ") [1]" << std::endl;
+        os << " [Client SPAN#" << Hex(span.span_id()) << "] Requested(" << span.request_size() << ") [1]" << std::endl;
     }
     if (PrintAnnotationsAndRealTimeSpan(os, span.received_real_us(),
                                         last_time, extr, num_extr, &span)) {
-        os << " [ClientSpan " << SPAN_ID_STR << '=' << Hex(span.span_id()) << "] Received response(" << span.response_size() << ")";
+        os << " [Client SPAN#" << Hex(span.span_id()) << "] Received response(" << span.response_size() << ")";
         if (span.base_cid() != 0 && span.ending_cid() != 0) {
             int64_t ver = span.ending_cid() - span.base_cid();
             if (ver >= 1) {
@@ -310,13 +318,13 @@ static void PrintClientSpan(
 
     if (PrintAnnotationsAndRealTimeSpan(os, span.start_parse_real_us(),
                                         last_time, extr, num_extr, &span)) {
-        os << " [ClientSpan " << SPAN_ID_STR << '=' << Hex(span.span_id()) << "] Processing the response in a new bthread" << std::endl;
+        os << " [Client SPAN#" << Hex(span.span_id()) << "] Processing the response in a new bthread" << std::endl;
     }
 
     if (PrintAnnotationsAndRealTimeSpan(
             os, span.start_callback_real_us(),
             last_time, extr, num_extr, &span)) {
-        os << " [ClientSpan " << SPAN_ID_STR << '=' << Hex(span.span_id()) << "] " << (span.async() ? " Enter user's done" : " Back to user's callsite") << std::endl;
+        os << " [Client SPAN#" << Hex(span.span_id()) << "] " << (span.async() ? " Enter user's done" : " Back to user's callsite") << std::endl;
     }
 
     PrintAnnotations(os, std::numeric_limits<int64_t>::max(),
@@ -340,9 +348,9 @@ static void PrintBthreadSpan(std::ostream& os, const RpczSpan& span, int64_t* la
     extr[num_extr++] = &client_extr;
 
     // Print span id for bthread span context identification
-    os << " [BthreadSpan " << SPAN_ID_STR << '=' << Hex(span.span_id());
+    os << " [Bthread SPAN#" << Hex(span.span_id());
     if (span.parent_span_id() != 0) {
-        os << " parent_span=" << Hex(span.parent_span_id());
+        os << " parent#" << Hex(span.parent_span_id());
     }
     os << "] ";
 
@@ -377,7 +385,7 @@ static void PrintServerSpan(std::ostream& os, const RpczSpan& span,
     if (PrintAnnotationsAndRealTimeSpan(
             os, span.start_parse_real_us(),
             &last_time, extr, ARRAY_SIZE(extr), &span)) {
-        os << " [ServerSpan " << SPAN_ID_STR << '=' << Hex(span.span_id()) << "] Processing the request in a new bthread" << std::endl;
+        os << " [Server SPAN#" << Hex(span.span_id()) << "] Processing the request in a new bthread" << std::endl;
     }
 
     bool entered_user_method = false;
@@ -385,7 +393,7 @@ static void PrintServerSpan(std::ostream& os, const RpczSpan& span,
             os, span.start_callback_real_us(),
             &last_time, extr, ARRAY_SIZE(extr), &span)) {
         entered_user_method = true;
-        os << " [ServerSpan " << SPAN_ID_STR << '=' << Hex(span.span_id()) << "] Enter " << WebEscape(span.full_method_name()) << std::endl;
+        os << " [Server SPAN#" << Hex(span.span_id()) << "] Enter " << WebEscape(span.full_method_name()) << std::endl;
     }
 
     const int nclient = span.client_spans_size();
@@ -402,16 +410,16 @@ static void PrintServerSpan(std::ostream& os, const RpczSpan& span,
             os, span.start_send_real_us(),
             &last_time, extr, ARRAY_SIZE(extr), &span)) {
         if (entered_user_method) {
-            os << " [ServerSpan " << SPAN_ID_STR << '=' << Hex(span.span_id()) << "] Leave " << WebEscape(span.full_method_name()) << std::endl;
+            os << " [Server SPAN#" << Hex(span.span_id()) << "] Leave " << WebEscape(span.full_method_name()) << std::endl;
         } else {
-            os << " [ServerSpan " << SPAN_ID_STR << '=' << Hex(span.span_id()) << "] Responding" << std::endl;
+            os << " [Server SPAN#" << Hex(span.span_id()) << "] Responding" << std::endl;
         }
     }
     
     if (PrintAnnotationsAndRealTimeSpan(
             os, span.sent_real_us(),
             &last_time, extr, ARRAY_SIZE(extr), &span)) {
-        os << " [ServerSpan " << SPAN_ID_STR << '=' << Hex(span.span_id()) << "] Responded(" << span.response_size() << ')' << std::endl;
+        os << " [Server SPAN#" << Hex(span.span_id()) << "] Responded(" << span.response_size() << ')' << std::endl;
     }
 
     PrintAnnotations(os, std::numeric_limits<int64_t>::max(),
--- a/src/brpc/controller.cpp
+++ b/src/brpc/controller.cpp
@@ -1405,6 +1405,7 @@ uint64_t Controller::trace_id() const {
     }
     return 0;
 }
+
 uint64_t Controller::span_id() const {
     if (auto span = _span.lock()) {
         return span->span_id();
@@ -1735,6 +1736,26 @@ void Controller::DoPrintLogPrefix(std::ostream& os) const {
     if (FLAGS_log_as_json) {
         os << "\"M\":\"";
     }
+}
+
+
+ControllerPrivateAccessor& ControllerPrivateAccessor::set_span(
+    const std::shared_ptr<Span>& span) {
+    _cntl->_span = span;
+    return *this;
+}
+
+ControllerPrivateAccessor& ControllerPrivateAccessor::set_span(Span* span) {
+    if (span) {
+        _cntl->_span = span->shared_from_this();
+    } else {
+        _cntl->_span.reset();
+    }
+    return *this;
+}
+
+std::shared_ptr<Span> ControllerPrivateAccessor::span() const {
+    return _cntl->_span.lock();
 }
 
 } // namespace brpc
--- a/src/brpc/span.cpp
+++ b/src/brpc/span.cpp
@@ -62,8 +62,7 @@ void DestroyRpczParentSpan(void* ptr) {
 void EndBthreadSpan() {
     std::shared_ptr<Span> span = GetTlsParentSpan();
     if (span) {
-        bthread_id_t id = {bthread_self()};
-        span->set_ending_cid(id);
+        span->set_ending_tid(bthread_self());
     }
 
     ClearTlsParentSpan();
@@ -73,10 +72,11 @@ void SetTlsParentSpan(std::shared_ptr<Span> span) {
     using namespace bthread;
     LocalStorage ls = BAIDU_GET_VOLATILE_THREAD_LOCAL(tls_bls);
     if (ls.rpcz_parent_span) {
-        delete static_cast<std::weak_ptr<Span>*>(ls.rpcz_parent_span);
+        *static_cast<std::weak_ptr<Span>*>(ls.rpcz_parent_span) = span;
+    } else {
+        ls.rpcz_parent_span = new std::weak_ptr<Span>(span);
+        BAIDU_SET_VOLATILE_THREAD_LOCAL(tls_bls, ls);
     }
-    ls.rpcz_parent_span = new std::weak_ptr<Span>(span);
-    BAIDU_SET_VOLATILE_THREAD_LOCAL(tls_bls, ls);
 }
 
 std::shared_ptr<Span> GetTlsParentSpan() {
@@ -94,9 +94,7 @@ void ClearTlsParentSpan() {
     using namespace bthread;
     LocalStorage ls = BAIDU_GET_VOLATILE_THREAD_LOCAL(tls_bls);
     if (ls.rpcz_parent_span) {
-        delete static_cast<std::weak_ptr<Span>*>(ls.rpcz_parent_span);
-        ls.rpcz_parent_span = nullptr;
-        BAIDU_SET_VOLATILE_THREAD_LOCAL(tls_bls, ls);
+        static_cast<std::weak_ptr<Span>*>(ls.rpcz_parent_span)->reset();
     }
 }
 
@@ -122,9 +120,6 @@ void SpanDeleter::operator()(Span* r) const {
     // children.
     r->_client_list.clear();
     r->_info.clear();
-    // Destroy the spinlocks, as the destructor might not be invoked.
-    pthread_spin_destroy(&r->_client_list_spinlock);
-    pthread_spin_destroy(&r->_info_spinlock);
     butil::return_object(r);
 }
 
@@ -200,7 +195,8 @@ Span::Span(Forbidden) {
 }
 
 Span::~Span() {
-    // The destruction of the spinlock has been handled in SpanDeleter.
+    pthread_spin_destroy(&_client_list_spinlock);
+    pthread_spin_destroy(&_info_spinlock);
 }
 
 std::shared_ptr<Span> Span::CreateClientSpan(const std::string& full_method_name,
@@ -212,7 +208,7 @@ std::shared_ptr<Span> Span::CreateClientSpan(const std::string& full_method_name
     std::shared_ptr<Span> span(span_raw, SpanDeleter());
     span->_log_id = 0;
     span->_base_cid = INVALID_BTHREAD_ID;
-    span->_ending_cid = INVALID_BTHREAD_ID;
+    span->_ending_cid = INVALID_BTHREAD_ID;  // Client Span uses ending_cid
     span->_type = SPAN_TYPE_CLIENT;
     span->_async = false;
     span->_protocol = PROTOCOL_UNKNOWN;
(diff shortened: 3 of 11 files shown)


lh2debug:
Yes. For example, in the Raft protocol, there is one leader and two followers. The leader only needs to wait for the first follower to complete before returning.

src/bthread/task_meta.h
```diff
@@ -43,10 +48,10 @@ struct ButexWaiter;
 struct LocalStorage {
     KeyTable* keytable;
     void* assigned_data;
-    void* rpcz_parent_span;
+    std::weak_ptr<brpc::Span> rpcz_parent_span;
```
wwbmmm: I think you 'd better not use brpc types in the bthread code.
In the brpc design, bthread is a basic library, and brpc is built on top of it. So bthread should no depend on brpc. This is why the original type is `void*`, not `brpc::Span*`.
My suggestion: you can keep `void* rpcz_parent_span` here, and in the brpc code, you can new a `std::weak_ptr<brpc::Span>`, then cast the pointer `std::weak_ptr<brpc::Span>*` to `void*`. This may be more complicated, but keep the brpc design clean.

lh2debug:
Thank you for your answer. This approach of implementing reference counting internally within a Span seems impractical.

Because reference Counting with void* rpcz_parent_span Fails Across Bthreads:
Raw pointers in LocalStorage cannot detect when the referenced Span is destroyed in another bthread. When a parent Span is deleted (after its reference count reaches zero), child bthreads still hold the same pointer value pointing to freed memory, causing use-after-free. Each Span's reference count only tracks its own operations and cannot prevent destruction when other bthreads still reference it through raw pointers.

Smart pointers solve this via control blocks that survive object destruction, allowing safe lifetime detection across threads.

@yanglimingcn

yanglimingcn:
I believe that spans are suitable for requesting related information and should not be more closely associated with threads or coroutines; they simply use thread-local storage to pass variables in their implementation.

wwbmmm:
> The span information is a tree structure, and this relationship is maintained through the ref information. For example, ParentSpan initially has ref=1, and two ClientSpans are created, so ref becomes 3.
> 
> When one ClientSpan returns, ref becomes 2; when the other ClientSpan returns, ref becomes 1. At this point, it can be destructed.

Let me complement, the first reference to the ParentSpan is held by tls, when the bthread exit, ref_count will -1, when the ref_count become 0 (only after all ClientSpan return and the bthread exit) , it can be destructed.

Moreover, when a new bthread with BTHREAD_INHERIT_SPAN is created, the ref_count will +1; when this bthread exit, the ref_count will -1.

lh2debug:
4. To ensure compatibility with advanced clients who modify their protocols, using the `BRPC_ENABLE_SHARED_SPAN` macro to control the use of `Span*` or `std::weak_ptr<Span>` in the `ControllerPrivateAccessor` layer can resolve compilation issues. However, the `Span*` approach may still cause crashes in certain scenarios.

If the implementation follows the `BRPC_ENABLE_SHARED_SPAN` macro control, could it mislead them into believing that the `Span*` approach is stable and reliable for performance analysis?

wwbmmm:
> 4. To ensure compatibility with advanced clients who modify their protocols, using the `BRPC_ENABLE_SHARED_SPAN` macro to control the use of `Span*` or `std::weak_ptr<Span>` in the `ControllerPrivateAccessor` layer can resolve compilation issues. However, the `Span*` approach may still cause crashes in certain scenarios.
> 
> If the implementation follows the `BRPC_ENABLE_SHARED_SPAN` macro control, could it mislead them into believing that the `Span*` approach is stable and reliable for performance analysis?

We can add some notes in the document and code comment. I think that's OK.
Moreover, it's not only for `ControllerPrivateAccessor`, if possible, please apply this approach to each public method you changed.

src/brpc/builtin/rpcz_service.cpp
```diff
@@ -372,22 +400,22 @@ static void PrintServerSpan(std::ostream& os, const RpczSpan& span,
...
-            os << " Responding" << std::endl;
+            os << " [ServerSpan " << SPAN_ID_STR << '=' << Hex(span.span_id()) << "] Responding" << std::endl;
         }
     }
     
     if (PrintAnnotationsAndRealTimeSpan(
             os, span.sent_real_us(),
-            &last_time, extr, ARRAY_SIZE(extr))) {
-        os << " Responded(" << span.response_size() << ')' << std::endl;
+            &last_time, extr, ARRAY_SIZE(extr), &span)) {
```
yanglimingcn, line 421: Could you please post some of the display information?

src/brpc/controller.cpp
```diff
@@ -1714,4 +1727,31 @@ void Controller::DoPrintLogPrefix(std::ostream& os) const {
...
+    return *this;
+}
+#endif
+
+SpanPtr ControllerPrivateAccessor::span() const {
+    std::shared_ptr<Span> span_internal = _cntl->_span.lock();
+#if BRPC_SPAN_ENABLE_SHARED_PTR_API
+    return span_internal;
+#else
+    return span_internal.get();
```
yanglimingcn: return Span* not hold a reference, This will cause problems, won't it?

src/brpc/policy/baidu_rpc_protocol.cpp
```diff
@@ -269,9 +269,9 @@ struct BaiduProxyPBMessages : public RpcPBMessages {
 // Used by UT, can't be static.
 void SendRpcResponse(int64_t correlation_id, Controller* cntl,
                      RpcPBMessages* messages, const Server* server,
-                     MethodStatus* method_status, int64_t received_us) {
+                     MethodStatus* method_status, int64_t received_us,
+                     std::shared_ptr<Span> span) {
```
yanglimingcn, line 276: Why SendRpcResponse need span parameter, It can be obtained from the accessor?

src/brpc/span.h
```diff
@@ -181,11 +269,38 @@ friend class SpanDB;
...
+    // In some scenarios, multiple bthreads may simultaneously create child spans
+    // (e.g.,raft leader parallel RPCs to followers) and push_back to parent's _client_list.
+    // Also protects against concurrent iteration (e.g., CountClientSpans, SpanDB::Index)
+    // while the list is being modified.
+    mutable pthread_spinlock_t _client_list_spinlock;
+};
+
+class SpanContainer : public bvar::Collected {
+public:
+    explicit SpanContainer(std::shared_ptr<Span> span) : _span(span) {}
```
yanglimingcn: const std::shared_ptr<Span>& span ?

src/bthread/key.cpp
```diff
@@ -405,6 +417,11 @@ static void cleanup_pthread(void* arg) {
         delete kt;
         // After deletion: tls may be set during deletion.
         tls_bls.keytable = NULL;
+
```
yanglimingcn: What is the relationship between rpcz_parent_span and key.cpp?

src/bthread/key.cpp
```diff
@@ -672,4 +694,19 @@ void* bthread_get_assigned_data() {
     return bthread::tls_bls.assigned_data;
 }
 
+int bthread_set_span_funcs(bthread_create_span_fn create_fn,
+                            bthread_destroy_span_fn destroy_fn,
```
yanglimingcn: Would it be more reasonable to put this function in bthread.cpp?

src/brpc/span.cpp
```diff
@@ -37,9 +37,97 @@
...
+// Returns a heap-allocated weak_ptr<Span>* as void*, or NULL if span creation fails.
+void* CreateBthreadSpanAsVoid() {
+    const int64_t received_us = butil::cpuwide_time_us();
+    const int64_t base_realtime = butil::gettimeofday_us() - received_us;
+    std::shared_ptr<Span> span = Span::CreateBthreadSpan("Bthread", base_realtime);
+
+    if (!span) {
+        return NULL;
+    }
+    return new std::weak_ptr<Span>(span);
```
yanglimingcn, line 53: When is this newly created `weak_ptr` deleted? This part seems a bit obscure, and it's easy for memory leaks to occur.

src/brpc/policy/baidu_rpc_protocol.cpp
lh2debug: SendRpcResponse needs to hold a strong reference count; otherwise, if the backend spancontainer releases the reference, the server span will be unable to set_start_send_us and set_response_size.

src/brpc/controller.cpp
lh2debug: If the server span waits for the child span to complete before returning (as is the case in most business scenarios), returning a raw pointer won't cause any problems.

For example, in a Raft scenario, if the server span returns before the second follower completes (causing the second follower's client span to be destroyed), the second follower's RPC callback will retrieve the destroyed dangling pointer through `ControllerPrivateAccessor::span`.

src/brpc/span.h
lh2debug: Okay, I'll change this later.

src/brpc/builtin/rpcz_service.cpp
lh2debug: As shown below, the type of span (server span, client span, bthread span) is indicated at the beginning to facilitate troubleshooting.

```
Fetching new trace TRACE#B from HOST#X

Received request REQ#65674 from HOST#Y via PROTO#1 log=0 trace=TRACE#B span=SPAN#S1

[ServerSpan span=SPAN#S1] Processing the request in a new worker
[ServerSpan span=SPAN#S1] ParseStats[cut=?, queue=?, worker=?, rpc=?]
[ServerSpan span=SPAN#S1] Enter Ecom.OrderService.PlaceOrder

[ClientSpan span=SPAN#C1] Cart.AddItem item=ITEM# qty=QTY# price=PRICE# user=USER# session=SESSION#
[ClientSpan span=SPAN#C1] Cart.VerifyOperation item=ITEM# op=OP# user=USER# session=SESSION#
[ClientSpan span=SPAN#C1] Cart.ProposeUpdate item=ITEM# op=OP# user=USER# session=SESSION# attempt=SN#
[ClientSpan span=SPAN#C1] Orchestration.Manager: enqueue step=PlaceOrder first=I# last=I#
[ClientSpan span=SPAN#C1] Orchestration.Queue: pending step

Requesting Ecom.Consensus.Append@HOST#Z PROTO#1 call=CALL# trace=TRACE#B span=SPAN#C1
[ClientSpan span=SPAN#C1] Requested(REQ#65778) [1]
[ClientSpan span=SPAN#C1] Received response(RSP#) of request[1]
[ClientSpan span=SPAN#C1] Processing the response in a new worker
[ClientSpan span=SPAN#C1] Enter client callback

[ClientSpan span=SPAN#C1] Workflow.OnApply, ongoing tasks: 0
[ClientSpan span=SPAN#C1] TaskQueue status running=R# queues=Q# active=A#
[ClientSpan span=SPAN#C1] OrderPipeline.WriteOrder
[ClientSpan span=SPAN#C1] Payment.Authorize
[ClientSpan span=SPAN#C1] Inventory.Reserve
[ClientSpan span=SPAN#C1] Promotion.Apply
[ClientSpan span=SPAN#C1] Promotion.Apply_done
[ClientSpan span=SPAN#C1] Notification.Prepare
[ClientSpan span=SPAN#C1] Notification.Prepare_done
[ClientSpan span=SPAN#C1] Inventory.Reserve_done
[ClientSpan span=SPAN#C1] Payment.Authorize_done
[ClientSpan span=SPAN#C1] OrderPipeline.Sync
[ClientSpan span=SPAN#C1] OrderPipeline.WriteOrder done (~2.7ms)
[ClientSpan span=SPAN#C1] Latency stats: avg≈58us p90≈66us p99≈160us
[ClientSpan span=SPAN#C1] Protocol response before join
[ClientSpan span=SPAN#C1] Protocol response after join

[ClientSpan span=SPAN#C1] FulfillmentWorker: ongoing tasks=3
[ClientSpan span=SPAN#C1] FulfillmentWorker: global queue running=R# queues=Q# active=A#
[ClientSpan span=SPAN#C1] FulfillmentWorker: enter
[ClientSpan span=SPAN#C1] Batcher.append
[ClientSpan span=SPAN#C1] Batcher.flush queue_wait≈~6.7ms

Requesting Ecom.Consensus.Append@HOST#Y PROTO#1 call=CALL# trace=TRACE#B span=SPAN#C2
[ClientSpan span=SPAN#C2] Requested(REQ#65778) [1]
[ClientSpan span=SPAN#C2] Received response(RSP#) of request[1]
[ClientSpan span=SPAN#C2] Processing the response in a new worker
[ClientSpan span=SPAN#C2] Enter client callback

[ServerSpan span=SPAN#S1] Leave Ecom.OrderService.PlaceOrder
[ServerSpan span=SPAN#S1] Responded
```

wwbmmm: This format is a little bit tedious. How about `[Server SPAN#S1] Processing the request in a new worker`?

src/brpc/controller.cpp
```diff
@@ -1387,8 +1390,18 @@ const Controller* Controller::sub(int index) const {
...
 
-uint64_t Controller::trace_id() const { return _span ? _span->trace_id() : 0; }
-uint64_t Controller::span_id() const { return _span ? _span->span_id() : 0; }
+uint64_t Controller::trace_id() const {
+    if (auto span = _span.lock()) {
+        return span->trace_id();
+    }
+    return 0;
+}
+uint64_t Controller::span_id() const {
```
wwbmmm, line 1409: add a new line

src/brpc/policy/baidu_rpc_protocol.cpp
```diff
@@ -642,12 +642,16 @@ void ProcessRpcRequest(InputMessageBase* msg_base) {
...
-    Span* span = NULL;
+    std::shared_ptr<Span> span;
     if (IsTraceable(request_meta.has_trace_id())) {
         span = Span::CreateServerSpan(
             request_meta.trace_id(), request_meta.span_id(),
             request_meta.parent_span_id(), msg->base_real_us());
+#if BRPC_SPAN_ENABLE_SHARED_PTR_API
         accessor.set_span(span);
+#else
+        accessor.set_span(span.get());
```
wwbmmm: you can add an overloaded function `set_span(std::shared_ptr<Span> span)`,
so that you don't need to use `BRPC_SPAN_ENABLE_SHARED_PTR_API` everywhere.

src/brpc/span.cpp
```diff
@@ -37,9 +37,97 @@
...
+    ClearTlsParentSpan();
+}
+
+void SetTlsParentSpan(std::shared_ptr<Span> span) {
+    using namespace bthread;
+    LocalStorage ls = BAIDU_GET_VOLATILE_THREAD_LOCAL(tls_bls);
+    if (ls.rpcz_parent_span) {
+        delete static_cast<std::weak_ptr<Span>*>(ls.rpcz_parent_span);
+    }
+    ls.rpcz_parent_span = new std::weak_ptr<Span>(span);
```
wwbmmm: I don't know why you need to delete and new the weak_ptr, can you directly reset the weak_ptr like `*ls.rpcz_parent_span = span`?

src/brpc/span.cpp
```diff
@@ -37,9 +37,97 @@
...
+
+    auto* weak_ptr = static_cast<std::weak_ptr<Span>*>(ls.rpcz_parent_span);
+    return weak_ptr->lock();
+}
+
+void ClearTlsParentSpan() {
+    using namespace bthread;
+    LocalStorage ls = BAIDU_GET_VOLATILE_THREAD_LOCAL(tls_bls);
+    if (ls.rpcz_parent_span) {
+        delete static_cast<std::weak_ptr<Span>*>(ls.rpcz_parent_span);
```
wwbmmm: I think you don't need to delete here, just `ls.rpcz_parent_span->reset()`

src/bthread/key.cpp
wwbmmm: I don't think `key.cpp` is responsible for span's lifecycle triggering points.
`key.cpp` is for the implementation of `bthread_get/setspecific`, it manage keytable_pool and keytable, which has no relationship to span.
What you need is to call the span functions in the bthread lifecycle (create/end/...), the better place is in the `task_group.cpp`, you can get/set bthread::tls_bls.rpcz_parent_span there.
And the implementation of `bthread_set_span_funcs` is better placed into `bthread.cpp`, because most bthread interface functions are there.

chenBright:
@lh2debug Friendly ping!

lh2debug:
Okay, I'll think about these suggestions. Thank you.

src/brpc/span.cpp
```diff
@@ -37,9 +37,96 @@
...
+void DestroyRpczParentSpan(void* ptr) {
+    if (ptr) {
+        delete static_cast<std::weak_ptr<Span>*>(ptr);
+    }
+}
+
+void EndBthreadSpan() {
+    std::shared_ptr<Span> span = GetTlsParentSpan();
+    if (span) {
+        bthread_id_t id = {bthread_self()};
```
chenBright: Why is bthread_t set to ending_cid?

src/brpc/span.cpp
```diff
@@ -37,9 +37,96 @@
...
+    }
+
+    // All children will be destroyed automatically along with the list.
+    // The list holds std::shared_ptr<> which will trigger deletion of
+    // children.
+    r->_client_list.clear();
+    r->_info.clear();
+    // Destroy the spinlocks, as the destructor might not be invoked.
+    pthread_spin_destroy(&r->_client_list_spinlock);
+    pthread_spin_destroy(&r->_info_spinlock);
```
chenBright: It appears that the spinlock is only initialized once in the Span constructor, but is destroyed once before each SpanDeleter::operator() call.

src/bthread/task_group.cpp
```diff
@@ -495,7 +507,11 @@ int TaskGroup::start_foreground(TaskGroup** pg,
     m->attr = using_attr;
     m->local_storage = LOCAL_STORAGE_INIT;
     if (using_attr.flags & BTHREAD_INHERIT_SPAN) {
-        m->local_storage.rpcz_parent_span = run_create_span_func();
+        if (g_create_bthread_span) {
+            m->local_storage.rpcz_parent_span = g_create_bthread_span();
```
chenBright, line 511: Who has a strong reference to bthread span?

Furthermore, after resolving the thread safety issues of spans, is it still necessary to use bthread spans?

chenBright (commented):
@lh2debug Please resolve conflicts in channel.cpp again and update the rpcz.md document.

src/brpc/span.cpp
chenBright: > (the lock is destroyed, and it is returned to the object pool)
> Subsequently, only new spans will be retrieved from the object pool for use.

At this point, when a pre-constructed span is retrieved from the object pool, its spinlock will not be initialized, but the spinlock was destroyed during the previous return.

src/brpc/span.cpp
chenBright: I think bthread_t is not suitable as ending_cid because they are of different types, and the value of bthread_t may conflict with the value of bthread_id_t.

How about adding a _ending_tid to be used as the ending id of the bthread span?

src/bthread/task_group.cpp
chenBright: > The parent Span's _client_list

What if the parent span is NULL?

src/brpc/span.cpp
chenBright: These two branches will not execute `placement new`, but will directly return the existing object.
https://github.com/apache/brpc/blob/0565d8d2c7a1cde87e94aa222fe4946e6f20c358/src/butil/object_pool_inl.h#L183-L193

src/bthread/task_group.cpp
chenBright: If the parent span was not returned, but never existed, it seems impossible to create a bthread span.

src/brpc/span.cpp
chenBright: `pthread_spin_destroy` needs to be called in the destructor.

src/bthread/task_group.cpp
lh2debug: Yes. Currently, if a bthread is created directly in a non-rpc context, the created bthread span will be destroyed immediately because there is no strong reference count. If you want to collect bthread span out of rpc context, the business layer needs to call CreateBthreadSpan and hold its strong reference count.

chenBright: OK.

PR #3145, opened 2025-11-09
Bugfix: SQ overflow

### LINKED ISSUE #3132: 滑动窗口bug：Fail to ibv_post_send: Cannot allocate memory, window=3, sq_current=98
**Describe the bug**
此错误代表：滑动窗口表示可发送，但RDMA的SQ（Send Queue）没有可用的空间发送请求；

滑动窗口代表发送方可继续发送的能力；当前，滑动窗口的更新时机为：poll到带有imm（也就是带有acks）的recv cqe，并且进行对应的处理时；
RDMA的SQ是一个环形buffer，其上存放发送但未完成的请求，SQ的指针更新时机为  poll 到 send cqe时；
这里我们可以发现：SQ和滑动窗口的更新时机不一致；

按照正常逻辑来讲，如果处理到带有imm的recv cqe，那么代表此前发送方所发出的RDMA Send已经完成，在recv cqe之前应该先生成send cqe；

如下图示意：

这是brpc期望的一个逻辑，在更新滑动窗口值前，先更新了SQ；

Image

但遇到该问题时，说明情况是这样的：
发送方调用大量RDMA Send操作，占用完SQ；
接着发送方poll到带有imm的recv cqe（并且此时一定没有其他send cqe，如果有的话，按照brpc一次poll 32个cqe，也会触发SQ指针更新），进入处理逻辑更新滑动窗口值；
更新窗口值后，某个线程发现窗口值非0，尝试发送，但SQ没有可用空间存放请求了，于是报错；

为什么在带有imm的recv cqe之前没有send cqe呢？
根据brpc的逻辑，带有imm就代表接收方收到数据（意味着RDMA Send已经完成），并向发送方发送确认；
而此时发送方已经生成带有imm的recv cqe，那么为什么已经完成的RDMA Send还没生成send cqe？

查看了IB spec，不同WQ共享CQ他们生成cqe是不能确保顺序的，只能确保同一个WQ内部的cqe生成顺序：

Image

所以，brpc的SQ和RQ共享了一个CQ后，不能确保发送方收到带有imm的recv cqe之前，一定生成send cqe；

可以看如下示意图：

Image

因此我认为此问题由两个原因共同导致：
1.滑动窗口更新时机和RDMA SQ更新时机不一致；
2.cqe生成顺序的不确定性。

**To Reproduce**
由业务触发，本地测试程序没有复现成功。不过发现日志中报错时业务均在发送大量数据。

**Expected behavior**
我们期望的是：滑动窗口更新时机与RDMA SQ的更新时机保持一致；
目前我们实现了一个patch，保证滑动窗口更新时机一定在RDMA SQ的更新时机之后。（确保RDMA SQ有空间再更新滑动窗口）
我们新增了_remote_acks记录接收方确认的数目；因为同一个WQ内是保证顺序的，我们使用_sq_to_update队列记录unsignaled的数目，使用_sq_update_flag队列记录已经poll到的send cqe数目；
当处理带有imm的recv cqe时，判断此前是否poll到过send cqe，若没有poll到，仅将此次的acks累加到_remote_acks中，不进行窗口值更新。

分析：
此patch不影响正常逻辑（即先生成send cqe，后生成recv cqe），这种情况在此patch下没有任何区别；
当先生成recv cqe，后生成send cqe时，此patch强制要求窗口值更新在send cqe后，保证SQ不会溢出。

```
diff --git a/src/brpc/rdma/rdma_endpoint.cpp b/src/brpc/rdma/rdma_endpoint.cpp
index 1d502a98..dac82bb7 100644
--- a/src/brpc/rdma/rdma_endpoint.cpp
+++ b/src/brpc/rdma/rdma_endpoint.cpp

@@ -101,7 +114,7 @@ static const uint32_t ACK_MSG_RDMA_OK = 0x1;

 static butil::Mutex* g_rdma_resource_mutex = NULL;
 static RdmaResource* g_rdma_resource_list = NULL;
-
+static butil::atomic<uint64_t> g_wr_id(0);
 struct HelloMessage {
     void Serialize(void* data) const;
     void Deserialize(void* data);
@@ -191,7 +204,11 @@ RdmaEndpoint::RdmaEndpoint(Socket* s)
     , _remote_window_capacity(0)
     , _window_size(0)
     , _new_rq_wrs(0)
+    , _remote_acks(0)
+    , _m_sq_unsignaled(0)
 {
+    LOG(INFO) << "_remote_acks: " << _remote_acks.load();
+    LOG(INFO) << "_m_sq_unsignaled: " << _m_sq_unsignaled;
     if (_sq_size < MIN_QP_SIZE) {
         _sq_size = MIN_QP_SIZE;
     }
@@ -208,6 +225,11 @@ RdmaEndpoint::RdmaEndpoint(Socket* s)
 }

 RdmaEndpoint::~RdmaEndpoint() {
+    LOG(INFO) << _window_size << " " << _remote_acks << " " << _sq_update_flag.size() << " " << _sq_to_update.size() << " " << _sq_unsignaled << " " << _sq_unsignaled;
+    while(_sq_to_update.empty() == false) {
+        LOG(INFO) << _sq_to_update.front();
+       _sq_to_update.pop();
+    }
     Reset();
     bthread::butex_destroy(_read_butex);
 }
@@ -231,6 +253,8 @@ void RdmaEndpoint::Reset() {
     _new_rq_wrs = 0;
     _sq_sent = 0;
     _rq_received = 0;
+    _remote_acks.store(0, butil::memory_order_relaxed);
+    _m_sq_unsignaled.store(0, butil::memory_order_relaxed);
 }

 void RdmaConnect::StartConnect(const Socket* socket,
@@ -878,15 +902,36 @@ ssize_t RdmaEndpoint::CutFromIOBufList(butil::IOBuf** from, size_t ndata) {
         }

         ibv_send_wr* bad = NULL;
-        int err = ibv_post_send(_resource->qp, &wr, &bad);
-        if (err != 0) {
-            // We use other way to guarantee the Send Queue is not full.
-            // So we just consider this error as an unrecoverable error.
-            LOG(WARNING) << "Fail to ibv_post_send: " << berror(err)
+
+        wr.wr_id = g_wr_id.fetch_add(1, butil::memory_order_relaxed);
+        {
+            BAIDU_SCOPED_LOCK(_m_sq_mutex);
+            int err = ibv_post_send(_resource->qp, &wr, &bad);
+            if (err != 0) {
+                // We use other way to guarantee the Send Queue is not full.
+                // So we just consider this error as an unrecoverable error.
+                LOG(WARNING) << "Fail to ibv_post_send: " << berror(err)
                          << ", window=" << window
                          << ", sq_current=" << _sq_current;
                 errno = err;
                 return -1;
+            }
+            _m_sq_unsignaled.fetch_add(1, butil::memory_order_release);
+            uint16_t cur_unsignaled = 0;
+            if(wr.send_flags & IBV_SEND_SIGNALED) {
+                cur_unsignaled = _m_sq_unsignaled.exchange(0, butil::memory_order_acquire);
+            }
+            if(cur_unsignaled != 0) {
+                BAIDU_SCOPED_LOCK(_sq_update_mutex);
+                _sq_to_update.push(cur_unsignaled);
+               //LOG(INFO) << "send signaled before: " << cur_unsignaled;
+            }
         }

         ++_sq_current;
@@ -924,13 +969,26 @@ int RdmaEndpoint::SendImm(uint32_t imm) {
     wr.send_flags |= IBV_SEND_SIGNALED;

     ibv_send_wr* bad = NULL;
-    int err = ibv_post_send(_resource->qp, &wr, &bad);
-    if (err != 0) {
-        // We use other way to guarantee the Send Queue is not full.
-        // So we just consider this error as an unrecoverable error.
-        LOG(WARNING) << "Fail to ibv_post_send: " << berror(err);
-        return -1;
+    wr.wr_id = g_wr_id.fetch_add(1, butil::memory_order_relaxed);
+    uint16_t cur_unsignaled = 0;
+    {
+        BAIDU_SCOPED_LOCK(_m_sq_mutex);
+        int err = ibv_post_send(_resource->qp, &wr, &bad);
+        if (err != 0) {
+            // We use other way to guarantee the Send Queue is not full.
+            // So we just consider this error as an unrecoverable error.
+            LOG(WARNING) << "Fail to ibv_post_send: " << berror(err);
+            return -1;
+        }
+        cur_unsignaled = _m_sq_unsignaled.exchange(0, butil::memory_order_relaxed);
+        {
+            BAIDU_SCOPED_LOCK(_sq_update_mutex);
+            _sq_to_update.push(cur_unsignaled);
+            //LOG(INFO) << "wr id: " << wr.wr_id <<  " SendImm before: " << cur_unsignaled;
+        }
+
     }
+
     return 0;
 }
@@ -938,8 +996,11 @@ ssize_t RdmaEndpoint::HandleCompletion(ibv_wc& wc) {
     bool zerocopy = FLAGS_rdma_recv_zerocopy;
     switch (wc.opcode) {
     case IBV_WC_SEND: {  // send completion
-        // Do nothing
-        break;
+        {
+            BAIDU_SCOPED_LOCK(_sq_flag_mutex);
+           _sq_update_flag.push(true);
+       }
+       break;
     }
     case IBV_WC_RECV: {  // recv completion
         // Please note that only the first wc.byte_len bytes is valid
@@ -959,24 +1020,66 @@ ssize_t RdmaEndpoint::HandleCompletion(ibv_wc& wc) {
         }
         if (wc.imm_data > 0) {
             // Clear sbuf here because we ignore event wakeup for send completions
-            uint32_t acks = butil::NetToHost32(wc.imm_data);
-            uint32_t num = acks;
-            while (num > 0) {
-                _sbuf[_sq_sent++].clear();
-                if (_sq_sent == _sq_size - RESERVED_WR_NUM) {
-                    _sq_sent = 0;
-                }
-                --num;
-            }
-            butil::subtle::MemoryBarrier();

             // Update window
-            uint32_t wnd_thresh = _local_window_capacity / 8;
-            if (_window_size.fetch_add(acks, butil::memory_order_relaxed) >= wnd_thresh
-                    || acks >= wnd_thresh) {
+            //uint32_t wnd_thresh = _local_window_capacity / 8;
+            //if (_window_size.fetch_add(acks, butil::memory_order_relaxed) >= wnd_thresh
+            //        || acks >= wnd_thresh) {
                 // Do not wake up writing thread right after _window_size > 0.
                 // Otherwise the writing thread may switch to background too quickly.
-                _socket->WakeAsEpollOut();
+            //    _socket->WakeAsEpollOut();
+            //}
+           uint32_t acks = butil::NetToHost32(wc.imm_data);
+           //LOG(INFO) << "acks: " << acks;
+            _remote_acks.fetch_add(acks, butil::memory_order_relaxed);
+            {
+                BAIDU_SCOPED_LOCK(_sq_update_mutex);
+                while(_sq_to_update.empty() == false &&  _remote_acks.load() >= _sq_to_update.front()) {
+                    {
+                        BAIDU_SCOPED_LOCK(_sq_flag_mutex);
+                        if(_sq_update_flag.empty() == true) {
+                            break;
+                        }
+                        _sq_update_flag.pop();
+                    }
+                    uint32_t wnd_to_update = _sq_to_update.front();
+                    _sq_to_update.pop();
+                    _remote_acks.fetch_sub(wnd_to_update, butil::memory_order_relaxed);
+
+                    //LOG(INFO)  << wnd_to_update << " " <<   _remote_acks << " " << _sq_update_flag.size() << " " << _sq_to_update.size();
+                   uint32_t num = wnd_to_update;
+                    while(num > 0) {
+                        _sbuf[_sq_sent++].clear();
+                        if (_sq_sent == _sq_size - RESERVED_WR_NUM) {
+                            _sq_sent = 0;
+                        }
+                        --num;
+                    }
+                    butil::subtle::MemoryBarrier();
+                    uint32_t wnd_thresh = _local_window_capacity / 8;
+                    if (_window_size.fetch_add(wnd_to_update, butil::memory_order_relaxed) >= wnd_thresh
+                            || acks >= wnd_thresh) {
+            // Do not wake up writing thread right after _window_size > 0.
+            // Otherwise the writing thread may switch to background too quickly.
+                                _socket->WakeAsEpollOut();
+                    }
+                }
             }
         }
diff --git a/src/brpc/rdma/rdma_endpoint.h b/src/brpc/rdma/rdma_endpoint.h
index de7cd5f6..114eb682 100644
--- a/src/brpc/rdma/rdma_endpoint.h
+++ b/src/brpc/rdma/rdma_endpoint.h
@@ -31,7 +31,7 @@
 #include "butil/containers/mpsc_queue.h"
 #include "brpc/socket.h"

-
+#include <queue>
 namespace brpc {
 class Socket;
 namespace rdma {
@@ -262,6 +262,13 @@ private:
     // The number of new WRs posted in the local Recv Queue
     butil::atomic<uint16_t> _new_rq_wrs;

+    butil::atomic<int> _remote_acks;
+    butil::Mutex _m_sq_mutex;
+    butil::Mutex _sq_update_mutex;
+    butil::Mutex _sq_flag_mutex;
+    butil::atomic<uint16_t> _m_sq_unsignaled;
+    std::queue<uint16_t> _sq_to_update;
+    std::queue<bool> _sq_update_flag;
     // butex for inform read events on TCP fd during handshake
     butil::atomic<int> *_read_butex;
```

**
**Versions**
OS:
Compiler:
brpc:
protobuf:

chenBright:

Issue Number: resolve #3132

Problem Summary:

Changed:

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/docs/cn/rdma.md
+++ b/docs/cn/rdma.md
@@ -35,7 +35,7 @@ RDMA与TCP不同，不使用socket接口进行通信。但是在实现上仍然�
 
 brpc内部使用RDMA RC模式，每个RdmaEndpoint对应一个QP。RDMA连接建立依赖于前置TCP建连，TCP建连后双方交换必要参数，如GID、QPN等，再发起RDMA连接并实现数据传输。这个过程我们称为握手（参见RdmaEndpoint）。因为握手需要TCP连接，因此RdmaEndpoint所在的Socket类中，原本的TCP fd仍然有效。握手过程采用了brpc中已有的AppConnect逻辑。注意，握手用的TCP连接在后续数据传输阶段并不会收发数据，但仍保持为EST状态。一旦TCP连接中断，其上对应的RDMA连接同样会置错。
 
-RdmaEndpoint数据传输逻辑的第一个重要特性是零拷贝。要发送的所有数据默认都存放在IOBuf的Block中，因此所发送的Block需要等到对端确认接收完成后才可以释放，这些Block的引用被存放于RdmaEndpoint::_sbuf中。而要实现接收零拷贝，则需要确保接受端所预提交的接收缓冲区必须直接在IOBuf的Block里面，被存放于RdmaEndpoint::_rbuf。注意，接收端预提交的每一段Block，有一个固定的大小（recv_block_size）。发送端发送时，一个请求最多只能有这么大，否则接收端则无法成功接收。
+RdmaEndpoint数据传输逻辑的第一个重要特性是零拷贝。要发送的所有数据默认都存放在IOBuf的Block中，因此所发送的Block需要等到发送CQE触发后才可以释放，这些Block的引用被存放于RdmaEndpoint::_sbuf中。而要实现接收零拷贝，则需要确保接受端所预提交的接收缓冲区必须直接在IOBuf的Block里面，被存放于RdmaEndpoint::_rbuf。注意，接收端预提交的每一段Block，有一个固定的大小（recv_block_size）。发送端发送时，一个请求最多只能有这么大，否则接收端则无法成功接收。
 
 RdmaEndpoint数据传输逻辑的第二个重要特性是滑动窗口流控。这一流控机制是为了避免发送端持续在发送，其速度超过了接收端处理的速度。TCP传输中也有类似的逻辑，但是是由内核协议栈来实现的。RdmaEndpoint内实现了这一流控机制，通过接收端显式回复ACK来确认接收端处理完毕。为了减少ACK本身的开销，让ACK以立即数形式返回，可以被附在数据消息里。
 
@@ -52,26 +52,26 @@ RDMA支持事件驱动和轮询两种模式，默认是事件驱动模式，通�
 # 参数
 
 可配置参数说明：
-* rdma_trace_verbose: 日志中打印RDMA建连相关信息，默认false
-* rdma_recv_zerocopy: 是否启用接收零拷贝，默认true
-* rdma_zerocopy_min_size: 接收零拷贝最小的msg大小，默认512B
-* rdma_recv_block_type: 为接收数据预准备的block类型，分为三类default(8KB)/large(64KB)/huge(2MB)，默认为default
-* rdma_prepared_qp_size: 程序启动预生成的QP的大小，默认128
-* rdma_prepared_qp_cnt: 程序启动预生成的QP的数量，默认1024
-* rdma_max_sge: 允许的最大发送SGList长度，默认为0，即采用硬件所支持的最大长度
-* rdma_sq_size: SQ大小，默认128
-* rdma_rq_size: RQ大小，默认128
-* rdma_cqe_poll_once: 从CQ中一次性poll出的CQE数量，默认32
-* rdma_gid_index: 使用本地GID表中的Index，默认为-1，即选用最大的可用GID Index
-* rdma_port: 使用IB设备的port number，默认为1
-* rdma_device: 使用IB设备的名称，默认为空，即使用第一个active的设备
-* rdma_memory_pool_initial_size_mb: 内存池的初始大小，单位MB，默认1024
-* rdma_memory_pool_increase_size_mb: 内存池每次动态增长的大小，单位MB，默认1024
-* rdma_memory_pool_max_regions: 最大的内存池块数，默认16
-* rdma_memory_pool_buckets: 内存池中为避免竞争采用的bucket数目，默认为4
-* rdma_memory_pool_tls_cache_num: 内存池中thread local的缓存block数目，默认为128
-* rdma_use_polling: 是否使用RDMA的轮询模式，默认false
-* rdma_poller_num: 轮询模式下的poller数目，默认1
-* rdma_poller_yield: 轮询模式下的poller是否主动放弃CPU，默认是false
-* rdma_edisp_unsched: 让事件驱动器不可以被调度，默认是false
-* rdma_disable_bthread: 禁用bthread，默认是false
+* rdma_trace_verbose: 日志中打印RDMA建连相关信息，默认false。
+* rdma_recv_zerocopy: 是否启用接收零拷贝，默认true。
+* rdma_zerocopy_min_size: 接收零拷贝最小的msg大小，默认512B。
+* rdma_recv_block_type: 为接收数据预准备的block类型，分为三类default(8KB)/large(64KB)/huge(2MB)，默认为default。
+* rdma_prepared_qp_size: 程序启动预生成的QP的大小，默认128。
+* rdma_prepared_qp_cnt: 程序启动预生成的QP的数量，默认1024。
+* rdma_max_sge: 允许的最大发送SGList长度，默认为0，即采用硬件所支持的最大长度。
+* rdma_sq_size: SQ大小，默认128。
+* rdma_rq_size: RQ大小，默认128。
+* rdma_cqe_poll_once: 从CQ中一次性poll出的CQE数量，默认32。
+* rdma_gid_index: 使用本地GID表中的Index，默认为-1，即选用最大的可用GID Index。
+* rdma_port: 使用IB设备的port number，默认为1。
+* rdma_device: 使用IB设备的名称，默认为空，即使用第一个active的设备。
+* rdma_memory_pool_initial_size_mb: 内存池的初始大小，单位MB，默认1024。
+* rdma_memory_pool_increase_size_mb: 内存池每次动态增长的大小，单位MB，默认1024。
+* rdma_memory_pool_max_regions: 最大的内存池块数，默认3。
+* rdma_memory_pool_buckets: 内存池中为避免竞争采用的bucket数目，默认为4。
+* rdma_memory_pool_tls_cache_num: 内存池中thread local的缓存block数目，默认为128。
+* rdma_use_polling: 是否使用RDMA的轮询模式，默认false。
+* rdma_poller_num: 轮询模式下的poller数目，默认1。
+* rdma_poller_yield: 轮询模式下的poller是否主动放弃CPU，默认是false。
+* rdma_edisp_unsched: 让事件驱动器不可以被调度，默认是false。
+* rdma_disable_bthread: 禁用bthread，默认是false。
--- a/docs/en/rdma.md
+++ b/docs/en/rdma.md
@@ -35,7 +35,7 @@ RDMA does not use socket API like TCP. However, the brpc::Socket class is still 
 
 brpc uses RDMA RC mode. Every RdmaEndpoint has its own QP. Before establishing RDMA connection, a TCP connection is necessary to exchange some information such as GID and QPN. We call this procedure handshake. Since handshake needs TCP connection, the TCP fd in the corresponding Socket is still valid. The handshake procedure is completed in the AppConnect way in brpc. The TCP connection will keep in EST state but not be used for data transmission after RDMA connection is established. Once the TCP connection is closed, the corresponding RDMA connection will be set error.
 
-The first key feature in RdmaEndpoint data transmission is zero copy. All data which need to transmit is in the Blocks of IOBuf. Thus all the Blocks need to be released after the remote side completes the receiving. The reference of these Blocks are stored in RdmaEndpoint::_sbuf. In order to realize receiving zero copy, the receive side must post receive buffers in Blocks of IOBuf, which are stored in RdmaEndpoint::_rbuf. Note that all the Blocks posted in the receive side has a fixed size (recv_block_size). The transmit side can only send message smaller than that. Otherwise the receive side cannot receive data successfully.
+The first key feature in RdmaEndpoint data transmission is zero copy. All data which need to transmit is in the Blocks of IOBuf. Thus all the Blocks need to be released after the sent CQEs are triggered. The reference of these Blocks are stored in RdmaEndpoint::_sbuf. In order to realize receiving zero copy, the receive side must post receive buffers in Blocks of IOBuf, which are stored in RdmaEndpoint::_rbuf. Note that all the Blocks posted in the receive side has a fixed size (recv_block_size). The transmit side can only send message smaller than that. Otherwise the receive side cannot receive data successfully.
 
 The second key feature in RdmaEndpoint data transmission is sliding window flow control. The flow control is to avoid fast transmit side overwhelming slow receive side. TCP has similar mechanism in kernel TCP stack. RdmaEndpoint implements this mechanism with explicit ACKs from receive side. to reduce the overhead of ACKs, the ACK number can be piggybacked in ordinary data message as immediate data.
 
@@ -50,21 +50,26 @@ RDMA is hardware-related. It has some different concepts such as device, port, G
 # Parameters
 
 Configurable parameters:
-* rdma_trace_verbose: to print RDMA connection information in log，default is false
-* rdma_recv_zerocopy: enable zero copy in receive side，default is true
-* rdma_zerocopy_min_size: the min message size for receive zero copy (in Byte)，default is 512
-* rdma_recv_block_type: the block type used for receiving, can be default(8KB)/large(64KB)/huge(2MB)，default is default
-* rdma_prepared_qp_size: the size of QPs created at the beginning of the application，default is 128
-* rdma_prepared_qp_cnt: the number of QPs created at the beginning of the application，default is 1024
-* rdma_max_sge: the max length of sglist, default is 0, which is the max length allowed by the device
-* rdma_sq_size: the size of SQ，default is 128
-* rdma_rq_size: the size of RQ，default is 128
-* rdma_cqe_poll_once: the number of CQE pooled from CQ once，default is 32
-* rdma_gid_index: the index of local GID table used，default is -1，which is the maximum GID index
-* rdma_port: the port number used，default is 1
-* rdma_device: the IB device name，default is empty，which is the first active device
-* rdma_memory_pool_initial_size_mb: the initial region size of RDMA memory pool (in MB)，default is 1024
-* rdma_memory_pool_increase_size_mb: the step increase region size of RDMA memory pool (in MB)，default is 1024
-* rdma_memory_pool_max_regions: the max number of regions in RDMA memory pool，default is 16
-* rdma_memory_pool_buckets: the number of buckets for avoiding mutex contention in RDMA memory pool，default is 4
-* rdma_memory_pool_tls_cache_num: the number of thread local cached blocks in RDMA memory pool，default is 128
+* rdma_trace_verbose: to print RDMA connection information in log，default is false.
+* rdma_recv_zerocopy: enable zero copy in receive side，default is true.
+* rdma_zerocopy_min_size: the min message size for receive zero copy (in Byte)，default is 512.
+* rdma_recv_block_type: the block type used for receiving, can be default(8KB)/large(64KB)/huge(2MB)，default is default.
+* rdma_prepared_qp_size: the size of QPs created at the beginning of the application，default is 128.
+* rdma_prepared_qp_cnt: the number of QPs created at the beginning of the application，default is 1024.
+* rdma_max_sge: the max length of sglist, default is 0, which is the max length allowed by the device.
+* rdma_sq_size: the size of SQ，default is 128.
+* rdma_rq_size: the size of RQ，default is 128.
+* rdma_cqe_poll_once: the number of CQE pooled from CQ once，default is 32.
+* rdma_gid_index: the index of local GID table used，default is -1，which is the maximum GID index.
+* rdma_port: the port number used，default is 1.
+* rdma_device: the IB device name，default is empty，which is the first active device.
+* rdma_memory_pool_initial_size_mb: the initial region size of RDMA memory pool (in MB)，default is 1024.
+* rdma_memory_pool_increase_size_mb: the step increase region size of RDMA memory pool (in MB)，default is 1024.
+* rdma_memory_pool_max_regions: the max number of regions in RDMA memory pool，default is 3.
+* rdma_memory_pool_buckets: the number of buckets for avoiding mutex contention in RDMA memory pool，default is 4.
+* rdma_memory_pool_tls_cache_num: the number of thread local cached blocks in RDMA memory pool，default is 128.
+* rdma_use_polling: Whether to use RDMA polling mode, default is false.
+* rdma_poller_num: The number of pollers in polling mode, default is 1.
+* rdma_poller_yield: Whether pollers in polling mode voluntarily relinquish the CPU, default is false.
+* rdma_edisp_unsched`: Prevents the event driver from being scheduled, default is false.
+* rdma_disable_bthread: Disables bthread, default is false.


chenBright:
We encountered the same problem as #3132. This PR's solution is similar to https://github.com/apache/brpc/issues/3132#issuecomment-3466917187 and can solve this problem. PTAL @sunce4t @yanglimingcn

src/brpc/rdma/rdma_endpoint.cpp
```diff
@@ -513,7 +520,10 @@ void* RdmaEndpoint::ProcessHandshakeAtClient(void* arg) {
             std::min(ep->_sq_size, remote_msg.rq_size) - RESERVED_WR_NUM;
         ep->_remote_window_capacity = 
             std::min(ep->_rq_size, remote_msg.sq_size) - RESERVED_WR_NUM,
-        ep->_window_size.store(ep->_local_window_capacity, butil::memory_order_relaxed);
+        ep->_remote_rq_window_size.store(
```
yanglimingcn, line 525: why _remote_rq_window_size not equal to _remote_window_capacity

src/brpc/rdma/rdma_endpoint.cpp
```diff
@@ -894,11 +913,14 @@ ssize_t RdmaEndpoint::CutFromIOBufList(butil::IOBuf** from, size_t ndata) {
             _sq_current = 0;
         }
 
-        // Update _window_size. Note that _window_size will never be negative.
+        // Update `_remote_rq_window_size' and `_sq_window_size'. Note that
```
yanglimingcn, line 917: The `remote_rq_window_size` should only be updated when `solicited` is true.

src/brpc/rdma/rdma_endpoint.cpp
```diff
@@ -972,9 +1001,9 @@ ssize_t RdmaEndpoint::HandleCompletion(ibv_wc& wc) {
 
             // Update window
             uint32_t wnd_thresh = _local_window_capacity / 8;
-            if (_window_size.fetch_add(acks, butil::memory_order_relaxed) >= wnd_thresh
+            if (_remote_rq_window_size.fetch_add(acks, butil::memory_order_relaxed) >= wnd_thresh
```
yanglimingcn: This part receives an update on the rq queue size from Recv, so it would be more appropriate to call this variable _rq_window_size, which corresponds to _sq_window_size mentioned above.

src/brpc/rdma/rdma_endpoint.cpp
chenBright: ibv_post_send is limited by the local sq and the remote rq, but _remote_window_capacity is the minimum value of the local rq and the remote sq, so the maximum value of _remote_rq_window_size should not be equal to _remote_window_capacity.

src/brpc/rdma/rdma_endpoint.cpp
```diff
@@ -938,8 +961,16 @@ ssize_t RdmaEndpoint::HandleCompletion(ibv_wc& wc) {
...
-        break;
+        if (SEND_TYPE_RDMA_IMM == wc.wr_id) {
+            // Do nothing for imm.
+            return 0;
+        }
+        // Update window
+        uint16_t wnd_to_update = _local_window_capacity / 4;
+        _sq_window_size.fetch_add(wnd_to_update, butil::memory_order_relaxed);
+        // Wake up writing thread right after every signaled sending cqe.
+        _socket->WakeAsEpollOut();
```
yanglimingcn: `_socket->WakeAsEpollOut` is called in both `IBV_WC_SEND` and `IBV_WC_RECV`, but each branch only has one condition. Shouldn't it activate only when both conditions are met?

src/brpc/rdma/rdma_endpoint.cpp
```diff
@@ -972,9 +1001,9 @@ ssize_t RdmaEndpoint::HandleCompletion(ibv_wc& wc) {
 
             // Update window
             uint32_t wnd_thresh = _local_window_capacity / 8;
-            if (_window_size.fetch_add(acks, butil::memory_order_relaxed) >= wnd_thresh
+            if (_remote_rq_window_size.fetch_add(acks, butil::memory_order_relaxed) >= wnd_thresh
                     || acks >= wnd_thresh) {
-                // Do not wake up writing thread right after _window_size > 0.
+                // Do not wake up writing thread right after _remote_rq_window_size > 0.
```
yanglimingcn, line 1019: comment update "> wnd_thresh"

src/brpc/rdma/rdma_endpoint.cpp
```diff
@@ -938,8 +961,16 @@ ssize_t RdmaEndpoint::HandleCompletion(ibv_wc& wc) {
...
     switch (wc.opcode) {
     case IBV_WC_SEND: {  // send completion
-        // Do nothing
-        break;
+        if (SEND_TYPE_RDMA_IMM == wc.wr_id) {
+            // Do nothing for imm.
+            return 0;
+        }
+        // Update window
+        uint16_t wnd_to_update = _local_window_capacity / 4;
```
yanglimingcn: With this modification, I think we no longer need `attr.cap.max_send_wr = sq_size * 5 / 4;`，This part is rather difficult to understand.

sunce4t: I think this is correct—there really is no need anymore.

src/brpc/rdma/rdma_endpoint.cpp
chenBright: An additional condition can be added: `_remote_rq_window_size > _local_window_capacity / 8`.

src/brpc/rdma/rdma_endpoint.cpp
chenBright: I agree.

chenBright: On the other hand, I don't understand why the number of reserved WRs in SQ/RQ for pure ACK is 3.

https://github.com/apache/brpc/blob/ede6b7763ab52829bd1436c4f0bab4a708800c3a/src/brpc/rdma/rdma_endpoint.cpp#L70-L72

src/brpc/rdma/rdma_endpoint.cpp
chenBright: I think the comments are fine. "_remote_rq_window_size > 0, do not wake up..."

src/brpc/rdma/rdma_endpoint.cpp
sunce4t: I think this may be reserved for SendImm?

yanglimingcn: I'm not sure why it's 3, but keeping this reserved space seems fine.

yanglimingcn:
Why is CQ divided into send_cq and recv_cq?

chenBright:
We encountered an error: requests were not being sent, causing a large number of client timeouts.

```shell
 [E1008]Reached timeout=60000ms @Socket{id=13 fd=1160 addr=xxx:xx} (0x0x7f957c964ec0) rdma info={rdma_state=ON, handshake_state=ESTABLISHED, rdma_remote_rq_window_size=63, rdma_sq_window_size=0, rdma_local_window_capacity=125, rdma_remote_window_capacity=125, rdma_sbuf_head=57, rdma_sbuf_tail=120, rdma_rbuf_head=36, rdma_unacked_rq_wr=0, rdma_received_ack=0, rdma_unsolicited_sent=0, rdma_unsignaled_sq_wr=1, rdma_new_rq_wrs=0, }
```

From the RDMA connection information, we found that because `ibv_req_notify_cq` was only solicited, send WCs did not generate a CQEs. Without  recv CQEs, send WCs could not be polled, so ·_sq_window_size` remained at 0. This is likely the reason why both the client and server are unable to send messages.

Using `ibv_req_notify_cq` with `solicited_only=0` could solve this problem, but it would generate too many events. Therefore, we split the CQ into `send_cq`(`solicited_only=0`) and `recv_cq`(`solicited_only=1`).

src/brpc/rdma/rdma_endpoint.cpp
chenBright: done

yanglimingcn:
> We encountered an error: requests were not being sent, causing a large number of client timeouts.
> 
> ```shell
>  [E1008]Reached timeout=60000ms @Socket{id=13 fd=1160 addr=xxx:xx} (0x0x7f957c964ec0) rdma info={rdma_state=ON, handshake_state=ESTABLISHED, rdma_remote_rq_window_size=63, rdma_sq_window_size=0, rdma_local_window_capacity=125, rdma_remote_window_capacity=125, rdma_sbuf_head=57, rdma_sbuf_tail=120, rdma_rbuf_head=36, rdma_unacked_rq_wr=0, rdma_received_ack=0, rdma_unsolicited_sent=0, rdma_unsignaled_sq_wr=1, rdma_new_rq_wrs=0, }
> ```
> 
> From the RDMA connection information, we found that because `ibv_req_notify_cq` was only solicited, send WCs did not generate a CQEs. Without recv CQEs, send WCs could not be polled, so ·_sq_window_size` remained at 0. This is likely the reason why both the client and server are unable to send messages.
> 
> Using `ibv_req_notify_cq` with `solicited_only=0` could solve this problem, but it would generate too many events. Therefore, we split the CQ into `send_cq`(`solicited_only=0`) and `recv_cq`(`solicited_only=1`).

With this modification, send_cq will generate one CQE for every 1/4 of the window？

src/brpc/rdma/rdma_endpoint.cpp
```diff
@@ -1373,22 +1443,39 @@ void RdmaEndpoint::PollCq(Socket* m) {
     }
     CHECK(ep == s->_rdma_ep);
 
+    bool send = false;
+    ibv_comp_channel* comp_channel = NULL;
+    ibv_cq* cq = NULL;
+    int solicited_only = 0;
```
yanglimingcn: Will CQEs be generated continuously in Polling mode?

chenBright:
> With this modification, send_cq will generate one CQE for every 1/4 of the window？

Yes.

src/brpc/rdma/rdma_endpoint.cpp
chenBright: No events are generated in polling mode. I think that this modification will not affect the logic of polling mode.

legionxiong:
> We encountered an error: requests were not being sent, causing a large number of client timeouts.
> 
> ```shell
>  [E1008]Reached timeout=60000ms @Socket{id=13 fd=1160 addr=xxx:xx} (0x0x7f957c964ec0) rdma info={rdma_state=ON, handshake_state=ESTABLISHED, rdma_remote_rq_window_size=63, rdma_sq_window_size=0, rdma_local_window_capacity=125, rdma_remote_window_capacity=125, rdma_sbuf_head=57, rdma_sbuf_tail=120, rdma_rbuf_head=36, rdma_unacked_rq_wr=0, rdma_received_ack=0, rdma_unsolicited_sent=0, rdma_unsignaled_sq_wr=1, rdma_new_rq_wrs=0, }
> ```
> 
> From the RDMA connection information, we found that because `ibv_req_notify_cq` was only solicited, send WCs did not generate a CQEs. Without recv CQEs, send WCs could not be polled, so ·_sq_window_size` remained at 0. This is likely the reason why both the client and server are unable to send messages.
> 
> Using `ibv_req_notify_cq` with `solicited_only=0` could solve this problem, but it would generate too many events. Therefore, we split the CQ into `send_cq`(`solicited_only=0`) and `recv_cq`(`solicited_only=1`).

It is unnecessary to split CQ into send_cq and recv_cq, the reason that sliding window goes wrong is the precondition it relies on is not guaranteed in RoCE environment.  The sliding window mechanism presumes that an local app receives the ack from remote app means that the underlying SQ has already been released,  in a lossless IB environment it maybe true. But in RoCE environment, it is not guaranteed. Because the local device releases SQ only if it receives the ack to a message from remote device, but the ack might be lost and the remote side has processed the message, which means the remote app has received the message and sent an application layer ack to local app. Unfortunately, the local app layer received the ack to a message from remote app and then increased the window, but the device layer is still waiting for the ack from remote device to release the SQ.  So the story is the window being increased wrongly for one or two times, till the window is absolutely wrong.

chenBright:
> Unfortunately, the local app layer received the ack to a message from remote app and then increased the window, but the device layer is still waiting for the ack from remote device to release the SQ.

@legionxiong  f the device-level ACK is lost, is it impossible to release the SQ?

> It is unnecessary to split CQ into send_cq and recv_cq, the reason that sliding window goes wrong is the precondition it relies on is not guaranteed in RoCE environment. 

Is there a better solution?

legionxiong:
> the device-level ACK is lost

  Local side will resend the message whose ACK was not received within `rdma_ack_timeout`,  and  remote side will do nothing but resend the ACK to the message that has been received before. Maybe we need another variable which checks the event of IBV_WC_SEND,  it is guaranteed that SQs are released if we polled the work completion of IBV_WC_SEND. And the number of unsignaled SQs could be carried with wr_id.

chenBright:
> Maybe we need another variable which checks the event of IBV_WC_SEND, it is guaranteed that SQs are released if we polled the work completion of IBV_WC_SEND.

>  [E1008]Reached timeout=60000ms @Socket{id=13 fd=1160 addr=xxx:xx} (0x0x7f957c964ec0) rdma info={rdma_state=ON, handshake_state=ESTABLISHED, rdma_remote_rq_window_size=63, rdma_sq_window_size=0, rdma_local_window_capacity=125, rdma_remote_window_capacity=125, rdma_sbuf_head=57, rdma_sbuf_tail=120, rdma_rbuf_head=36, rdma_unacked_rq_wr=0, rdma_received_ack=0, rdma_unsolicited_sent=0, rdma_unsignaled_sq_wr=1, rdma_new_rq_wrs=0, }

In this case, it seems we cannot poll any work completion of  IBV_WC_SEND for a long time. With solicited_only=1, if device layer ACKs follows all application layer ACKs, we may be unable to poll IBV_WC_SEND after polling IBV_WC_RECV because there are no CQEs.

> Maybe we need another variable which checks the event of IBV_WC_SEND

How can we check for the event of IBV_WC_SEND？

> the number of unsignaled SQs could be carried with wr_id.

This solution for updating the SQ window is better.

randomkang:
@chenBright Hi, i use the patch(commit:db3673acf276d9baec8aff95afd07bdbc5811437) in my task, but i still get the error:
"[wk-10] W1124 07:27:34.681490 82718     0 external/brpc/src/brpc/rdma/rdma_endpoint.cpp:952 CutFromIOBufList] Fail to ibv_post_send: Cannot allocate memory, remote_rq_window_size=57, sq_window_size=5, sq_current=5".

My task is model trainning and the brpc gdr is open.

chenBright:
> @chenBright Hi, i use the patch(commit:db3673acf276d9baec8aff95afd07bdbc5811437) in my task, but i still get the error: "[wk-10] W1124 07:27:34.681490 82718 0 external/brpc/src/brpc/rdma/rdma_endpoint.cpp:952 CutFromIOBufList] Fail to ibv_post_send: Cannot allocate memory, remote_rq_window_size=57, sq_window_size=5, sq_current=5".
> 
> My task is model trainning and the brpc gdr is open.

Could it be that some unexpected imms are occupying SQ? Try to set sq_size back to its original value of sq_size * 5 / 4.

```c++
resource->qp = AllocateQp(resource->send_cq, resource->recv_cq, sq_size * 5 / 4, rq_size);
```

yanglimingcn:
Does this mean that the 5/4 window still can't be removed with the current modifications?

randomkang:
> > @chenBright Hi, i use the patch(commit:db3673acf276d9baec8aff95afd07bdbc5811437) in my task, but i still get the error: "[wk-10] W1124 07:27:34.681490 82718 0 external/brpc/src/brpc/rdma/rdma_endpoint.cpp:952 CutFromIOBufList] Fail to ibv_post_send: Cannot allocate memory, remote_rq_window_size=57, sq_window_size=5, sq_current=5".
> > My task is model trainning and the brpc gdr is open.
> 
> Could it be that some unexpected imms are occupying SQ? Try to set sq_size back to its original value of sq_size * 5 / 4.
> 
> ```c++
> resource->qp = AllocateQp(resource->send_cq, resource->recv_cq, sq_size * 5 / 4, rq_size);
> ```

I set  sq_size back to its original value of sq_size * 5 / 4 and run the same task two times. one last 1383 minutes and the other last 752 minutes. The error of "Fail to ibv_post_send: Cannot allocate memory"  do not happen. @chenBright @yanglimingcn

randomkang:
> > > @chenBright Hi, i use the patch(commit:db3673acf276d9baec8aff95afd07bdbc5811437) in my task, but i still get the error: "[wk-10] W1124 07:27:34.681490 82718 0 external/brpc/src/brpc/rdma/rdma_endpoint.cpp:952 CutFromIOBufList] Fail to ibv_post_send: Cannot allocate memory, remote_rq_window_size=57, sq_window_size=5, sq_current=5".
> > > My task is model trainning and the brpc gdr is open.
> > 
> > 
> > Could it be that some unexpected imms are occupying SQ? Try to set sq_size back to its original value of sq_size * 5 / 4.
> > ```c++
> > resource->qp = AllocateQp(resource->send_cq, resource->recv_cq, sq_size * 5 / 4, rq_size);
> > ```
> 
> I set sq_size back to its original value of sq_size * 5 / 4 and run the same task two times. one last 1383 minutes and the other last 752 minutes. The error of "Fail to ibv_post_send: Cannot allocate memory" do not happen. @chenBright @yanglimingcn

Update: one task run 1063 minutes, and report an error: "[wk-7] W1127 16:13:16.662405 67606 158119221015812 external/brpc/src/brpc/rdma/rdma_endpoint.cpp:997 SendImm] Fail to ibv_post_send: Cannot allocate memory".

randomkang:
> > > > @chenBright Hi, i use the patch(commit:db3673acf276d9baec8aff95afd07bdbc5811437) in my task, but i still get the error: "[wk-10] W1124 07:27:34.681490 82718 0 external/brpc/src/brpc/rdma/rdma_endpoint.cpp:952 CutFromIOBufList] Fail to ibv_post_send: Cannot allocate memory, remote_rq_window_size=57, sq_window_size=5, sq_current=5".
> > > > My task is model trainning and the brpc gdr is open.
> > > 
> > > 
> > > Could it be that some unexpected imms are occupying SQ? Try to set sq_size back to its original value of sq_size * 5 / 4.
> > > ```c++
> > > resource->qp = AllocateQp(resource->send_cq, resource->recv_cq, sq_size * 5 / 4, rq_size);
> > > ```
> > 
> > 
> > I set sq_size back to its original value of sq_size * 5 / 4 and run the same task two times. one last 1383 minutes and the other last 752 minutes. The error of "Fail to ibv_post_send: Cannot allocate memory" do not happen. @chenBright @yanglimingcn
> 
> Update: one task run 1063 minutes, and report an error: "[wk-7] W1127 16:13:16.662405 67606 158119221015812 external/brpc/src/brpc/rdma/rdma_endpoint.cpp:997 SendImm] Fail to ibv_post_send: Cannot allocate memory".

Update2: The other task run 1902 minutes, and also report an error "[wk-13] W1127 19:41:55.088111 98359 102791452311833 external/brpc/src/brpc/rdma/rdma_endpoint.cpp:997 SendImm] Fail to ibv_post_send: Cannot allocate memory".

chenBright:
Latest changes:
1. Send CQ and recv CQ share a comp channel. @yanglimingcn 
2. When ibv_post_send fails, it prints the number of inflight imm for debugging. @randomkang

yanglimingcn:
LGTM, @randomkang Please try the latest PR again; we haven't encountered the problem you mentioned.

randomkang:
> LGTM, @randomkang Please try the latest PR again; we haven't encountered the problem you mentioned.

The brpc version which i used is 1.14.1, does it matter?

yanglimingcn:
I don't think it matters; just focus on the modifications to the RDMA section.

randomkang:
I run three tasks, they lasts for 1742、1551、1852 minutes, and the error of "Fail to ibv_post_send: Cannot allocate memory" does not happpen again. @yanglimingcn @chenBright

yanglimingcn:
LGTM

chenBright:
> I run three tasks, they lasts for 1742、1551、1852 minutes, and the error of "Fail to ibv_post_send: Cannot allocate memory" does not happpen again. @yanglimingcn @chenBright

@randomkang Is this using the latest commit?

randomkang:
yes, i use the commit of "0794a476ee88e19f91cb9ce43166b4d2f90077b3"

yanglimingcn:
In fact, I still don't understand the specific function of the 5/4 window in the latest Premiere Pro.

chenBright:
> In fact, I still don't understand the specific function of the 5/4 window in the latest Premiere Pro.

I think it would be more reasonable to add a sliding window to limit the number of IMM.

@randomkang Please try the latest PR again

yanglimingcn:
So how should this window be set up reasonably? It seems the number of imm items generated is not quite certain?

chenBright:
> So how should this window be set up reasonably? It seems the number of imm items generated is not quite certain?

The window size is related to SendAck.

When _remote_window_capacity is divisible by 2, a maximum of 2 imm are required.

When _remote_window_capacity is not divisible by 2, a maximum of 3 imm are required.

Therefore, RESERVED_WR_NUM is a reasonable value.

https://github.com/apache/brpc/blob/7cd0c25a720262267531dcbb1135158b25de09ca/src/brpc/rdma/rdma_endpoint.cpp#L518-L519

https://github.com/apache/brpc/blob/7cd0c25a720262267531dcbb1135158b25de09ca/src/brpc/rdma/rdma_endpoint.cpp#L911-L916

yanglimingcn:
Since RESERVED_WR_NUM is already present, is the 5/4 no longer needed?

chenBright:
> Since RESERVED_WR_NUM is already present, is the 5/4 no longer needed?

Yes. The capacity of RESERVED_WR_NUM is sufficient; by using the imm window to limit the number of imm elements, 5/4 is no longer needed.

randomkang:
> > In fact, I still don't understand the specific function of the 5/4 window in the latest Premiere Pro.
> 
> I think it would be more reasonable to add a sliding window to limit the number of IMM.
> 
> @randomkang Please try the latest PR again

I run 2 tasks, and the error of "Fail to ibv_post_send: Cannot allocate memory" does not happpen again. 

One last for 1790 mins, and fail due to "[wk-15] Thread E1217 16:32:30.900705 106575 12820477405150 external/brpc/src/brpc/rdma/rdma_helper.cpp:220 BlockAllocate] Fail to get block from memory pool". This error is not related to communication.

The other last for 1787mins, and fail due to "[tf-ps-7] W1217 04:18:39.676858 24875 78292958891373 tensorflow/contrib/pms/rpc/brpc_remote_worker.cc:383 operator()] Fail to recv tensorflow.RecvTensorResponse from 10.28.132.10:3642 : Aborted: [E10010][10.28.132.10:3642][E10010]Step 60851325530873895". I don't see the error message of brpc, maybe this error is caused by tensorflow. @chenBright

randomkang:
> I run three tasks, they lasts for 1742、1551、1852 minutes, and the error of "Fail to ibv_post_send: Cannot allocate memory" does not happpen again. @yanglimingcn @chenBright

By the way, these three tasks also failed in final.

The first task failed due to "[wk-8] E1204 21:01:41.101131 72310 62642098032795 external/brpc/src/brpc/rdma/block_pool.cpp:362 AllocBlockFrom] Fail to extend new region. You can set the size of memory pool larger. Refer to the help message of these flags: rdma_memory_pool_initial_size_mb, rdma_memory_pool_increase_size_mb, rdma_memory_pool_max_regions." This error is not related to communication.

The second task failed, but i don't see any error message. Maybe it is killed by the matchine failure.

The third task failed due to "tf op is stuck".

chenBright:
@randomkang In summary, the error of "Fail to ibv_post_send: Cannot allocate memory" has been resolved. Right?

randomkang:
> @randomkang In summary, the error of "Fail to ibv_post_send: Cannot allocate memory" has been resolved. Right?

Yes, i think so.

chenBright:
@yanglimingcn Please review it again.

src/brpc/rdma/rdma_endpoint.cpp
```diff
@@ -924,26 +944,55 @@ int RdmaEndpoint::SendImm(uint32_t imm) {
...
-        // Do nothing
-        break;
+        if (0 == wc.wr_id) {
+            _sq_imm_window_size += 1;
+            // If there are any unacknowledged recvs, send an ack.
+            SendAck(0);
+            return 0;
+        }
+        // Update SQ window.
+        uint16_t wnd_to_update = wc.wr_id;
```
yanglimingcn, line 979: I saw in the documentation that sbuf cleanup should be done after the other end has finished receiving. How can this be guaranteed when this part is moved to IBV_WC_SEND?

chenBright: > I saw in the documentation that sbuf cleanup should be done after the other end has finished receiving. How can this be guaranteed when this part is moved to IBV_WC_SEND?

Previously, because there was only an application-layer sliding window, the sbuf could only be released by relying on the application-layer ACK.

Now, when the local device receives a send WC, it means that it has received an ACK from the remote device, the SQ has been released, the data has been sent to the remote device, so the sbuf can be safely released.

yanglimingcn: OK, The documentation for this section should also be updated.

chenBright: > OK, The documentation for this section should also be updated.

done

yanglimingcn:
LGTM

PR #3151, opened 2025-11-14
Fix bvar's bad return value

songqing:
Resolve issue: #3152 
Problem Summary:
Fix bug.
When read stat(such as proc, memory and load) failed, we need to return false, but return -1 now, which is true instead.

Changed:

Side effects:
- Performance effects:

---
### Check List:


wwbmmm:
LGTM

songqing:
@wwbmmm hi, could you help a review?

chenBright (approved):
LGTM

PR #3155, opened 2025-11-18
Support dynamic expansion of RDMA block pool

chenBright:

Issue Number: 

Problem Summary:

When `FLAGS_rdma_memory_pool_buckets > 1`, the block pool cannot be dynamically expanded.

Changed:

Dynamically expanded memory is stored in an `expansion_list` protected by `extend_lock`, and each bucket only retrieves memory from the `expansion_list` at the corresponding index.

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/brpc/rdma/block_pool.cpp
+++ b/src/brpc/rdma/block_pool.cpp
@@ -25,9 +25,8 @@
 #include "butil/iobuf.h"
 #include "butil/object_pool.h"
 #include "butil/thread_local.h"
-#include "bthread/bthread.h"
+#include "butil/memory/scope_guard.h"
 #include "brpc/rdma/block_pool.h"
-
 
 namespace brpc {
 namespace rdma {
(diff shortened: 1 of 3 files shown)


yanglimingcn:
When FLAGS_rdma_memory_pool_buckets is greater than 1, a bug may occur due to lock granularity issues.

chenBright:
> When FLAGS_rdma_memory_pool_buckets is greater than 1, a bug may occur due to lock granularity issues.

This PR can solve this problem.

yanglimingcn:
FLAGS_rdma_memory_pool_user_specified_memory is used to allow users to register specific memory, such as SPDK. In these scenarios, memory needs to be allocated by SPDK and then registered in RDMA, which can achieve zero copy.

chenBright:
> FLAGS_rdma_memory_pool_user_specified_memory is used to allow users to register specific memory, such as SPDK. In these scenarios, memory needs to be allocated by SPDK and then registered in RDMA, which can achieve zero copy.

Ok, I will revert it.

When the idle_list is out of memory in AllocBlockFrom function, it seems that the user cannot immediately dynamically expand specific memory. Is that right?

yanglimingcn:
In this scenario, the memory should be pre-registered, and dynamic allocation requires the user to provide memory too.

chenBright:
> > FLAGS_rdma_memory_pool_user_specified_memory is used to allow users to register specific memory, such as SPDK. In these scenarios, memory needs to be allocated by SPDK and then registered in RDMA, which can achieve zero copy.
> 
> Ok, I will revert it.

@yanglimingcn done

yanglimingcn:
LGTM

PR #3158, opened 2025-11-20
Support naming bthread to help debug

ZhengweiZhu:
The bthread name is shown when checking bthread status by curl ip:port/bthreads/xxx, which helps to debug when bthread trace is not enabled.

Issue Number:  enhancement for #3088 

Problem Summary:

Changed:
add `name` field in bthread_attr_t and set name when creating bthread

Side effects:
- Performance effects: NO

- Breaking backward compatibility: NO

---
### Check List:

### DIFF
--- a/src/bthread/task_group.cpp
+++ b/src/bthread/task_group.cpp
@@ -49,7 +49,7 @@
 namespace bthread {
 
 static const bthread_attr_t BTHREAD_ATTR_TASKGROUP = {
-    BTHREAD_STACKTYPE_UNKNOWN, 0, NULL, BTHREAD_TAG_INVALID };
+    BTHREAD_STACKTYPE_UNKNOWN, 0, NULL, BTHREAD_TAG_INVALID, {0} };
 
 DEFINE_bool(show_bthread_creation_in_vars, false, "When this flags is on, The time "
             "from bthread creation to first run will be recorded and shown in /vars");
@@ -1141,6 +1141,7 @@ void print_task(std::ostream& os, bthread_t tid, bool enable_trace,
            << "\nattr={stack_type=" << attr.stack_type
            << " flags=" << attr.flags
            << " specified_tag=" << attr.tag
+           << " name=" << attr.name
            << " keytable_pool=" << attr.keytable_pool
            << "}\nhas_tls=" << has_tls
            << "\nuptime_ns=" << butil::cpuwide_time_ns() - cpuwide_start_ns
--- a/src/bthread/types.h
+++ b/src/bthread/types.h
@@ -97,12 +97,14 @@ typedef struct {
     size_t nfree;
 } bthread_keytable_pool_stat_t;
 
+static const size_t BTHREAD_NAME_MAX_LENGTH = 31;
 // Attributes for thread creation.
 typedef struct bthread_attr_t {
     bthread_stacktype_t stack_type;
     bthread_attrflags_t flags;
     bthread_keytable_pool_t* keytable_pool;
     bthread_tag_t tag;
+    char name[BTHREAD_NAME_MAX_LENGTH + 1]; // do not use std::string to keep POD
 
 #if defined(__cplusplus)
     void operator=(unsigned stacktype_and_flags) {
@@ -120,6 +122,8 @@ typedef struct bthread_attr_t {
 #endif  // __cplusplus
 } bthread_attr_t;
 
+void bthread_attr_set_name(bthread_attr_t* attr, const char* name);
+
 // bthreads started with this attribute will run on stack of worker pthread and
 // all bthread functions that would block the bthread will block the pthread.
 // The bthread will not allocate its own stack, simply occupying a little meta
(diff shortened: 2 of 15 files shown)


src/bthread/types.h
```diff
@@ -118,6 +120,12 @@ typedef struct bthread_attr_t {
         return tmp;
     }
 #endif  // __cplusplus
+    void set_name(const char* val) {
```
wwbmmm: This is C++ style interface
but the original bthread interface is designed as C style.
maybe you can add a global function `bthread_attr_set_name`

src/bthread/types.h
```diff
@@ -97,12 +97,14 @@ typedef struct {
...
 } bthread_keytable_pool_stat_t;
 
+static const size_t BTHREAD_NAME_MAX_LENGTH = 31;
 // Attributes for thread creation.
 typedef struct bthread_attr_t {
     bthread_stacktype_t stack_type;
     bthread_attrflags_t flags;
     bthread_keytable_pool_t* keytable_pool;
     bthread_tag_t tag;
+    char name[BTHREAD_NAME_MAX_LENGTH + 1]; // do not use std::string to keep POD
```
wwbmmm, line 107: Some existing user code may create bthread_attr_t without initialize the `name`.
For backward compatibility, I think you can add a flag to the `bthread_attrflags_t flags`  like `BTHREAD_HAS_NAME`. Only when this flag is on, the name will be used.

src/bthread/types.h
```diff
@@ -118,6 +120,12 @@ typedef struct bthread_attr_t {
         return tmp;
     }
 #endif  // __cplusplus
+    void set_name(const char* val) {
+        if (val) {
+            strncpy(name, val, BTHREAD_NAME_MAX_LENGTH);
+            name[BTHREAD_NAME_MAX_LENGTH] = '\0';
```
chenBright: The null terminator ('\0') should be set according to the actual length of the string being copied.

src/bthread/task_group.cpp
```diff
@@ -1141,6 +1141,7 @@ void print_task(std::ostream& os, bthread_t tid, bool enable_trace,
            << "\nattr={stack_type=" << attr.stack_type
            << " flags=" << attr.flags
            << " specified_tag=" << attr.tag
+           << " name=" << attr.name
```
chenBright, line 1144: The name may be invalid.

src/bthread/types.h
ZhengweiZhu: For static constants like BTHREAD_ATTR_NORMAL, the name field is all initialized with {0}, which sets all elements of the array (including the last one) to zero. so user is not required to set name and the name will be printed as empty string, which is backward compatibility

src/bthread/task_group.cpp
ZhengweiZhu: what do u mean by invalid?  The empty string?

src/bthread/types.h
ZhengweiZhu: updated

src/bthread/types.h
ZhengweiZhu: For static constants like BTHREAD_ATTR_NORMAL, the name field is initialized with {0}, which sets all elements of the array (including the last one) to zero. This means the name is already fully null-initialized, and no further action is needed to ensure it is null-terminated.  This line name[BTHREAD_NAME_MAX_LENGTH] = '\0'; is actually not needed but kept here for program style.
The current implementation handles well whether the actual length being copied exceeds BTHREAD_NAME_MAX_LENGTH or not.

src/bthread/types.h
wwbmmm: OK, I see the `bthread_attr_init` function use `BTHREAD_ATTR_NORMAL`.

ZhengweiZhu: actually there's no usage of bthread_attr_init.  BTHREAD_ATTR_NORMAL is used when attr is null. 😂

src/bthread/types.h
chenBright: What if BTHREAD_ATTR_NORMAL is not used?

ZhengweiZhu: 1. Where bthread_attr_t is used, where the predefined static constants like 'BTHREAD_ATTR_*' is used. I have not seen any code constructing bthread_attr_t manually and no reason to do so.
2. Even if  'BTHREAD_ATTR_*' is not used, current implementation also handles well. when the bthread name is set by user, the string be copied should be null terminated. if the string exceeds BTHREAD_NAME_MAX_LENGTH, then name[BTHREAD_NAME_MAX_LENGTH] = '\0'; works to avoid buffer overflow.

chenBright: OK. My concern is that If 'BTHREAD_ATTR_*' is not used and the bthread name is not set, the bthread name is random char. Other fields also have this issue, but so far they're all working fine.

wwbmmm: Actually there is a `bthread_attr_init` function. It is the right way to init a `bthread_attr_t` without using pre-defined macros like 'BTHREAD_ATTR_*'. This is a more extensible way. 
In current way, some users may define `BTHREAD_ATTR_XXX` by themselves, but that code will be broken if we add new members to `bthread_attr_t`.

ZhengweiZhu: > Actually there is a `bthread_attr_init` function. It is the right way to init a `bthread_attr_t` without using pre-defined macros like 'BTHREAD_ATTR_*'. This is a more extensible way.
> In current way, some users may define `BTHREAD_ATTR_XXX` by themselves, but that code will be broken if we add new members to `bthread_attr_t`.

Yes. As the bthread_attr_t struct is defined in C code style without defining ctor function, so when new members is added and user have defined its  BTHREAD_ATTR_XXX by themselves and not used `bthread_attr_init`, the new added name filed will be uninitialized.  Shall we consider this scenario and how?

wwbmmm: > 
> Yes. As the bthread_attr_t struct is defined in C code style without defining ctor function, so when new members is added and user have defined its BTHREAD_ATTR_XXX by themselves and not used `bthread_attr_init`, the new added name filed will be uninitialized. Shall we consider this scenario and how?

I think the current code is OK. Maybe we need to add to document that user must use `bthread_attr_init` when they initialize `bthread_attr_t` without using pre-defined macros.

ZhengweiZhu: I think the advanced usage of user defined BTHREAD_ATTR_XXX is really rare and if so they should be able to handle the compatibility.   I have not found any document suitable for explaining it.

ZhengweiZhu:
ping

wwbmmm:
LGTM

chenBright (approved):
LGTM

PR #3163, opened 2025-11-25
Support progressive read timeout bthread handler

### LINKED ISSUE #3133: brpc持续下载，设置每一段超时
问题：
brpc server持续下载，使用brpc::ProgressiveReader，如何实现：若后端一段时间内没有发送新的数据到brpc server，就断开与后端的连接

zchuango:

Issue Number: resolve #3133 

Problem Summary:  when Controller response_will_be_read_progressively() it needs  background bthread monitor handler to solve progressive reader idle timeout，when hit the progressive reader idle timeout duration should close current client socket connection.

Changed:

1. add HandleIdleProgressiveReader method will run in and when controller call response_will_be_read_progressively method will trigger  HandleIdleProgressiveReader monitor idle progressive reader on bthread_timer.
2. add progressive idle case on http_c++ example http_server with enable_progressive_timeout arg to trigger ProgressiveAttachment write timeout and http_client with progressive and progressive_read_timeout_ms args test the progressive reader idle timeout case.

Side effects:

Performance effects: NO

Breaking backward compatibility: NO

---
### Check List:


src/brpc/controller.cpp
```diff
@@ -1027,6 +1037,43 @@ void Controller::SubmitSpan() {
     _span = NULL;
 }
 
+void* Controller::HandleIdleProgressiveReader(void* arg) {
+    auto* cntl = static_cast<Controller*>(arg);
+    const uint64_t CHECK_INTERVAL_US = 1000000UL;
```
wwbmmm: It seems that the check interval is 1s, but the resolution of progressive_read_timeout is 1ms

src/brpc/controller.cpp
```diff
@@ -1027,6 +1037,43 @@ void Controller::SubmitSpan() {
...
+    auto log_idle = FLAGS_log_idle_progressive_read_close;
+    std::vector<SocketId> remove_socket_ids;
+    while (bthread_usleep(CHECK_INTERVAL_US) == 0) {
+        // TODO: this is not efficient for a lot of connections(>100K)
+        auto socketIds = cntl->_checking_progressive_read_fds;
```
wwbmmm: Do you need to lock the Controller when you access `cntl->_checking_progressive_read_fds`?

src/brpc/controller.cpp
```diff
@@ -1027,6 +1037,43 @@ void Controller::SubmitSpan() {
     _span = NULL;
 }
 
+void* Controller::HandleIdleProgressiveReader(void* arg) {
+    auto* cntl = static_cast<Controller*>(arg);
+    const uint64_t CHECK_INTERVAL_US = 1000000UL;
+    auto log_idle = FLAGS_log_idle_progressive_read_close;
+    std::vector<SocketId> remove_socket_ids;
+    while (bthread_usleep(CHECK_INTERVAL_US) == 0) {
+        // TODO: this is not efficient for a lot of connections(>100K)
```
wwbmmm: In what scenarios will a controller have >100K connections?

src/brpc/controller.cpp
```diff
@@ -1179,6 +1226,11 @@ void Controller::IssueRPC(int64_t start_realtime_us) {
         // Tag the socket so that when the response comes back, the parser will
         // stop before reading all body.
         _current_call.sending_sock->read_will_be_progressive(_connection_type);
+        auto socket_id = _current_call.sending_sock->id();
+        if (_progressive_read_timeout_ms > 0 && _checking_progressive_read_fds.seek(socket_id) == NULL) {
+            _checking_progressive_read_fds.insert(socket_id);
+            LOG(INFO) << "insert the progressive read fd : " << socket_id << " socket fds size : " << _checking_progressive_read_fds.size();
```
wwbmmm: This is like DEBUG log. It should not be INFO level.

src/brpc/controller.h
```diff
@@ -323,7 +325,19 @@ friend void policy::ProcessThriftRequest(InputMessageBase*);
...
+    void response_will_be_read_progressively() { 
+        if(has_flag(FLAGS_READ_PROGRESSIVELY) && _progressive_read_idle_tid > 0) {
+            return;
+        }
+        bthread_attr_t tmp = BTHREAD_ATTR_NORMAL;
+        tmp.tag = _bthread_tag;
+        if(bthread_start_background(&_progressive_read_idle_tid, &tmp, HandleIdleProgressiveReader, this) != 0){
+            LOG(FATAL) << "Failed to start controller bthread id : " << _progressive_read_idle_tid;
+        }
+        LOG(INFO) << "Start Response progressive reader idle checker close idle_tid : " << _progressive_read_idle_tid
```
wwbmmm: This is like DEBUG log. It should not be INFO level.

src/brpc/controller.h
```diff
@@ -323,7 +325,19 @@ friend void policy::ProcessThriftRequest(InputMessageBase*);
...
     // Make the RPC end when the HTTP response has complete headers and let
     // user read the remaining body by using ReadProgressiveAttachmentBy().
-    void response_will_be_read_progressively() { add_flag(FLAGS_READ_PROGRESSIVELY); }
+    void response_will_be_read_progressively() { 
+        if(has_flag(FLAGS_READ_PROGRESSIVELY) && _progressive_read_idle_tid > 0) {
+            return;
+        }
+        bthread_attr_t tmp = BTHREAD_ATTR_NORMAL;
+        tmp.tag = _bthread_tag;
+        if(bthread_start_background(&_progressive_read_idle_tid, &tmp, HandleIdleProgressiveReader, this) != 0){
```
wwbmmm: Only when user set progressive_read_timeout_ms, you need to start a bthread here.

src/brpc/controller.h
```diff
@@ -837,6 +851,11 @@ friend void policy::ProcessThriftRequest(InputMessageBase*);
     int32_t _timeout_ms;
     int32_t _connect_timeout_ms;
     int32_t _backup_request_ms;
+    int32_t _progressive_read_timeout_ms;
+    butil::FlatSet<SocketId> _checking_progressive_read_fds;
+    bthread_t _progressive_read_idle_tid;
```
wwbmmm: This bthread is never joined, it will cause memory leak.

src/brpc/controller.h
```diff
@@ -837,6 +851,11 @@ friend void policy::ProcessThriftRequest(InputMessageBase*);
...
+    // Controller belongs to this tag
+    bthread_tag_t _bthread_tag = bthread_self_tag();
```
wwbmmm: It is not suitable to init bthread_tag here, the bthread_self_tag() will be execute when the Controller create, but that thread may not be the thread which issues rpc.

src/brpc/controller.h
```diff
@@ -837,6 +851,11 @@ friend void policy::ProcessThriftRequest(InputMessageBase*);
     int32_t _timeout_ms;
     int32_t _connect_timeout_ms;
     int32_t _backup_request_ms;
+    int32_t _progressive_read_timeout_ms;
+    butil::FlatSet<SocketId> _checking_progressive_read_fds;
```
wwbmmm: Maybe you don't need to add this FlatSet, you can access the Controller's _current_call and _unfinished_call to find existing connections. I think there are at most 2 connections in one Controller.

wwbmmm:
Hi @zchuang185 Thank you for you contribution! I have left some comments, but I think the biggest problem is that this solution is too heavy and too intrusive to Controller. Is there any other solution? For example, after rpc success, set a bthread_timer with progressive_read_timeout_ms and check the Socket when timeout?

src/brpc/controller.cpp
zchuango: yeah，it refer Acceptor CloseIdleConnections method Code comments copy here

src/brpc/controller.cpp
zchuango: ok fix it later

src/brpc/controller.h
zchuango: ok fix it later

src/brpc/controller.h
zchuango: great suggestion! I am seeking how could get the socket for the client-side issue. I'll try it  next commit.

zchuango:
@wwbmmm I've added  ProgressiveTimeoutRead which is a ProgressiveReader  wrapper class, when OnReadPart is invoked, it sets up a bthread_timer to monitor the idle timeout for the progressive socket reader. ProgressiveTimeoutRead is only activated when progressive_read_timeout_ms has been explicitly configured.

src/brpc/controller.cpp
```diff
@@ -1542,6 +1660,9 @@ void Controller::ReadProgressiveAttachmentBy(ProgressiveReader* r) {
                          __FUNCTION__));
     }
     add_flag(FLAGS_PROGRESSIVE_READER);
+    if (progressive_read_timeout_ms() > 0) {
+        return  _rpa->ReadProgressiveAttachmentBy(new ProgressiveTimeoutRead(this, r));
```
wwbmmm: I think a more elegant way is to add a virtual GetSocket() function to ReadableProgressiveAttachment:

```cpp
class ReadableProgressiveAttachment : public SharedObject {
public:
    virtual Socket* GetSocket() = 0;
};
```
and then return the socket in HttpContext :

```
class HttpContext : public ReadableProgressiveAttachment
                  , public InputMessageBase
                  , public HttpMessage {
public:
    Socket* GetSocket() { return _socket.get(); }
};
```

then we can just pass the socket and `_progressive_read_timeout_ms` to the new `ProgressiveTimeoutRead` here:
```
    if (progressive_read_timeout_ms() > 0) {
        return  _rpa->ReadProgressiveAttachmentBy(new ProgressiveTimeoutRead(_rpa->GetSocket(), _progressive_read_timeout_ms, r));
    }
```

Then we don't need to hold the controller in the `ProgressiveTimeoutRead`, and don't need the `_checking_progressive_read_fds` and `_progressive_read_lock` any more.

src/brpc/controller.cpp
```diff
@@ -174,6 +175,61 @@ class IgnoreAllRead : public ProgressiveReader {
     void OnEndOfMessage(const butil::Status&) {}
 };
 
+class ProgressiveTimeoutRead : public ProgressiveReader {
```
wwbmmm: ProgressiveTimeoutRead -> ProgressiveTimeoutReader

src/brpc/controller.cpp
chenBright: > Then we don't need to hold the controller in the ProgressiveTimeoutRead

Agreed. The controller may have been destroyed and can no longer be used.

src/brpc/controller.cpp
```diff
@@ -1027,6 +1092,54 @@ void Controller::SubmitSpan() {
...
+    butil::AutoLock guard(cntl->_progressive_read_lock);
+    auto socketIds = cntl->_checking_progressive_read_fds;
+    for (auto socket_id :  socketIds){
+        SocketUniquePtr s;
+        if (Socket::Address(socket_id, &s) == 0) {
+            int64_t pre_idle_duration_us = 0;
+            int64_t idle_duration_us = butil::cpuwide_time_us() - s->last_active_time_us();
+            while (progressive_read_timeout_us > idle_duration_us && idle_duration_us > pre_idle_duration_us) {
+                auto sleep_ms = (progressive_read_timeout_us - idle_duration_us) / 1000;
+                bthread_usleep(sleep_ms > 0 ? sleep_ms : 1);
```
chenBright: bthread_usleep will block TimerThread, which will affect the execution of other timers.

src/brpc/controller.cpp
```diff
@@ -1027,6 +1092,54 @@ void Controller::SubmitSpan() {
...
+                << " pre_idle_duration_us : " << pre_idle_duration_us;
+                return;
+            }
+            LOG_IF(INFO, log_idle) << "progressive read timeout socket id : " << socket_id
+            << " progressive read timeout us : " << progressive_read_timeout_us
+            << " progressive read idle duration : " << idle_duration_us;
+            if (s->parsing_context() != NULL) {
+                s->parsing_context()->Destroy();
+            }
+            s->ReleaseReferenceIfIdle(0);
```
chenBright: Does the user get a timeout error in OnEndOfMessage?

src/brpc/controller.cpp
zchuango: ok，I am trying pass the socket ptr to hold client sending socket.

src/brpc/controller.cpp
zchuango: This loop mainly addresses the situation where idle_duration_us is smaller than progressive_read_timeout_us. Generally, the difference between the two values is very small, primarily because the handler where this loop resides is triggered by a timer, and the trigger time is set by the parameter progressive_read_timeout_us.

src/brpc/controller.cpp
zchuango: There is currently no timeout error, but it can be passed through ProgressiveTimeoutRead, I will  including this feature next commit.

src/brpc/controller.cpp
chenBright: I think the fundamental problem here is deadlock. bthread_usleep is waiting for TimerThread to wake it up, and TimerThread is waiting for bthread_usleep to finish executing.

zchuango: @chenBright  I without any idea for this deadlock，I have tested it and still behaves normally. could you provide some idea resolve it?

src/brpc/controller.cpp
zchuango: @wwbmmm I tried to obtain the socket using the virtual getSocket() function, it only get a null pointer. Then I attempted to assign the socketId in the http_rpc_protocol ParseHttpMessage function, and it worked properly.
just like this：
```
class ReadableProgressiveAttachment : public SharedObject {
public:
    virtual void ReadProgressiveAttachmentBy(ProgressiveReader* r) = 0;
    virtual SocketId GetSocketId() = 0;
    virtual void SetSocketId(SocketId id) = 0;
};
```
```
class HttpContext : public ReadableProgressiveAttachment
                  , public InputMessageBase
                  , public HttpMessage {
public:
    SocketId GetSocketId() override {
        return _socket_id;
    }

    void SetSocketId(SocketId id) override {
        _socket_id = id;
    }
```
```
ParseResult ParseHttpMessage(butil::IOBuf *source, Socket *socket,
                             bool read_eof, const void* arg) {
xxx
    if (http_imsg == NULL) {
        LOG(FATAL) << "Fail to new HttpContext";
       return MakeParseError(PARSE_ERROR_NO_RESOURCE);
    }
    http_imsg->SetSocketId(socket->id());
xxx
}
```

src/brpc/controller.cpp
```diff
@@ -1028,6 +1118,44 @@ void Controller::SubmitSpan() {
     _span = NULL;
 }
 
+void Controller::HandleIdleProgressiveReader(void* arg) {
```
wwbmmm: Since this function don't access Controller now, it no longer need to be a member of Controller.

src/brpc/progressive_reader.h
```diff
@@ -84,6 +85,8 @@ class ReadableProgressiveAttachment : public SharedObject {
     // Any error occurred should destroy the reader by calling r->Destroy().
     // r->Destroy() should be guaranteed to be called once and only once.
     virtual void ReadProgressiveAttachmentBy(ProgressiveReader* r) = 0;
+    virtual SocketId GetSocketId() = 0;
+    virtual void SetSocketId(SocketId id) = 0;
```
wwbmmm: The  ReadableProgressiveAttachment interface doesn't need a virtual SetSocketId function.
The SetSocketId function can be only a member of HttpContext.

src/brpc/controller.cpp
wwbmmm: maybe you can save some status to the member of ProgressiveTimeoutReader, reset the bthread_timer and return here.

src/brpc/controller.cpp
wwbmmm: It can be a member of ProgressiveTimeoutReader

src/brpc/policy/http_rpc_protocol.h
```diff
@@ -122,6 +130,7 @@ class HttpContext : public ReadableProgressiveAttachment
 
 private:
     bool _is_stage2;
+    SocketId _socket_id;
```
chenBright, line 134: Initialization is missed.

src/brpc/controller.cpp
```diff
@@ -174,6 +175,86 @@ class IgnoreAllRead : public ProgressiveReader {
...
+        if(_read_timeout_ms < _add_timer_delay_ms) {
+            AddToTimer();
+            return;
+        }
+        auto current_ms = butil::cpuwide_time_ms();
+        if (current_ms - _latest_add_timer_ms < _add_timer_delay_ms) {
+            return;
+        }
+        _latest_add_timer_ms = current_ms;
+        AddToTimer();
```
chenBright: Why do we need to add multiple timers?

src/brpc/controller.cpp
```diff
@@ -174,6 +175,86 @@ class IgnoreAllRead : public ProgressiveReader {
...
+    }
+
+private:
+    void AddToTimer() {
+        if (_timeout_id > 0) {
+            bthread_timer_del(_timeout_id);
+        }
+        bthread_timer_add(&_timeout_id,
+            butil::milliseconds_from_now(_read_timeout_ms),
+            Controller::HandleIdleProgressiveReader,
```
chenBright: Controller::HandleIdleProgressiveReader -> ProgressiveTimeoutReader::HandleIdleProgressiveReader

src/brpc/controller.cpp
```diff
@@ -1028,6 +1118,44 @@ void Controller::SubmitSpan() {
...
+        idle_duration_us = butil::cpuwide_time_us() - s->last_active_time_us();
+    }
+    if (idle_duration_us <= pre_idle_duration_us) {
+        LOG_IF(INFO, log_idle) << "stop progressive read timeout checking process!"
+        << " progressive_read_timeout_us : " << progressive_read_timeout_us
+        << " idle_duration_us : " << idle_duration_us
+        << " pre_idle_duration_us : " << pre_idle_duration_us;
+        return;
+    }
+    reader->set_read_timeout();
```
chenBright: There is a thread safety issue here. The process will crash if the reader has already been destructed.

src/brpc/controller.cpp
zchuango: This is a delay protection logic to prevent frequent addition of timer events within the interval specified _add_timer_delay_ms. if _progressive_read_timeout_ms is less than _add_timer_delay_ms, the minimum time interval will be used.

chenBright: I think a timeout for a total progressive read is sufficient. No need to pay attention to each read. This will also prevent timer deadlock issues.

src/brpc/controller.cpp
zchuang185: ok，it can refer @chenBright suggestion idea Controller::HandleIdleProgressiveReader -> ProgressiveTimeoutReader::HandleIdleProgressiveReader

src/brpc/controller.cpp
zchuang185: deleting the timer when the reader is destructed should solve this problem？
```
~ProgressiveTimeoutReader() {
    if(_timeout_id > 0) {
        bthread_timer_del(_timeout_id);
    }
}
```

chenBright: This issue deserves attention.

src/brpc/controller.cpp
```diff
@@ -1543,6 +1629,10 @@ void Controller::ReadProgressiveAttachmentBy(ProgressiveReader* r) {
                          __FUNCTION__));
     }
     add_flag(FLAGS_PROGRESSIVE_READER);
+    if (progressive_read_timeout_ms() > 0) {
+        auto reader = new ProgressiveTimeoutReader(_rpa->GetSocketId(), _progressive_read_timeout_ms, r);
```
chenBright, line 1656: I believe that _current_call.peer_id can be used instead of GetSocketId.

src/brpc/controller.cpp
```diff
@@ -174,6 +175,80 @@ class IgnoreAllRead : public ProgressiveReader {
...
+        }
+    }
+
+    butil::Status OnReadOnePart(const void* data, size_t length) {
+        return _reader->OnReadOnePart(data, length);
+    }
+
+    void OnEndOfMessage(const butil::Status& status) {
+        if (_is_read_timeout) {
+            _reader->OnEndOfMessage(butil::Status(ECONNRESET, "The progressive read timeout"));
```
chenBright: Use EProgReadTimeout(1019) instead of ECONNRESET, and add it  it to errno.proto.

src/brpc/controller.cpp
zchuango: I have refine the code logical, not need the bthread_usleep wait timeout

wwbmmm:
LGTM

src/brpc/controller.cpp
chenBright: This problem doesn't seem to be resolved yet.

PR #3169, opened 2025-12-09
fix(protocol): fix attachment size limit issue when size exceeds uint32

### LINKED ISSUE #1348: 附件大小限制问题，当大小大于uint32, 会有问题
**Describe the bug (描述bug)**
message RpcMeta {
    optional RpcRequestMeta request = 1;
    optional RpcResponseMeta response = 2;
    optional int32 compress_type = 3;
    optional int64 correlation_id = 4;
    optional int32 attachment_size = 5;
    optional ChunkInfo chunk_info = 6;
    optional bytes authentication_data = 7;
    optional StreamSettings stream_settings = 8;   
}
attachment_size   是int32 ，但附件没有限制大小
ProcessRpcResponse

        // Parse response message iff error code from meta is 0
        butil::IOBuf res_buf;
        **const int res_size** = msg->payload.length();
        butil::IOBuf* res_buf_ptr = &msg->payload;
        if (meta.has_attachment_size()) {
            if (meta.attachment_size() > res_size) {
                cntl->SetFailed(
                    ERESPONSE,
                    "attachment_size=%d is larger than response_size=%d",
                    meta.attachment_size(), res_size);
                break;
            }
            **int att_size** = res_size - meta.attachment_size();
            msg->payload.cutn(&res_buf, att_size);
            res_buf_ptr = &res_buf;
            cntl->response_attachment().swap(msg->payload);
        }

在获取的时候都使用了int，当附件大小大于4G时会有问题。
**To Reproduce (复现方法)**

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
brpc:
protobuf:

InJunee:
- Change attachment_size from int32 to int64 in protobuf definitions (baidu_rpc_meta.proto, rpc_dump.proto, nshead_meta.proto)
- Fix type mismatches in ProcessRpcRequest and ProcessRpcResponse:
  - Change req_size/res_size from int to size_t
  - Change body_without_attachment_size from int to size_t
  - Add overflow checks and proper type casting
- Fix size inconsistency in PackRpcHeader and SerializeRpcHeaderAndMeta:
  - Change payload_size parameter from int to size_t
  - Add overflow check for meta_size + payload_size exceeding UINT32_MAX
- Update format strings to use PRId64 and %zu for proper large value display

This fix allows the protocol to correctly handle attachment sizes greater than 4GB (uint32 max), preventing integer overflow and size mismatches.

Fixes #1348

Issue Number: resolve 

Problem Summary:

Changed:

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/brpc/policy/baidu_rpc_protocol.cpp
+++ b/src/brpc/policy/baidu_rpc_protocol.cpp
@@ -26,6 +26,7 @@
 #include <google/protobuf/text_format.h>
 #include "butil/logging.h"                       // LOG()
 #include "butil/iobuf.h"                         // butil::IOBuf
+#include "butil/macros.h"                        // ALLOW_UNUSED
 #include "butil/raw_pack.h"                      // RawPacker RawUnpacker
 #include "butil/memory/scope_guard.h"
 #include "json2pb/json_to_pb.h"
@@ -97,7 +98,7 @@ static void SetAttachmentSize(RpcMeta* meta, size_t size) {
 
 // Helper function to get attachment size from RpcDumpMeta, with backward compatibility
 // Marked unused to avoid -Werror-unused-function when not referenced.
-static __attribute__((unused)) int64_t GetAttachmentSizeFromDump(const RpcDumpMeta& meta) {
+static int64_t ALLOW_UNUSED GetAttachmentSizeFromDump(const RpcDumpMeta& meta) {
     if (meta.has_attachment_size_long()) {
         return meta.attachment_size_long();
     }
--- a/test/brpc_server_unittest.cpp
+++ b/test/brpc_server_unittest.cpp
@@ -21,11 +21,13 @@
 
 #include <sys/types.h>
 #include <sys/socket.h>
+#include <cstdint>
 #include <fstream>
 #include <gtest/gtest.h>
 #include <google/protobuf/descriptor.h>
 #include "butil/time.h"
 #include "butil/macros.h"
+#include "butil/raw_pack.h"
 #include "butil/fd_guard.h"
 #include "butil/files/scoped_file.h"
 #include "brpc/socket.h"
@@ -53,6 +55,9 @@
 #include "brpc/socket_map.h"
 #include "brpc/controller.h"
 #include "brpc/compress.h"
+#include "brpc/policy/baidu_rpc_meta.pb.h"
+#include "brpc/policy/baidu_rpc_protocol.h"
+#include "brpc/policy/most_common_message.h"
 #include "echo.pb.h"
 #include "v1.pb.h"
 #include "v2.pb.h"
@@ -2068,6 +2073,45 @@ TEST_F(ServerTest, auth) {
 
     ASSERT_EQ(0, server.Stop(0));
     ASSERT_EQ(0, server.Join());
+}
+
+TEST_F(ServerTest, baidu_extended_header_parse) {
+    brpc::policy::RpcMeta meta;
+    meta.set_correlation_id(123);
+    std::string meta_buf;
+    ASSERT_TRUE(meta.SerializeToString(&meta_buf));
+
+    const std::string payload = "extended-format-payload";
+    const uint64_t body_size = meta_buf.size() + payload.size();
+
+    char header[20];
+    memcpy(header, "PRPC", 4);
+    butil::RawPacker(header + 4)
+        .pack32(UINT32_MAX)
+        .pack32(static_cast<uint32_t>(meta_buf.size()))
+        .pack64(body_size);
+
+    butil::IOBuf buf;
+    buf.append(header, sizeof(header));
+    buf.append(meta_buf);
+    buf.append(payload);
+
+    brpc::ParseResult result = brpc::policy::ParseRpcMessage(&buf, NULL, false, NULL);
+    ASSERT_TRUE(result.is_ok());
+
+    brpc::policy::MostCommonMessage* msg =
+        static_cast<brpc::policy::MostCommonMessage*>(result.message());
+    std::string parsed_payload;
+    msg->payload.copy_to(&parsed_payload);
+    EXPECT_EQ(payload, parsed_payload);
+
+    std::string parsed_meta;
+    msg->meta.copy_to(&parsed_meta);
+    brpc::policy::RpcMeta parsed_meta_pb;
+    ASSERT_TRUE(parsed_meta_pb.ParseFromString(parsed_meta));
+    EXPECT_EQ(meta.correlation_id(), parsed_meta_pb.correlation_id());
+
+    msg->Destroy();
 }
 
 } //namespace


src/brpc/policy/baidu_rpc_meta.proto
```diff
@@ -28,7 +28,7 @@ message RpcMeta {
     optional RpcResponseMeta response = 2;
     optional int32 compress_type = 3;
     optional int64 correlation_id = 4;
-    optional int32 attachment_size = 5;
+    optional int64 attachment_size = 5;
```
wwbmmm: You cannot directly change the type of an existing protobuf field. This will cause client and server of different version not compatible.
You could add a new field, like `optional int64 attachment_size_long = 13`.
If the attachment size not exceed int32, you can use the original `attachment_size` field for compatibility. If the attachment size exceed int32, you can use the `attachment_size_long` field.

src/brpc/policy/baidu_rpc_protocol.cpp
```diff
@@ -69,16 +71,29 @@ DECLARE_bool(pb_enum_as_number);
...
-        .pack32(meta_size + payload_size)
-        .pack32(meta_size);
+    // Check for overflow: meta_size + payload_size must fit in uint32_t
+    const uint64_t total_size = static_cast<uint64_t>(meta_size) + payload_size;
+    if (total_size > UINT32_MAX) {
+        LOG(ERROR) << "Total size (meta_size=" << meta_size 
+                   << " + payload_size=" << payload_size 
+                   << " = " << total_size 
+                   << ") exceeds uint32_t maximum (" << UINT32_MAX << ")";
+        // Truncate to maximum, but this will cause protocol error
```
wwbmmm: You know this will cause protocol error, so what's the meaning of this code?
Maybe you can use pack32(UINT32_MAX) as a flag, then pack64(total_size) after it?
Then you can update the corresponding parsing code in `ParseRpcMessage`.

src/brpc/nshead_meta.proto
```diff
@@ -30,7 +30,7 @@ message NsheadMeta {
     
     optional int64 correlation_id = 2;
     optional int64 log_id = 3;
-    optional int32 attachment_size = 4;
+    optional int64 attachment_size = 4;
```
wwbmmm: No need to change nshead protocol.

InJunee:
I have made modifications to the code according to your comments. Please check it. @wwbmmm

tools/rpc_replay/rpc_replay.cpp
```diff
@@ -181,11 +181,23 @@ static void* replay_thread(void* arg) {
...
+                if (attachment_size > 0 && 
+                    static_cast<size_t>(attachment_size) < sample->request.size()) {
+                    sample->request.cutn(
+                        &req.serialized_data(),
+                        sample->request.size() - static_cast<size_t>(attachment_size));
+                    cntl->request_attachment() = sample->request.movable();
+                } else {
+                    req.serialized_data() = sample->request.movable();
+                }
             } else {
```
wwbmmm: This else is never reached

wwbmmm:
Please fix the CI failure, there are some compilation errors.

InJunee:
I have made modifications to the code. Please check it. @wwbmmm

InJunee:
I have made modifications to the code. Please check it. @wwbmmm

wwbmmm:
LGTM

src/brpc/policy/baidu_rpc_protocol.cpp
```diff
@@ -62,36 +65,101 @@ DEFINE_bool(baidu_std_protocol_deliver_timeout_ms, false,
...
+    if (size > INT32_MAX_VALUE) {
+        meta->set_attachment_size_long(static_cast<int64_t>(size));
+    } else {
+        meta->set_attachment_size(static_cast<int32_t>(size));
+    }
+}
+
+// Helper function to get attachment size from RpcDumpMeta, with backward compatibility
+// Marked unused to avoid -Werror-unused-function when not referenced.
+static __attribute__((unused)) int64_t GetAttachmentSizeFromDump(const RpcDumpMeta& meta) {
```
chenBright: Please use ALLOW_UNUSED instead of __attribute__((unused)).

InJunee:
@chenBright I have made some modifications to the code. Please have a look. Please let me know where the code still needs to be modified so that the code can be merged.

InJunee:
It's a great honor to receive your approval. May I ask when my PR can be merged? This is of great significance to me. @chenBright

src/brpc/policy/baidu_rpc_protocol.cpp
```diff
@@ -62,36 +66,101 @@ DEFINE_bool(baidu_std_protocol_deliver_timeout_ms, false,
 DECLARE_bool(pb_enum_as_number);
 
 // Notes:
-// 1. 12-byte header [PRPC][body_size][meta_size]
+// 1. Header format:
+//    - Normal format (12 bytes): [PRPC][body_size(32bit)][meta_size(32bit)]
+//    - Extended format (20 bytes): [PRPC][UINT32_MAX][meta_size(32bit)][body_size(64bit)]
+//      Extended format is used when body_size > UINT32_MAX
```
chenBright, line 72: If body_size equals UINT32_MAX, the client uses the normal format, while the server uses the extended format.

chenBright: - Normal format (12 bytes): [PRPC][body_size(32bit)][meta_size(32bit)]
- Extended format (16 bytes): [LRPC][body_size(64bit)][meta_size(32bit)] , L represents long

The two formats have the same structure, only the magic number and body size differ. This makes them simpler to process.

@wwbmmm What do you think of the approach?

InJunee: Is there anything else that needs to be modified in the code?

chenBright: I think there's nothing else that needs to be changed.

wwbmmm: > * Normal format (12 bytes): [PRPC][body_size(32bit)][meta_size(32bit)]
> * Extended format (16 bytes): [LRPC][body_size(64bit)][meta_size(32bit)] , L represents long
> 
> The two formats have the same structure, only the magic number and body size differ. This makes them simpler to process.
> 
> @wwbmmm What do you think of the approach?

I think this approach is clearer, but current approach is also acceptable.

chenBright: > If body_size equals UINT32_MAX, the client uses the normal format, while the server uses the extended format.

@wwbmmm I think that this incompatible case is unacceptable.

wwbmmm: > > If body_size equals UINT32_MAX, the client uses the normal format, while the server uses the extended format.
> 
> @wwbmmm I think that this incompatible case is unacceptable.

I agree with you.

Issue #3183, opened 2025-12-28
Fix SIGSEGV during static initialization when /proc/self/io is missing (glog uninitialized)


jesson1:
**Describe the bug**
It appears that brpc is initializing global variables during startup and encounters a failure in accessing `/proc/self/io`. This triggers a call to glog to output a log message, but glog itself has not been initialized yet, leading to a `SIGSEGV`.

```
(gdb) bt
#0  0x00000055933de478 in google::LogMessage::Init(char const*, int, int, void (google::LogMessage::*)()) ()
#1  0x00000055933de9d4 in google::LogMessage::LogMessage(char const*, int, int, int, void (google::LogMessage::*)()) ()
#2  0x0000005592f1c6f0 in bvar::read_proc_io (s=s@entry=0x7fc9ae7448) at /usr/include/glog/logging.h:1381
#3  0x0000005592f1f2c8 in bvar::ProcIOReader::operator() (stat=0x7fc9ae7448, this=<optimized out>) at /brpc/src/bvar/default_variables.cpp:466
#4  bvar::CachedReader<bvar::ProcIO>::get_value<bvar::ProcIOReader> (fn=...) at /brpc/src/bvar/default_variables.cpp:154
#5  bvar::ProcIOReader::get_field<unsigned long, 0ul> () at /brpc/src/bvar/default_variables.cpp:471
#6  0x0000005592f1de60 in bvar::PassiveStatus<unsigned long>::get_value (this=<optimized out>) at /brpc/src/bvar/passive_status.h:140
#7  bvar::detail::ReducerSampler<bvar::PassiveStatus<unsigned long>, unsigned long, bvar::detail::AddTo<unsigned long>, bvar::detail::MinusFrom<unsigned long> >::take_sample (this=this@entry=0x7fac590080) at /brpc/src/bvar/detail/sampler.h:137
#8  0x0000005592f227a8 in bvar::detail::ReducerSampler<bvar::PassiveStatus<unsigned long>, unsigned long, bvar::detail::AddTo<unsigned long>, bvar::detail::MinusFrom<unsigned long> >::ReducerSampler (reducer=0x5593c55b38 <bvar::g_rchar>, this=0x7fac590080)
    at /brpc/src/bvar/detail/sampler.h:98
#9  bvar::PassiveStatus<unsigned long>::get_sampler (this=0x5593c55b38 <bvar::g_rchar>) at /brpc/src/bvar/passive_status.h:145
[...]

PR #3184, opened 2025-12-29
fix: avoid SIGSEGV in read_proc_io during static initialization

jesson1:

Issue Number: resolve https://github.com/apache/brpc/issues/3183

Problem Summary:
When /proc/self/io is inaccessible, read_proc_io calls PLOG during global static construction. If glog is not yet initialized, it triggers a SIGSEGV.

Changed:
Replace PLOG_ONCE with a guarded fprintf(stderr, ...) to ensure safety during early startup. A static flag is used to maintain the "print once" behavior.

Side effects:
- Performance effects: none

- Breaking backward compatibility: none

---
### Check List:


wwbmmm:
LGTM

PR #3189, opened 2026-01-07
surpport tag for selective channel

altman08:

Issue Number: resolve 

Problem Summary: To support tags for SelectiveChannel due to the requirements of the loadbalance plugin.

Changed:

Side effects:
1.SelectiveChannel 支持tag传递给load balance插件
2.修复一个SelectiveChannel 开启backup request的时候，可能的core问题

---
### Check List:

### DIFF
--- a/src/brpc/selective_channel.cpp
+++ b/src/brpc/selective_channel.cpp
@@ -83,9 +83,9 @@ public:
     ChannelBalancer() {}
     ~ChannelBalancer();
     int Init(const char* lb_name);
-    int AddChannel(ChannelBase* sub_channel,
+    int AddChannel(ChannelBase* sub_channel, const std::string& tag,
                    SelectiveChannel::ChannelHandle* handle);
-    void RemoveAndDestroyChannel(SelectiveChannel::ChannelHandle handle);
+    void RemoveAndDestroyChannel(const SelectiveChannel::ChannelHandle& handle);
     int SelectChannel(const LoadBalancer::SelectIn& in, SelectOut* out);
     int CheckHealth();
     void Describe(std::ostream& os, const DescribeOptions&);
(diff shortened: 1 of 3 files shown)


src/brpc/selective_channel.cpp
```diff
@@ -534,16 +533,22 @@ bool SelectiveChannel::initialized() const {
...
                                  ChannelHandle* handle) {
+    return AddChannel(sub_channel, "", handle);
+}
+
+int SelectiveChannel::AddChannel(ChannelBase* sub_channel,
+                                 const std::string& tag,
+                                 ChannelHandle* handle) {
     schan::ChannelBalancer* lb =
-        static_cast<schan::ChannelBalancer*>(_chan._lb.get());
+    static_cast<schan::ChannelBalancer*>(_chan._lb.get());
```
chenBright: No need to modify this line.

wwbmmm:
Can you explain in what scenarios does a selective channel need a tag ?

altman08:
> Can you explain in what scenarios does a selective channel need a tag ?

I need to implement my own load balancer plugin, which requires passing parameters via load balancer tags. Similarly, if using built-in load balancer plugins like WRR, I also need to pass parameters via tags.

wwbmmm:
LGTM

yanglimingcn:
LGTM

altman08:
> LGTM

When using the selective channel, I encountered a coredump issue. In the case of a selective channel timeout, the main_cntl rest the lb but does not destroy the call_id. When the sub channel's RPC ends, it can still lock the call_id and call lb's feedback, causing a core dump. I simply added a check for the validity of lb in the feedback function, but I'm not sure if you have better suggestions to resolve this issue.

src/brpc/selective_channel.cpp
altman08: > No need to modify this line.

fixed

chenBright:
> I simply added a check for the validity of lb in the feedback function

I think this will work.

Furthermore, Is it possible that if sub_done finds that main_cntl is Failed with the error code ERPCTIMEDOUT in SubDone::Run,  it can skip executing OnVersionedRPCReturned function and return directly？

altman08:
> SubDone::Run

If main_cntl times out and SubDone::Run returns directly without calling OnVersionedRPCReturned, it will prevent Sender::Run from being invoked to clear resource. However, if OnVersionedRPCReturned is called, the issue of lb being set to null still persists.

chenBright (approved):
LGTM

PR #3193, opened 2026-01-09
Fix port parsing validation in str2endpoint

### LINKED ISSUE #3192: 端口解析验证问题(Incorrect port number validation in address parsing)
**Describe the bug**
file: butil/endpoint.cpp
function: int str2endpoint(const char* str, EndPoint* point) {
line: 291 for (++end; isspace(*end); ++end);
first "++end" is bug?  should be for (; isspace(*end); ++end);
第一个"++end"是bug吧？应该是  for ( ; isspace(*end); ++end);
char* end = NULL;
    point->port = strtol(str + i, &end, 10);
    if (end == str + i) {
        return -1;
    } else if (*end) {
        for (++end; isspace(*end); ++end);  // <-- BUG: increments end before checking
        if (*end) {
            return -1;
        }
    }
example: "0.0.0.0:8000a" is ok
例如"0.0.0.0:8000a" 中的a就被跳过了。
**To Reproduce**

**Expected behavior**

**Versions**  1.14.1
OS:
Compiler:
brpc:
protobuf:

AnantShukla-Git-Hub:
Issue Number: resolves #3192

Problem Summary:
`str2endpoint(const char* str, EndPoint* point)` incorrectly accepts invalid
endpoint strings where the port number is followed by non-whitespace
characters (e.g. "127.0.0.1:8000a").  
This happens because the current validation logic skips the first non-digit
character before checking for whitespace, causing invalid ports to be treated
as valid.

Changed:
- Fixed port validation logic in `str2endpoint` by ensuring that only
  whitespace characters are skipped after port parsing.
- Invalid port strings with trailing non-whitespace characters are now
  correctly rejected.

### DIFF
--- a/test/endpoint_unittest.cpp
+++ b/test/endpoint_unittest.cpp
@@ -115,6 +115,19 @@ TEST(EndPointTest, endpoint) {
     ASSERT_EQ(289, p6.port);
 #endif
 }
+TEST(EndPointTest, endpoint_reject_trailing_characters_after_port) {
+    butil::EndPoint ep;
+
+    // invalid: non-whitespace after port
+    ASSERT_EQ(-1, butil::str2endpoint("127.0.0.1:8000a", &ep));
+    ASSERT_EQ(-1, butil::str2endpoint("127.0.0.1:8000#", &ep));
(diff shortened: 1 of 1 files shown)


AnantShukla-Git-Hub:
👍Added unit tests in endpoint_unittest.cpp.

PR #3198, opened 2026-01-16
support change ownship for SelectiveChannel

altman08:

Issue Number: resolve 

Problem Summary:

Changed:

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/brpc/selective_channel.h
+++ b/src/brpc/selective_channel.h
@@ -56,6 +56,11 @@ public:
         std::string tag;
     };
 
+    struct SubChannelOptions {
+        std::string tag;
+        ChannelOwnership ownership = OWNS_CHANNEL;
+    };
+
     SelectiveChannel();
     ~SelectiveChannel();
 
@@ -69,8 +74,16 @@ public:
     // On success, handle is set with the key for removal.
     // NOTE: Different from pchan, schan can add channels at any time.
     // Returns 0 on success, -1 otherwise.
-    int AddChannel(ChannelBase* sub_channel, ChannelHandle* handle);
-    int AddChannel(ChannelBase* sub_channel, const std::string& tag, ChannelHandle* handle);
+    int AddChannel(ChannelBase* sub_channel, ChannelHandle* handle) {
+        return AddChannel(sub_channel, SubChannelOptions(), handle);
+    }
+    int AddChannel(ChannelBase* sub_channel, const std::string& tag, ChannelHandle* handle) {
+        SubChannelOptions option;
+        option.tag = tag;
+        return AddChannel(sub_channel, option, handle);
+    }
+    int AddChannel(ChannelBase* sub_channel, const SubChannelOptions& option,
+                   ChannelHandle* handle);
 
     // Remove and destroy the sub_channel associated with `handle'.
     void RemoveAndDestroyChannel(const ChannelHandle& handle);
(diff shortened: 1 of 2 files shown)


yanglimingcn:
LGTM

src/brpc/selective_channel.h
```diff
@@ -69,8 +69,19 @@ class SelectiveChannel : public ChannelBase/*non-copyable*/ {
...
+    int AddChannel(ChannelBase* sub_channel, const std::string& tag,
+                   ChannelHandle* handle) {
+      return AddChannel(sub_channel, tag, OWNS_CHANNEL, handle);
+    }
+    int AddChannel(ChannelBase* sub_channel, ChannelOwnership ownership,
+                   ChannelHandle* handle) {
+      return AddChannel(sub_channel, "", ownership, handle);
+    }
+    int AddChannel(ChannelBase* sub_channel, const std::string& tag,
+                   ChannelOwnership ownership, ChannelHandle* handle);
```
chenBright: I think AddChannel has a combinatorial explosion problem.

```c++
int AddChannel(SelectiveChannel::SubChannelOptions);
```

Using the SubChannelOptions overloaded AddChannel function would be more suitable, as it offers better extensibility.

In the meantime, please update the comments for the AddChannel function.

wwbmmm: agree

altman08: fixed

src/brpc/selective_channel.h
```diff
@@ -56,6 +56,11 @@ class SelectiveChannel : public ChannelBase/*non-copyable*/ {
         std::string tag;
     };
 
+    struct SubChannelOptions {
```
yanglimingcn, line 59: Do we need a default value?

altman08: > Do we need a default value?

fixed

yanglimingcn:
LGTM

wwbmmm:
LGTM

src/brpc/selective_channel.h
```diff
@@ -69,8 +74,11 @@ class SelectiveChannel : public ChannelBase/*non-copyable*/ {
     // On success, handle is set with the key for removal.
     // NOTE: Different from pchan, schan can add channels at any time.
     // Returns 0 on success, -1 otherwise.
-    int AddChannel(ChannelBase* sub_channel, ChannelHandle* handle);
-    int AddChannel(ChannelBase* sub_channel, const std::string& tag, ChannelHandle* handle);
```
chenBright, line 73: For compatibility reasons, this function should be retained.

```c++
int AddChannel(ChannelBase* sub_channel, const std::string& tag, ChannelHandle* handle);
```

src/brpc/selective_channel.h
chenBright, line 59: Is it acceptable to make ChannelHandle a member variable of SubChannelOptions?

src/brpc/selective_channel.h
altman08: This function was merged a few days ago but hasn't been released yet. I think it's ok to remove it.

src/brpc/selective_channel.h
altman08: ChannelHandle is not suitable to be placed inside the option because it is returned to the user.

src/brpc/selective_channel.h
chenBright: This commit is already included in the currently released version [1.16](https://github.com/apache/brpc/tree/release-1.16).

altman08: ok, i add it back

PR #3222, opened 2026-02-22
feat(redis): add native Redis Cluster channel support

wayslog:
## Summary
- add brpc RedisClusterChannel for native Redis Cluster slot-based routing
- support MOVED/ASK redirection retry, topology refresh from CLUSTER SLOTS with CLUSTER NODES fallback
- do not persist ASK redirection into the slot cache (only MOVED updates the stable slot map)
- protect the slot->endpoint cache with butil::DoublyBufferedData to reduce lock contention on high-QPS reads (writes happen on MOVED or topology refresh)
- add a unit test that verifies slot cache updates on topology change via periodic refresh
- support sync/async calls, pipeline, and common multi-key commands (MGET/MSET/DEL/EXISTS/UNLINK/EVAL/EVALSHA)
- add a new Redis Cluster example client under example/redis_c++/redis_cluster_client.cpp
- update Redis client docs in both English and Chinese

## Tests
- bazel test //test:brpc_redis_cluster_test --test_output=errors

### DIFF
--- a/src/brpc/redis_cluster.cpp
+++ b/src/brpc/redis_cluster.cpp
@@ -166,7 +166,10 @@ RedisClusterChannel::RedisClusterChannel()
     : _stop_refresh(false)
     , _refresh_started(false)
     , _refresh_tid(0) {
-    _slot_to_endpoint.resize(kRedisClusterSlotCount);
+    _db_slot_to_endpoint.Modify([](std::vector<std::string>& bg) -> size_t {
+        bg.assign(kRedisClusterSlotCount, std::string());
+        return 1;
+    });
 }
 
 RedisClusterChannel::~RedisClusterChannel() {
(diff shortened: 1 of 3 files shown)


wayslog:
Looks like this CI failure is not code-related but network-related:  
`rules_cc-0.1.1.tar.gz` download from GitHub returned **502 Bad Gateway / Proxy Error** during Bazel external fetch.
In other words: our code is fine, but the network was flaky.
I have triggered a rerun. If it flakes again with the same fetch error, we can treat this as infra noise and rerun once more.

wayslog:
I investigated the flaky test failure in RedisClusterChannelTest.async_concurrent_calls_with_mixed_redirections and pushed a fix.

- commit: 90ada390
- root cause: ASK redirection was being written into the slot cache, which is incorrect (ASK is temporary during migration).
- effect: subsequent requests could be routed directly to the ASK target without sending ASKING, then return ERR ASKING required intermittently under concurrency.
- fix: only persist slot mapping for MOVED, not ASK.
- test: added ask_redirection_does_not_override_slot_cache and re-ran bazel test //test:brpc_redis_cluster_test --test_output=errors.

This should eliminate the CI flake for that case.

wwbmmm:
Can compare this with PR #2228. I prefer current PR because it only add a RedisClusterChannel class, without modifying bRPC core classes, that's amazing!

wwbmmm:
LGTM

wayslog:
thanks for @wwbmmm to review code change again. I just replaced mutex with DoublyBufferedData to reduce lock cost.
The CI still fail due to apt source issues, but code is fine. Perhaps we need re-trigger it after all.

chenBright:
Please merge or rebase master branch which fixed CI failure.

wayslog:
All tests was passed.

PR #3223, opened 2026-02-23
fix: remove references to non-existent tracked_objects.h in platform_thread_freebsd.cc

tuaris:
### What problem does this PR solve?

Issue Number: N/A

Problem Summary:

`platform_thread_freebsd.cc` includes `butil/tracked_objects.h` and calls
`tracked_objects::ThreadData::InitializeThreadContext(name)`, but this header
and class were part of Chromium's base library profiling subsystem and were
never ported to brpc. This causes a compile error on FreeBSD:

```
fatal error: 'butil/tracked_objects.h' file not found
```

The Linux (`platform_thread_linux.cc`) and macOS (`platform_thread_mac.mm`)
equivalents already had these references removed.

This is PR 1 of 3 for FreeBSD support:
1. **This PR** — Remove dead `tracked_objects.h` references
2. #3224
3. #3225

This first PR is intentionally small to get a feel for the project's review workflow.

### What is changed and the side effects?

Changed:

- Removed `#include "butil/tracked_objects.h"` from `platform_thread_freebsd.cc`
- Removed the call to `tracked_objects::ThreadData::InitializeThreadContext(name)` from `PlatformThread::SetName()`

This is pure dead code removal — the referenced header does not exist anywhere in the brpc source tree.

Side effects:
- Performance effects:
None

- Breaking backward compatibility:
None — the removed code could never have compiled successfully.

---
### Check List:
- Please make sure your changes are compatible.
- When providing us with a new feature, it is best to add related tests.
- Please follow [Contributor Covenant Code of Conduct](https://github.com/apache/brpc/blob/master/CODE_OF_CONDUCT.md).

PR #3224, opened 2026-02-23
feat: improve FreeBSD platform support

tuaris:

Issue Number: N/A

Problem Summary:

brpc has partial FreeBSD support (27 source files already contain `OS_FREEBSD`/`__FreeBSD__` references), but it does not compile or run on FreeBSD. This PR completes that support.

The most critical issue is a **static initialization deadlock**: bthread interposes `pthread_mutex_lock` for contention profiling. On FreeBSD, during early process init, `libc++.so`'s `ios_base::Init::Init()` calls `pthread_mutex_lock` to initialize `std::locale::classic()`. bthread's interposed version calls `pthread_once(init_sys_mutex_lock)`, which itself needs `pthread_mutex_lock` — causing infinite recursion. Without the `NO_PTHREAD_MUTEX_HOOK` fix in this PR, **any FreeBSD application linking brpc will deadlock at startup**.

This is PR 2 of 3 for FreeBSD support:
1. #3223
2. **This PR** — FreeBSD platform support
3. #3225 

Changed:

**All source changes are guarded by `#if defined(OS_FREEBSD)` — zero impact on Linux or macOS builds.** Most changes piggyback on existing macOS/Darwin code paths since both use kqueue.

CMakeLists.txt:
- Add `NO_PTHREAD_MUTEX_HOOK` on FreeBSD to prevent static init deadlock caused by bthread's `pthread_mutex_lock` interposition conflicting with `libc++` `ios_base::Init` during early process initialization
- Add `-lexecinfo` for backtrace support
- Wire existing FreeBSD source files into the build (`platform_thread_freebsd.cc`, `sys_string_conversions_posix.cc` — already present in the tree but not compiled)

Source changes:
- `butil/compat.h`: Add FreeBSD kqueue path and `pthread_getthreadid_np()`
- `butil/process_util.cc`: FreeBSD `ReadCommandLine` via `ps` (like macOS)
- `butil/fd_utility.cpp`: kqueue fd handling
- `butil/logging.cc`: Rename `FLAGS_v`/`vmodule`/`minloglevel` to avoid gflags duplicate registration with system glog on FreeBSD
- `brpc/socket.cpp`: FreeBSD `TCP_INFO` support
- `bthread/`: FreeBSD futex emulation, context switching, fd handling
- `bvar/default_variables.cpp`: FreeBSD `/proc` fallback

Side effects:
- Performance effects:
None — all changes are behind `OS_FREEBSD` guards and do not affect existing platforms.

None — this is additive FreeBSD support only.

---
### Check List:
- Please make sure your changes are compatible.


Huixxi:
LGTM

PR #3247, opened 2026-03-23
cputime optimize for arm64

dwh110:

Issue Number: n/a

Problem Summary: cpugettime optimize for ARM64

Changed:
  detail::invariant_cpu_freq value get  by CNTFRQ_EL0

Side effects:
- Performance effects:   

- Breaking backward compatibility: 无

---
### Check List:


src/butil/time.h
```diff
@@ -279,7 +279,7 @@ extern int64_t invariant_cpu_freq;
 // note: Inlining shortens time cost per-call for 15ns in a loop of many
 //       calls to this function.
 inline int64_t cpuwide_time_ns() {
-#if !defined(BAIDU_INTERNAL)
+#if !defined(BAIDU_INTERNAL) && !defined(__aarch64__)
```
wasphin: `BAIDU_INTERNAL` and `__aarch64__` are options at different dimensions. It is not recommended to combine them in a single conditional directive simply because the same code could be reused.

In addition, please also consider whether BAIDU_INTERNAL still needs to be retained. cc @wwbmmm

src/butil/time.cpp
```diff
@@ -152,7 +152,17 @@ int64_t read_invariant_cpu_frequency() {
...
 }
 
-int64_t invariant_cpu_freq = -1;
+static int64_t init_invariant_cpu_freq() {
+    int64_t base_freq_hz = -1;
+#if defined(__aarch64__)
+    __asm__ __volatile__("mrs %0, CNTFRQ_EL0" : "=r"(base_freq_hz));
+#else
+    base_freq_hz = detail::read_invariant_cpu_frequency();
+#endif
```
wasphin: Would it be better to just implement inside function `read_invariant_cpu_frequency`?

src/butil/time.h
wwbmmm: I don't know. Perhaps we can create a PR to remove all BAIDU_INTERNAL code, but in this PR I suggest to keep compatibility.

dwh110:
@wwbmmm  @wasphin  Could you help take a look at how to resolve this?
image

dwh110:
@wwbmmm @wasphin  please check it, looking forward to merging it into the main branch.

wwbmmm:
LGTM

src/butil/time.h
chenBright: Add a macro switch, such as BUTIL_USE_CPU_FREQUENCY, to let users choose whether to enable it instead of enabling it by default on arm64.

chenBright (approved):

src/butil/time.h
dwh110: @chenBright  Add the macro switch BUTIL_USE_CPU_FREQUENCY while preserving the original default behavior for ARM64

chenBright (changes requested):
The build script needs to support enabling `BUTIL_USE_CPU_FREQUENCY`, see `BRPC_WITH_GLOG`.

dwh110:
> The build script needs to support enabling `BUTIL_USE_CPU_FREQUENCY`, see `BRPC_WITH_GLOG`.

@chenBright OK，Referenced `BRPC_WITH_GLOG` for the following adaptations：
1、Activation options: CMake -DWITH_CPU_FREQUENCY=ON, Bazel --define BUTIL_USE_CPU_FREQUENCY=true, script argument --with-cpu-frequency.
2、Disabled by default for consistent legacy behavior.

dwh110:
@chenBright  pls review again，tks

wasphin:
LGTM

chenBright (approved):
LGTM

PR #3250, opened 2026-03-25
Fix abort when binding response streams to a failed socket

amoxic:

Issue Number: resolves #3245

Problem Summary:

When a peer disconnects after `StreamAccept()` succeeds but before the RPC response binds the response stream to the host socket, `Socket::AddStream()` can fail inside `Stream::SetHostSocket()`. The previous code treated this as an impossible state and aborted the server with `CHECK(false)`.

This PR turns that path into a normal runtime failure so the current RPC/stream is failed and the server process stays alive.

Changed:
- Make `Stream::SetHostSocket()` return `-1` with `errno` instead of aborting when `Socket::AddStream()` fails.
- Handle response stream bind failures in `SendRpcResponse()` by setting controller failure, failing the affected stream(s), and returning early.
- Keep response write completion joined through a scope guard so tracing data is still finalized on early returns.
- Add unit tests for `SetHostSocket()` failure and the server-side disconnect-before-response path.

Side effects:
- Performance effects:

  No expected steady-state performance impact. Extra work is only added on the error path.

  No API change. The behavior changes from process abort to RPC/stream failure when the host socket is already failed.

---
### Check List:

### DIFF
--- a/src/brpc/policy/baidu_rpc_protocol.cpp
+++ b/src/brpc/policy/baidu_rpc_protocol.cpp
@@ -360,7 +360,19 @@ void SendRpcResponse(int64_t correlation_id, Controller* cntl,
             Stream* s = (Stream *) stream_ptr->conn();
             StreamSettings *stream_settings = meta.mutable_stream_settings();
             s->FillSettings(stream_settings);
-            s->SetHostSocket(sock);
+            if (s->SetHostSocket(sock) != 0) {
+                const int errcode = errno;
+                LOG_IF(WARNING, errcode != EPIPE)
+                    << "Fail to bind response stream=" << response_stream_id
+                    << " to " << sock->description() << ": "
+                    << berror(errcode);
+                cntl->SetFailed(errcode, "Fail to bind response stream to %s",
+                                sock->description().c_str());
+                Stream::SetFailed(response_stream_ids, errcode,
+                                  "Fail to bind response stream to %s",
+                                  sock->description().c_str());
+                return;
+            }
             for (size_t i = 1; i < response_stream_ids.size(); ++i) {
                 stream_settings->mutable_extra_stream_ids()->Add(response_stream_ids[i]);
             }
@@ -390,6 +402,15 @@ void SendRpcResponse(int64_t correlation_id, Controller* cntl,
 
     ResponseWriteInfo args;
     bthread_id_t response_id = INVALID_BTHREAD_ID;
+    auto response_write_guard = butil::MakeScopeGuard([&response_id, &args, span] {
+        if (response_id == INVALID_BTHREAD_ID) {
+            return;
+        }
+        bthread_id_join(response_id);
+        // Do not care about the result of background writing.
+        // TODO: this is not sent
+        span->set_sent_us(args.sent_us);
+    });
     if (span) {
         span->set_response_size(res_buf.size());
         CHECK_EQ(0, bthread_id_create(&response_id, &args, HandleResponseWritten));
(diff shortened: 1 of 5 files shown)


hjwsm1989:
@chenBright please help review

chenBright:
cc @jenrryyou

amoxic:
Thanks for the review. Addressed all Copilot comments:

1) Stream::SetHostSocket now records the first failure in _set_host_socket_ec and always sets errno from that value on failure (no more call_once stack-local loss).
2) All callers now check SetHostSocket return and fail controller/streams on error: SendRpcResponse main stream + extra streams, Controller::HandleStreamConnection.
3) Bind failures propagate errno directly (no reliance on sock->non_zero_error_code).
4) Stream::SetFailed now preformats the reason string (SetFailedWithReason helper) to avoid varargs UB; all new call sites use the safe path.
5) Added regression tests covering SetHostSocket failure and server-side disconnect-before-response; both pass locally (`brpc_streaming_rpc_unittest` filtered run).

amoxic:
@chenBright could you please review this pr again

src/brpc/policy/baidu_rpc_protocol.cpp
```diff
@@ -360,7 +360,19 @@ void SendRpcResponse(int64_t correlation_id, Controller* cntl,
...
+                LOG_IF(WARNING, errcode != EPIPE)
+                    << "Fail to bind response stream=" << response_stream_id
+                    << " to " << sock->description() << ": "
+                    << berror(errcode);
+                cntl->SetFailed(errcode, "Fail to bind response stream to %s",
+                                sock->description().c_str());
+                Stream::SetFailed(response_stream_ids, errcode,
+                                  "Fail to bind response stream to %s",
+                                  sock->description().c_str());
+                return;
```
chenBright, line 374: I think that if binding the response stream fails, we shouldn't just return directly; we should still try writing a response.

src/brpc/policy/baidu_rpc_protocol.cpp
```diff
@@ -426,7 +447,21 @@ void SendRpcResponse(int64_t correlation_id, Controller* cntl,
...
                 Stream* extra_stream = (Stream *) extra_stream_ptr->conn();
-                extra_stream->SetHostSocket(sock);
+                if (extra_stream->SetHostSocket(sock) != 0) {
+                    const int errcode = errno;
+                    LOG_IF(WARNING, errcode != EPIPE)
+                        << "Fail to bind response stream=" << extra_stream_id
+                        << " to " << sock->description() << ": "
+                        << berror(errcode);
+                    cntl->SetFailed(errcode, "Fail to bind response stream to %s",
+                                    sock->description().c_str());
```
chenBright, line 457: Just call `cntl->SetFailed` once.

src/brpc/policy/baidu_rpc_protocol.cpp
```diff
@@ -360,7 +360,19 @@ void SendRpcResponse(int64_t correlation_id, Controller* cntl,
             Stream* s = (Stream *) stream_ptr->conn();
             StreamSettings *stream_settings = meta.mutable_stream_settings();
             s->FillSettings(stream_settings);
-            s->SetHostSocket(sock);
+            if (s->SetHostSocket(sock) != 0) {
+                const int errcode = errno;
```
chenBright, line 364: Why use errno as the reason for SetHostSocket failure? SetHostSocket failure does not set errno.

chenBright:
> @chenBright could you please review this pr again

@amoxic Sorry, I missed the message. I've already reviewed the PR.

PR #3255, opened 2026-03-29
support rdma ece

randomkang:
https://github.com/apache/brpc/issues/3246

Changed:

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/brpc/rdma/rdma_endpoint.cpp
+++ b/src/brpc/rdma/rdma_endpoint.cpp
@@ -46,6 +46,8 @@ extern ibv_qp* (*IbvCreateQp)(ibv_pd*, ibv_qp_init_attr*);
 extern int (*IbvModifyQp)(ibv_qp*, ibv_qp_attr*, ibv_qp_attr_mask);
 extern int (*IbvQueryQp)(ibv_qp*, ibv_qp_attr*, ibv_qp_attr_mask, ibv_qp_init_attr*);
 extern int (*IbvDestroyQp)(ibv_qp*);
+extern int (*IbvQueryEce)(ibv_qp*, ibv_ece*);
+extern int (*IbvSetEce)(ibv_qp*, ibv_ece*);
 extern bool g_skip_rdma_init;
 
 DEFINE_int32(rdma_sq_size, 128, "SQ size for RDMA");
@@ -64,6 +66,7 @@ DEFINE_int32(rdma_poller_num, 1, "Poller number in RDMA polling mode.");
 DEFINE_bool(rdma_poller_yield, false, "Yield thread in RDMA polling mode.");
 DEFINE_bool(rdma_edisp_unsched, false, "Disable event dispatcher schedule");
 DEFINE_bool(rdma_disable_bthread, false, "Disable bthread in RDMA");
+DEFINE_bool(rdma_ece, false, "Open ece in RDMA, should use this feature when rdma nics are from the same merchant.");
 
 static const size_t IOBUF_BLOCK_HEADER_LEN = 32; // implementation-dependent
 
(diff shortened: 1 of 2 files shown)


src/brpc/rdma/rdma_helper.cpp
```diff
@@ -386,6 +388,8 @@ static int ReadRdmaDynamicLib() {
     LoadSymbol(g_handle_ibverbs, IbvGetAsyncEvent, "ibv_get_async_event");
     LoadSymbol(g_handle_ibverbs, IbvAckAsyncEvent, "ibv_ack_async_event");
     LoadSymbol(g_handle_ibverbs, IbvEventTypeStr, "ibv_event_type_str");
+    LoadSymbol(g_handle_ibverbs, IbvQueryEce, "ibv_query_ece");
+    LoadSymbol(g_handle_ibverbs, IbvSetEce, "ibv_set_ece");
```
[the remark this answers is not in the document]
yanglimingcn: The LoadSymbol macro call currently does not handle load failures, so ECE-related API loading failures will not have any impact.

yanglimingcn:
Have you tested the ECE logic?
1. Both client and server are enabled.
2. Client is enabled, server is disabled.
3. Client is disabled, server is enabled.
4. Client is disabled, server is disabled.

src/brpc/rdma/rdma_endpoint.cpp
```diff
@@ -1250,6 +1206,21 @@ int RdmaEndpoint::BringUpQp(uint16_t lid, ibv_gid gid, uint32_t qp_num) {
...
     }
 
+    if (FLAGS_rdma_ece) {
+        struct ibv_ece ece;
+        int err = IbvQueryEce(_resource->qp, &ece);
+        if (err != 0) {
+            LOG(WARNING) << "Fail to IbvQueryEce: " << berror(err);
+            return -1;
+        }
+        // ToDo: should check if remote qp support ece
```
chenBright, line 1216: What happens if one side does not support ECE? Is negotiation necessary?

randomkang:
> Have you tested the ECE logic?
> 
> 1. Both client and server are enabled.
> 2. Client is enabled, server is disabled.
> 3. Client is disabled, server is enabled.
> 4. Client is disabled, server is disabled.

I test 1 and 4, and it works well. ECE is used for the server and client to negotiate the advanced features of the RDMA network card, so 2 and 3 is illegal and i don't test these cases.

src/brpc/rdma/rdma_endpoint.cpp
randomkang: I don't know. All of our clusters use the same type of RDMA network card, so we can't test what the effect would be when different network card models all have ECE enabled.

yanglimingcn:
> > Have you tested the ECE logic?
> > 
> > 1. Both client and server are enabled.
> > 2. Client is enabled, server is disabled.
> > 3. Client is disabled, server is enabled.
> > 4. Client is disabled, server is disabled.
> 
> I test 1 and 4, and it works well. ECE is used for the server and client to negotiate the advanced features of the RDMA network card, so 2 and 3 is illegal and i don't test these cases.

So for cases 2 and 3, we should get an error. Could you try running it with this configuration and see what the result is?

randomkang:
> > > Have you tested the ECE logic?
> > > 
> > > 1. Both client and server are enabled.
> > > 2. Client is enabled, server is disabled.
> > > 3. Client is disabled, server is enabled.
> > > 4. Client is disabled, server is disabled.
> > 
> > 
> > I test 1 and 4, and it works well. ECE is used for the server and client to negotiate the advanced features of the RDMA network card, so 2 and 3 is illegal and i don't test these cases.
> 
> So for cases 2 and 3, we should get an error. Could you try running it with this configuration and see what the result is?

I test 2 and 3 on the model training task, and the task does not report any errors. The rdma nic i used is Mellanox Technologies MT2910 Family（ConnectX-7）. The ece option for this rdma nic is 0011 0000 0000 0000 0000 0000 0000 0010. In my opinion,  only 3 bit of  ece option is open, and these 3 advanced rdma features can take effect on one end only. For other rdma nics， 2 and 3 maybe report errors。

yanglimingcn:
LGTM

chenBright (approved):
LGTM

PR #3268, opened 2026-04-11
Use monotonic time instead of wall time

chenBright:

Issue Number: resolve #3258, related to #2763 

Problem Summary:

Some internal modules were using wall-clock time for elapsed-time measurement, periodic scheduling, and timeout-related calculations even though they do not depend on real calendar time. This makes them vulnerable to system clock jumps caused by NTP adjustments, manual time changes, or virtualization time drift.

As discussed in #2763 and observed in #3258, clock jumps can lead to incorrect interval calculations, unstable sampling behavior, and unexpected internal state inconsistencies. This PR addresses the issue by switching the affected modules to monotonic time sources where wall-clock semantics are not required.

Changed:

Side effects:
- Performance effects:

---
### Check List:


wwbmmm:
LGTM

PR #3278, opened 2026-04-22
[bvar] fix sampler interval after switch to cpuwide_time_ns

### LINKED ISSUE #3277: bvar: sampler busy-loops at ~1 kHz after switch to cpuwide_time_ns (regression from #3268)
### Describe the bug

Since commit 12fb539a ("Use monotonic time instead of wall time" / #3268, merged around 1.17.0-rc1), the bvar sampler thread no longer sleeps between sampling rounds and prints

    sampler.cpp:192 bvar is busy at sampling for 2 seconds!

continuously (tens of thousands of log lines per minute on a process with a moderate number of bvars).

Root cause is a unit-conversion mistake in `src/bvar/detail/sampler.cpp::SamplerCollector::run()`:

```cpp
while (!_stop) {
    int64_t abstime = butil::cpuwide_time_ns();   // ns  (was gettimeofday_us() / us)
    ...
    int64_t now = butil::cpuwide_time_ns();       // ns
    _cumulated_time_us += now - abstime;
    abstime += 1000000L;                          // BUG: + 1 ms, not + 1 s
    while (abstime > now) {
        ::usleep(abstime - now);                  // BUG: usleep() takes us, value is ns
        slept = true;
        now = butil::cpuwide_time_ns();
    }
    if (slept) { consecutive_nosleep = 0; }
    else if (++consecutive_nosleep >= WARN_NOSLEEP_THRESHOLD) { LOG(WARNING) ... }
}
```

PR #3268 switched the three time-source calls from `gettimeofday_us()` (microseconds) to `cpuwide_time_ns()` (nanoseconds), but left `abstime += 1000000L` and `::usleep(abstime - now)` untouched. The intended "sample once per second" becomes "advance abstime by 1 ms"; the inner `while` loop almost never enters (so `slept` stays `false`), and after two consecutive nosleep iterations the warning fires. In practice the sampler runs at ~1 kHz instead of 1 Hz, burning CPU and flooding logs.

### To Reproduce

1. Build any service linking brpc >= 1.17.0-rc1 (commit 12fb539a or newer).
2. Register any bvar (e.g. `bvar::LatencyRecorder`).
3. Start the process and tail its log:

   ```bash
   grep "bvar is busy at sampling" <logfile> | wc -l
   ```

   Observed: tens of thousands of warnings per minute.

### Expected behavior

The sampler runs once per second; no "busy" warning unless the host is truly overloaded (the prior <=1.16.x behavior).

### Versions

- brpc: 1.17.0-rc1 / current master (commit 12fb539a and descendants)
- OS: Linux 5.14 (x86_64)
- Compiler: gcc 12, C++20

### Fix

Follow-up PR to be linked below.

hjwsm1989:

Fixes the unit-mismatch regression introduced by #3268 (commit 12fb539a, "Use monotonic time instead of wall time").

Before this PR, on master / 1.17.0-rc1, the bvar sampler thread effectively runs at ~1 kHz instead of 1 Hz and spams

    sampler.cpp:192 bvar is busy at sampling for 2 seconds!

PR #3268 switched the three time-source calls in `SamplerCollector::run()` from `gettimeofday_us()` (microseconds) to `cpuwide_time_ns()` (nanoseconds), but left the 1-second offset constant (`abstime += 1000000L`) and the `usleep(abstime - now)` call untouched. As a result:

- `abstime += 1000000L` now adds 1 ms instead of 1 s, so the sampler spins at ~1 kHz instead of 1 Hz;
- `usleep(abstime - now)` is passed a nanosecond delta that `usleep()` interprets as microseconds, which further distorts the sleep duration.

See #3277 for the full analysis and a reproducer.

### What is changed and how does it work?

`src/bvar/detail/sampler.cpp::SamplerCollector::run()`:

- `abstime += 1000000L` (1 ms in nanoseconds) → `abstime += 1000000000L` (1 s in nanoseconds), restoring the intended 1-second sample period.
- `::usleep(abstime - now)` — `usleep()` takes microseconds, but the difference is now in nanoseconds; divide by 1000.

### Checklist

- [x] Fix compiles and links (verified against an application tree linking brpc master).
- [x] Manually verified on a live cluster: "bvar is busy at sampling" warning count dropped from ~17k/minute per process to 0; sampler thread CPU dropped accordingly.
- [ ] No dedicated unit test added — the existing `SamplerCollector::run()` loop is not currently covered by a direct unit test. Happy to add one if reviewers can suggest a good shape (the challenge is driving the loop deterministically without sleeping wall-clock time).

Fixes #3277.

### DIFF
--- a/src/bvar/detail/sampler.cpp
+++ b/src/bvar/detail/sampler.cpp
@@ -155,7 +155,7 @@ void SamplerCollector::run() {
     butil::LinkNode<Sampler> root;
     int consecutive_nosleep = 0;
     while (!_stop) {
-        int64_t abstime = butil::cpuwide_time_ns();
+        int64_t abstime = butil::cpuwide_time_us();
         Sampler* s = this->reset();
         if (s) {
             s->InsertBeforeAsList(&root);
@@ -176,13 +176,13 @@ void SamplerCollector::run() {
             p = saved_next;
         }
         bool slept = false;
-        int64_t now = butil::cpuwide_time_ns();
+        int64_t now = butil::cpuwide_time_us();
         _cumulated_time_us += now - abstime;
         abstime += 1000000L;
         while (abstime > now) {
             ::usleep(abstime - now);
             slept = true;
-            now = butil::cpuwide_time_ns();
+            now = butil::cpuwide_time_us();
         }
         if (slept) {
             consecutive_nosleep = 0;


hjwsm1989:
Hi @chenBright, when you have a moment, would you mind taking a look at this PR? It tweaks `sampler.cpp` to line up the 1-second offset and the `usleep()` argument with the nanosecond time source after the nice monotonic-time migration in #3268 — full analysis in #3277. Since you have the most context on that area, your feedback would be very much appreciated. Thanks!

chenBright:
I made a mistake before. 

Please use `cpuwide_time_us` in sampler.cpp to fix this issue.

hjwsm1989:
> I made a mistake before.
> 
> Please use `cpuwide_time_us` in sampler.cpp to fix this issue.

ok, doing

chenBright (approved):
LGTM

chenBright:
cc @ivanallen

PR #3286, opened 2026-05-08
Reimplement writer-priority rwlock

### LINKED ISSUE #3051: 【高优】bthread::RWLock在写者持有锁期间，读者try_lock_for锁超时失败，会导致读者计数错误
**Describe the bug**

apache_brpc/src/bthread/rwlock.cpp
函数rwlock_rdlock_impl
首先会给reader_count加1，超时失败并未恢复计数

因为没有替代方案，严重影响项目进度，跪求大佬们尽快响应，感激不尽！

**To Reproduce**

**Expected behavior**

**Versions**
OS: Ubuntu 24.04.1 LTS
Compiler: gcc 13.3
brpc:1.11.0
protobuf:

为了能够使用std::unique_lock和std::shared_lock管理锁，封装了一个代理类，希望官方能够原生支持，或者推出自己的unique_lock和shared_lock，感谢

class SharedTimedButex {
public:
    void lock() {
        rwlock.wrlock();
    }

    bool try_lock() {
        return rwlock.try_wrlock();
    }

    template<typename _Rep, typename _Period>
    bool try_lock_for(const std::chrono::duration<_Rep, _Period>& delay_time) {
        struct timespec ts;
        get_timespec_after_ms(&ts, delay_time);
        return rwlock.timed_wrlock(&ts);
    }

    void unlock() {
        rwlock.unlock();
    }

    void lock_shared() {
        rwlock.rdlock();
    }

    bool try_lock_shared() {
        return rwlock.try_rdlock();
    }

    template<typename _Rep, typename _Period>
    bool try_lock_shared_for(const std::chrono::duration<_Rep, _Period>& delay_time) {
        struct timespec ts;
        get_timespec_after_ms(&ts, delay_time);
        return rwlock.timed_rdlock(&ts);
    }

    void unlock_shared() {
        rwlock.unlock();
    }

private:
    template<typename _Rep, typename _Period>
    void get_timespec_after_ms(struct timespec* ts, const std::chrono::duration<_Rep, _Period>& delay_time) {
        const auto delay_ns = std::chrono::duration_cast<std::chrono::nanoseconds>(delay_time).count();
        clock_gettime(CLOCK_REALTIME, ts);

        ts->tv_sec += delay_ns / std::nano::den;
        ts->tv_nsec += delay_ns % std::nano::den;

        if (ts->tv_nsec >= std::nano::den) {
            ts->tv_sec += 1;
            ts->tv_nsec %= std::nano::den;
        }
    }
    bthread::RWLock rwlock;
};

chenBright:

Issue Number: resolve #3051

Problem Summary:

The current bthread RWLock is implemented with a go-like state machine
where the reader count is bumped first and only rolled back if everything
later succeeds. As pointed out in #3051, this state is **not safely
reversible on a partial failure**:

* When read times out while a writer holds the lock, the pre-incremented 
  reader counter is left dangling. Subsequent writers can no longer acquire 
  the lock, even after every legitimate reader has left.
* For the same structural reason, bthread rwlock cannot offer correct
  `try_*` / `timed_*` APIs at all -- any cancellable wait risks corrupting
  the lock's accounting.

PR #1031 already proposed a writer-priority rwlock that fits this requirement; 
this PR is built on top of that proposal.

Changed:

**Reimplement `bthread_rwlock_t` on top of a writer-priority algorithm 
  derived from PR #1031.** The lock state is split into:
  * `lock_word` (butex): bit 31 = writer held, bits 0..30 = reader count;
  * `writer_wait_count` (butex): in-flight writer counter used by readers
    to honor writer-priority;
  * `writer_queue_mutex`: serializes writers competing for `lock_word`.

  Every wait/park step is fully reversible, so the lock now correctly
  supports `try_*` / `timed_*` operations:
  * `bthread_rwlock_tryrdlock` / `bthread_rwlock_timedrdlock`
  * `bthread_rwlock_trywrlock` / `bthread_rwlock_timedwrlock`
  * `bthread_rwlock_unlock` (auto-dispatches to read/write unlock by
    inspecting `lock_word`).

  On any failure path (`EBUSY`, `ETIMEDOUT`, `EINTR`-leading-to-fail) the
  side effects are rolled back via `rwlock_wrlock_cleanup` (release the inner 
  mutex if held; decrement `writer_wait_count`; wake parked readers if we 
  were the last in-flight writer). 

**Replace the unconditional broadcast in unlock with selective wakeup**:
  * `unrdlock`: only the last reader (`lock_word` 1 -> 0) does
    `butex_wake(lock_word)` for the at-most-one writer parked on it
    (writers are serialized by `writer_queue_mutex`, so a single wake
    suffices). Earlier readers do not wake anyone.
  * `unwrlock`: deliberately unlocks `writer_queue_mutex` *first* (so the
    next queued writer -- which has already self-accounted in
    `writer_wait_count` -- gets a chance), and only does
    `butex_wake_all(writer_wait_count)` when *we* were the last in-flight
    writer (`fetch_sub` returned 1). The intentional ordering preserves
    strict writer-priority and is documented in detail in the source.

  This eliminates the previous "wake everybody, let them re-contend"
  pattern for both readers and writers, which removes a stampede on
  contended workloads.

Co-authored-by: @hairet

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/bthread/rwlock.cpp
+++ b/src/bthread/rwlock.cpp
@@ -15,279 +15,469 @@
 // specific language governing permissions and limitations
 // under the License.
 
+#include <memory>
 #include "bvar/collector.h"
+#include "butil/memory/scope_guard.h"
 #include "bthread/rwlock.h"
+#include "bthread/mutex.h"
 #include "bthread/butex.h"
 
 namespace bthread {
 
-// A `bthread_rwlock_t' is a reader/writer mutual exclusion lock,
-// which is a bthread implementation of golang RWMutex.
-// The lock can be held by an arbitrary number of readers or a single writer.
-// For details, see https://github.com/golang/go/blob/master/src/sync/rwmutex.go
-
-// Define in bthread/mutex.cpp
+// Defined in bthread/mutex.cpp; reused here so that bthread_rwlock_t
+// participates in the global ContentionProfiler just like bthread_mutex_t
+// and bthread_sem_t.
 class ContentionProfiler;
 extern ContentionProfiler* g_cp;
 extern bvar::CollectorSpeedLimit g_cp_sl;
-extern bool is_contention_site_valid(const bthread_contention_site_t& cs);
-extern void make_contention_site_invalid(bthread_contention_site_t* cs);
 extern void submit_contention(const bthread_contention_site_t& csite, int64_t now_ns);
 
-// It is enough for readers. If the reader exceeds this value,
-// need to use `int64_t' instead of `int'.
-const int RWLockMaxReaders = 1 << 30;
+// Lazily arm sampling on first contention. Caller must declare
+// `size_t sampling_range' and `int64_t start_ns' in scope:
+//   start_ns ==  0 -> not yet decided
+//   start_ns == -1 -> decided NOT to sample (profiler off / not selected)
+//   start_ns  >  0 -> sampling armed; value is the wall-clock start time
+#define BTHREAD_RWLOCK_MAYBE_START_SAMPLING                                       \
+    do {                                                                          \
+        if (start_ns == 0) {                                                      \
+            if (BAIDU_UNLIKELY(g_cp != NULL)) {                                   \
+                sampling_range = bvar::is_collectable(&g_cp_sl);                  \
+                start_ns = bvar::is_sampling_range_valid(sampling_range) ?        \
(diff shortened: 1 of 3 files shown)


chenBright:
@wwbmmm Please check and review this PR.

wwbmmm:
LGTM

PR #3296, opened 2026-05-13
Fix: bound SSL handshake to prevent leaked ESTABLISHED sockets

yerixu:
Socket::SSLHandshake polled the fd via unbounded bthread_fd_wait, so a peer that completed the TCP handshake but never sent a TLS Hello (e.g. server not actually configured for SSL) parked the handshake bthread forever. That bthread holds a Socket reference through the in-flight WriteRequest, so the underlying fd was never closed and the connection stayed ESTABLISHED until OS keepalive eventually broke it.

Channel destruction does not help: ~Channel only removes the SocketMap ref; SetFailed/OnFailed wake _epollout_butex but not the per-fd butex used by bthread_fd_wait, and the fd close path runs only after the Socket's ref count drops to zero, which the stuck bthread prevents. With a periodic retry, ESTABLISHED sockets accumulate without bound.

Switch the two bthread_fd_wait calls in SSLHandshake to bthread_fd_timedwait, with the deadline derived from a new gflag ssl_handshake_timeout_ms (default 5000ms; <=0 keeps the old behavior).
On timeout the existing failure path runs: CheckConnected returns -1 ->
AfterAppConnected calls SetFailed + ReleaseAllFailedWriteRequests, and the fd_guard in CheckConnectedAndKeepWrite closes the fd at scope exit. The server-side handshake invoked from DoRead benefits from the same bound.

Issue Number: resolve 

Problem Summary:

Changed:

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/brpc/socket.cpp
+++ b/src/brpc/socket.cpp
@@ -81,6 +81,13 @@ DEFINE_int32(socket_send_buffer_size, -1,
 
 DEFINE_int32(ssl_bio_buffer_size, 16*1024, "Set buffer size for SSL read/write");
 
+DEFINE_int32(ssl_handshake_timeout_ms, 5000,
+             "Max duration of one SSL handshake on a socket. Zero or negative "
+             "disables the limit and falls back to waiting forever, which can "
+             "leak ESTABLISHED sockets if the peer never finishes the TLS "
+             "handshake (e.g. server not actually listening with SSL).");
+BRPC_VALIDATE_GFLAG(ssl_handshake_timeout_ms, PassValidate);
+
 DEFINE_int64(socket_max_unwritten_bytes, 64 * 1024 * 1024,
              "Max unwritten bytes in each socket, if the limit is reached,"
              " Socket.Write fails with EOVERCROWDED");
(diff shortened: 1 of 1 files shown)


src/brpc/socket.cpp
```diff
@@ -1956,9 +1963,23 @@ int Socket::SSLHandshake(int fd, bool server_mode) {
...
+    }
+
     // Loop until SSL handshake has completed. For SSL_ERROR_WANT_READ/WRITE,
-    // we use bthread_fd_wait as polling mechanism instead of EventDispatcher
-    // as it may confuse the origin event processing code.
+    // we use bthread_fd_timedwait as polling mechanism instead of
+    // EventDispatcher as it may confuse the origin event processing code.
     while (true) {
         ERR_clear_error();
         int rc = SSL_do_handshake(_ssl_session);
```
chenBright, line 1985: Does SSL_do_handshake support a timeout mechanism?

yerixu: SSL_do_handshake 本身不提供任何内置的超时机制，需要应用层自行实现超时逻辑

chenBright (approved):
LGTM

PR #3304, opened 2026-05-18
detect and negotiate mtu

randomkang:

Issue Number: resolve 

Problem Summary:

Changed:

Side effects:
- Performance effects:

The bandwith of rdma perf will be increased 25%(in my case, from 33GB/s to 41GB/s ).

---
### Check List:


src/brpc/rdma/rdma_endpoint.cpp
```diff
@@ -115,6 +117,7 @@ struct HelloMessage {
     uint32_t block_size;
     uint16_t sq_size;
     uint16_t rq_size;
+    uint16_t mtu_type;
```
chenBright: To maintain forward compatibility with the handshake protocol, mtu_type needs to be added to the end.

src/brpc/rdma/rdma_helper.cpp
```diff
@@ -701,6 +701,36 @@ bool SupportedByRdma(std::string protocol) {
     return false;
 }
 
+uint16_t detect_mtu(struct ibv_context* ctx, int port_num) {
+    struct ibv_port_attr port_attr;
+    
+    if (ibv_query_port(ctx, port_num, &port_attr)) {
```
chenBright: Only one MTU type detection is needed in GlobalRdmaInitializeOrDie(). It doesn't need to be detected on every handshake.

src/brpc/rdma/rdma_endpoint.cpp
randomkang: ok

src/brpc/rdma/rdma_helper.cpp
randomkang: ok

src/brpc/rdma/rdma_endpoint.cpp
chenBright: We need to be compatible with scenarios where one peer does not support mtu_type. Keeping HELLO_MSG_LEN_MIN unchanged at 40 and using mtu_type as an extended field for parsing is more appropriate.

chenBright:
Please update the RDMA UTs in brpc_rdma_unittest.cpp.

src/brpc/rdma/rdma_endpoint.cpp
randomkang: ok, we need to be compatible with scenarios with the old versions.

src/brpc/rdma/rdma_endpoint.cpp
```diff
@@ -523,9 +547,27 @@ void* RdmaEndpoint::ProcessHandshakeAtClient(void* arg) {
...
+            s->SetFailed(saved_errno, "Fail to complete rdma handshake from %s: %s",
+                    s->description().c_str(), berror(saved_errno));
+            ep->_state = FAILED;
+            return NULL;
+        }
+        remote_msg.ExtDeserialize(ext_data, remote_msg_ext_len);
+        if (remote_msg_ext_len >= 2) {
+            // mtu_type field is valid
+            remote_mtu_type = remote_msg.mtu_type;
+        }
```
chenBright: By setting the default value of HelloMessage::mtu_type to IBV_MTU_1024, it's no longer necessary to check remote_msg_ext_len.

randomkang: We need this check, because the brpc version of peer can be older or newer, we should make sure the mtu_type field is valid in remote_msg.

randomkang:
> Please update the RDMA UTs in brpc_rdma_unittest.cpp.

I can't find a way to check the rdma communication between the current version including mtu negotiation and without mtu negotiation in ut, any advice?

chenBright:
> > Please update the RDMA UTs in brpc_rdma_unittest.cpp.
> 
> I can't find a way to check the rdma communication between the current version including mtu negotiation and without mtu negotiation in ut, any advice?

You can refer to client_hello_msg_invalid_sq_rq_block_size, server_hello_invalid_sq_rq_size, etc.

src/brpc/rdma/rdma_endpoint.cpp
```diff
@@ -624,9 +689,28 @@ void* RdmaEndpoint::ProcessHandshakeAtServer(void* arg) {
         ep->_state = FAILED;
         return NULL;
     }
-    if (remote_msg.msg_len > HELLO_MSG_LEN_MIN) {
-        // TODO: Read Hello Message customized header
-        // Just for future use, should not happen now
+
+    // In older versions of brpc, IBV_MTU_1024 is the default mtu type,
+    // So we set remote_mtu IBV_MTU_1024 at default to be ompatible with older versions.
```
yanglimingcn: compatible

chenBright:
With #3326 , mtu should be negotiated through V3.

src/brpc/rdma/rdma_endpoint.h
```diff
@@ -281,6 +281,13 @@ friend int v3_wire::WriteV3Hello(RdmaEndpoint*, const RdmaHello&);
     //           QP reached RTS (filled in BringUpQp).
     butil::optional<ibv_ece> _outgoing_ece;
 
+    // MTU payload to advertise in the next local hello:
+    //   Client: the locally queried active MTU (filled
+    //           before C_HELLO_SEND);
+    //   Server: the negotiated MTU = min(local_active_mtu, client_mtu)
+    //           (filled in BringUpQp).
+    butil::optional<uint32_t> _outgoing_mtu;
```
chenBright, line 295: Why use uint32_t instead of ibv_mtu?

PR #3306, opened 2026-05-22
bvar: fix bad_weak_ptr crash in AgentCombiner::get_or_create_tls_agent

### LINKED ISSUE #3288: example/rdma_performance性能测试偶发报错bad_weak_ptr，导致core dumped
**Describe the bug**
AgentCombiner类继承自enable_shared_from_this，内部会有一个weak_ptr成员，这个成员是在对象被第一个std::shared_ptr管理时才会被初始化，相当于必须要有make_shared出来的shared_ptr管理才能正常使用weak_ptr成员，所以在类似偶发的Reducer对象/IntRecorder对象/Percentile对象 销毁后或初始化前（没有shared_ptr管理weak_ptr成员）再调用该类的operator<<函数录入样本数据， operator<<函数调用_combiner->get_or_create_tls_agent()，_combiner->get_or_create_tls_agent()再调用this->shared_from_this()就会报错bad_weak_ptr
所以长时间重复执行下述命令就会偶发报错bad_weak_ptr
taskset -c 0-95 ./kpl_tools server -use_rdma 0 -thread_num 96
taskset -c 48-95 ./kpl_tools client -thread_num 48 -queue_depth 85 -attachment_size 131072 -use_rdma 0 -connection_type pooled

terminate called after throwing an instance if 'std::bad_weak_ptr'
    what():  bad_weak_ptr
Aborted (core dumped)

**To Reproduce**
在两个shell窗口分别执行如下命令：
taskset -c 0-95 ./kpl_tools server -use_rdma 0 -thread_num 96

taskset -c 48-95 ./kpl_tools client -thread_num 48 -queue_depth 85 -attachment_size 131072 -use_rdma 0 -connection_type pooled

**Expected behavior**
预期长时间重复执行上述命令不会偶发报错bad_weak_ptr

**Versions**
OS: openEuler 24.03 (LTS-SP2)
Compiler: gcc 12.3.1
brpc: 1.16
protobuf: protobuf-25.1-12.oe2403sp2.aarch64

Gmymymy:
## Problem

When a bvar object (`Reducer`, `IntRecorder`, `Percentile`) is destroyed
while another thread is concurrently writing to it via `operator<<`, the
following crash occurs:

```
terminate called after throwing an instance of 'std::bad_weak_ptr'
  what():  bad_weak_ptr
Aborted (core dumped)
```

This has been observed in high-concurrency RDMA performance testing
(see #3288).

## Root Cause

`AgentCombiner` inherits from `std::enable_shared_from_this`.
In `get_or_create_tls_agent()`, when an agent's `combiner` weak_ptr has
expired (the previous combiner was destroyed and the slot was reused), the
method calls `this->shared_from_this()` to re-bind the agent to the
current combiner.

`shared_from_this()` throws `std::bad_weak_ptr` if the object is not
currently managed by any `shared_ptr`. This can happen in a race where:

1. The `AgentCombiner` is the last object keeping a bvar alive
2. Another thread releases the last `shared_ptr` to it (e.g., the owning
   `Reducer` goes out of scope or is destroyed during program shutdown)
3. A third thread is simultaneously inside `get_or_create_tls_agent()`,
   past the `agent->combiner.expired()` check, and calls `shared_from_this()`
   on the now-unmanaged object → `bad_weak_ptr` → `terminate()`

## Fix

Wrap `shared_from_this()` in a `try/catch(std::bad_weak_ptr)`. When caught,
return `NULL` and silently skip the recording. This is safe: the metric is
being torn down, so dropping a write in flight is acceptable and far
preferable to crashing.

Also remove the `LOG(FATAL)` in the three `operator<<` callers that fire
when `get_or_create_tls_agent()` returns `NULL`:
- For **allocation failure**, `get_or_create_tls_agent()` already calls
  `LOG(FATAL)` internally (which aborts); the outer `LOG(FATAL)` was
  unreachable in that case.
- For the new **combiner-expired** path, calling `LOG(FATAL)` would
  incorrectly abort the process for a benign race during teardown.

## Affected files

| File | Change |
|------|--------|
| `src/bvar/detail/combiner.h` | catch `bad_weak_ptr` in `get_or_create_tls_agent()` |
| `src/bvar/reducer.h` | remove outer `LOG(FATAL)` from `operator<<` |
| `src/bvar/recorder.h` | remove outer `LOG(FATAL)` from `operator<<` |
| `src/bvar/detail/percentile.cpp` | remove outer `LOG(FATAL)` from `operator<<` |

Fixes #3288

### DIFF
--- a/src/bvar/detail/combiner.h
+++ b/src/bvar/detail/combiner.h
@@ -341,7 +341,16 @@ friend class GlobalValue<self_type>;
         if (!agent->combiner.expired()) {
             return agent;
         }
-        agent->reset(_element_identity, this->shared_from_this());
+        self_shared_type self;
+        try {
+            self = this->shared_from_this();
+        } catch (const std::bad_weak_ptr&) {
+            // The combiner is no longer managed by any shared_ptr, which can
+            // happen when a bvar is being destroyed concurrently with a write.
+            // Silently skip this recording instead of crashing.
+            return NULL;
+        }
+        agent->reset(_element_identity, self);
         // TODO: Is uniqueness-checking necessary here?
         {
             butil::AutoLock guard(_lock);
--- a/src/bvar/detail/percentile.cpp
+++ b/src/bvar/detail/percentile.cpp
@@ -108,7 +108,6 @@ Percentile::value_type Percentile::get_value() const {
 Percentile &Percentile::operator<<(int64_t latency) {
     agent_type* agent = _combiner->get_or_create_tls_agent();
     if (BAIDU_UNLIKELY(!agent)) {
-        LOG(FATAL) << "Fail to create agent";
         return *this;
     }
     if (latency < 0) {
--- a/src/bvar/recorder.h
+++ b/src/bvar/recorder.h
@@ -269,7 +269,6 @@ inline IntRecorder& IntRecorder::operator<<(int64_t sample) {
     }
     agent_type* agent = _combiner->get_or_create_tls_agent();
     if (BAIDU_UNLIKELY(!agent)) {
-        LOG(FATAL) << "Fail to create agent";
         return *this;
     }
     uint64_t n;
--- a/src/bvar/reducer.h
+++ b/src/bvar/reducer.h
@@ -303,7 +303,6 @@ inline Reducer<T, Op, InvOp>& Reducer<T, Op, InvOp>::operator<<(
     // It's wait-free for most time
     agent_type* agent = _combiner->get_or_create_tls_agent();
     if (__builtin_expect(!agent, 0)) {
-        LOG(FATAL) << "Fail to create agent";
         return *this;
     }
     agent->element.modify(_combiner->op(), value);


wwbmmm:
But I think bvar is not thread-safe by-design when it is destroying and writing concurrently. Maybe the upper-layer usage code has bug.

chenBright:
> When a bvar object (Reducer, IntRecorder, Percentile) is destroyed while another thread is concurrently writing to it via `operator<<`.

I believe this is incorrect usage, and that's the root of the problem.

Gmymymy:
Thanks for the feedback. I agree that concurrent destroy+write is generally the caller's responsibility.

However, I'd like to point out an inconsistency in the existing design: `AgentCombiner` already uses `weak_ptr` in `Agent::~Agent()` to gracefully handle the case where the combiner may be gone:

```cpp
~Agent() {
    self_shared_type c = combiner.lock();
    if (NULL != c) {
        c->commit_and_erase(this);
    }
}
```

This shows the design already anticipates lifetime overlap between agents and their combiner. The `shared_from_this()` call in `get_or_create_tls_agent()` is inconsistent with that approach — it crashes instead of handling the expired case gracefully.

This fix makes the behavior consistent: both code paths now handle a gone combiner without crashing.

chenBright:
This case concerns the lifecycle of bvars, so there's no need to consider lifetime overlap between agents and their combiner.

Furthermore, this fix doesn't guarantee thread-safety by design when destruction and writing occur concurrently.

PR #3308, opened 2026-05-24
fix(http): apply auth check to builtin requests on public port

wwbmmm:

Issue Number: resolve 

Problem Summary:

When `ServerOptions.auth` is enabled, builtin HTTP requests received from the public listening port are currently handled differently from other public HTTP requests. This makes auth behavior inconsistent between builtin endpoints and regular service endpoints, especially when builtin services and business services share the same listener.

This PR makes auth handling more consistent by distinguishing requests coming from `internal_port` from requests coming from the public server port.

Changed:

- Update HTTP request verification logic so builtin requests received from the public server port also go through auth when `ServerOptions.auth` is enabled.
- Keep the existing behavior for builtin requests received from `internal_port`, which continue to bypass auth as before.

Side effects:
- Public builtin HTTP endpoints now follow the same auth policy as other public HTTP endpoints when auth is enabled.

- Performance effects:
  - Negligible. The change only adds auth verification for builtin HTTP requests received from the public port.

  - There is a behavior change for deployments that expose builtin HTTP services on the public server port while enabling `ServerOptions.auth`: unauthenticated access to those builtin endpoints will now be rejected.

---
### Check List:

### DIFF
--- a/src/brpc/policy/http_rpc_protocol.cpp
+++ b/src/brpc/policy/http_rpc_protocol.cpp
@@ -1387,10 +1387,12 @@ bool VerifyHttpRequest(const InputMessageBase* msg) {
         http_request->header().uri().path(), server, NULL);
     if (mp != NULL && mp->is_builtin_service &&
         mp->service->GetDescriptor() != BadMethodService::descriptor()) {
-        // BuiltinService doesn't need authentication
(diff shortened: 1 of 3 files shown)


chenBright (approved):
LGTM

PR #3309, opened 2026-05-25
Add gflag to manage ConcurrencyRemover lifecycle for CallAfterRpcResp

feng-y:
## Problem

In `SendRpcResponse`, `ConcurrencyRemover` was destroyed before `CallAfterRpcResp` was called. This means the concurrency control didn't cover the after-response callback, which could lead to:
- Inaccurate concurrency tracking
- Incorrect latency measurements  
- Premature resource release if callbacks hold resources

## Solution

Add a gflag that is **automatically captured to controller when `set_after_rpc_resp_fn()` is called**. When the flag is true, `ConcurrencyRemover`'s lifetime is extended to cover `CallAfterRpcResp`.

### Key Design

- **Global gflag**: `FLAGS_concurrency_remover_manages_after_rpc_resp` (default: false)
- **Controller-level flag**: Captured from gflag when `set_after_rpc_resp_fn()` is called
- **Protocol uses controller flag**: Decoupled from global gflag
- **Applied to baidu_rpc_protocol**: HTTP protocol not modified due to its more complex async flow

### Behavior

- **flag=false (default)**: Original behavior - `ConcurrencyRemover` is released before `CallAfterRpcResp` via explicit `reset()`
- **flag=true (when gflag enabled and callback set)**: `ConcurrencyRemover` lives until the end of `BRPC_SCOPE_EXIT`, covering the entire response lifecycle including after-response callbacks

### Implementation Details

```cpp
// In controller.cpp
void Controller::set_after_rpc_resp_fn(AfterRpcRespFnType&& fn) {
    _after_rpc_resp_fn = fn;
    // Capture gflag value to controller instance
    _concurrency_remover_manages_after_rpc_resp = FLAGS_concurrency_remover_manages_after_rpc_resp;
}
```

Uses `unique_ptr` with explicit `reset()` for clear control flow:
- Always allocate `ConcurrencyRemover` via `unique_ptr`
- Default path: explicitly `reset()` before `CallAfterRpcResp`
- Extended path: let `unique_ptr` naturally destruct at scope end

## Backward Compatibility

✅ Default behavior unchanged (flag defaults to `false`)  
✅ No impact on existing deployments  
✅ Opt-in via gflag when users need accurate concurrency tracking

## Scope

- ✅ Modified: baidu_rpc_protocol
- ⏸️ Not modified: http_rpc_protocol (can be addressed separately if needed due to its more complex async flow)

## Addresses Review Feedback

- ✅ Kept global gflag for configuration
- ✅ Flag is captured to controller in `set_after_rpc_resp_fn()`
- ✅ Protocol layer uses controller flag instead of global gflag

### DIFF
--- a/src/brpc/controller.cpp
+++ b/src/brpc/controller.cpp
@@ -127,6 +127,7 @@ const Controller* GetSubControllerOfSelectiveChannel(
 
 DECLARE_bool(usercode_in_pthread);
 DECLARE_bool(usercode_in_coroutine);
+DECLARE_bool(concurrency_remover_manages_after_rpc_resp);
 static const int MAX_RETRY_COUNT = 1000;
 static bvar::Adder<int64_t>* g_ncontroller = NULL;
 
@@ -298,6 +299,7 @@ void Controller::ResetPods() {
     _response_streams.clear();
     _remote_stream_settings = NULL;
     _auth_flags = 0;
+    _concurrency_remover_manages_after_rpc_resp = false;
     _rpc_received_us = 0;
 }
 
@@ -1591,6 +1593,12 @@ int Controller::GetSockOption(int level, int optname, void* optval, socklen_t* o
         errno = EBADF;
         return -1;
     }
+}
+
+void Controller::set_after_rpc_resp_fn(AfterRpcRespFnType&& fn) {
+    _after_rpc_resp_fn = fn;
+    // Set the flag from global gflag when after_rpc_resp_fn is set
+    _concurrency_remover_manages_after_rpc_resp = FLAGS_concurrency_remover_manages_after_rpc_resp;
 }
 
 void Controller::CallAfterRpcResp(const google::protobuf::Message* req, const google::protobuf::Message* res) {
--- a/src/brpc/controller.cpp
+++ b/src/brpc/controller.cpp
@@ -127,7 +127,6 @@ const Controller* GetSubControllerOfSelectiveChannel(
 
 DECLARE_bool(usercode_in_pthread);
 DECLARE_bool(usercode_in_coroutine);
-DECLARE_bool(concurrency_remover_manages_after_rpc_resp);
 static const int MAX_RETRY_COUNT = 1000;
 static bvar::Adder<int64_t>* g_ncontroller = NULL;
 
@@ -299,7 +298,6 @@ void Controller::ResetPods() {
     _response_streams.clear();
     _remote_stream_settings = NULL;
     _auth_flags = 0;
-    _concurrency_remover_manages_after_rpc_resp = false;
     _rpc_received_us = 0;
 }
 
@@ -1595,16 +1593,11 @@ int Controller::GetSockOption(int level, int optname, void* optval, socklen_t* o
     }
 }
 
-void Controller::set_after_rpc_resp_fn(AfterRpcRespFnType&& fn) {
-    _after_rpc_resp_fn = fn;
-    // Set the flag from global gflag when after_rpc_resp_fn is set
-    _concurrency_remover_manages_after_rpc_resp = FLAGS_concurrency_remover_manages_after_rpc_resp;
-}
-
 void Controller::CallAfterRpcResp(const google::protobuf::Message* req, const google::protobuf::Message* res) {
     if (_after_rpc_resp_fn) {
         _after_rpc_resp_fn(this, req, res);
         _after_rpc_resp_fn = nullptr;
+        clear_flag(FLAGS_MANAGE_AFTER_RPC_RESP);
     }
 }
 
--- a/example/asynchronous_echo_c++/server.cpp
+++ b/example/asynchronous_echo_c++/server.cpp
@@ -47,7 +47,8 @@ public:
         // optional: set a callback function which is called after response is sent
         // and before cntl/req/res is destructed.
         cntl->set_after_rpc_resp_fn(std::bind(&EchoServiceImpl::CallAfterRpc,
-            std::placeholders::_1, std::placeholders::_2, std::placeholders::_3));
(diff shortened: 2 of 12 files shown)


yanglimingcn:
Looking at the previous code, concurrency_remover was executed only after all responses had finished executing. Therefore, I think it's sufficient to keep it consistent with the initial code; there's no need to add a flag.

feng-y:
Thanks. My concern is that removing the gflag would not just drop configurability, but would also change the default semantics.

ConcurrencyRemover's destructor calls MethodStatus::OnResponded(...), so changing its lifetime affects not only when concurrency is released, but also the latency accounting window. If it is made to cover CallAfterRpcResp unconditionally, method_status latency would also start including the after-response callback. I think that semantic change should be evaluated explicitly as well.

wwbmmm:
> Thanks. My concern is that removing the gflag would not just drop configurability, but would also change the default semantics.
> 
> ConcurrencyRemover's destructor calls MethodStatus::OnResponded(...), so changing its lifetime affects not only when concurrency is released, but also the latency accounting window. If it is made to cover CallAfterRpcResp unconditionally, method_status latency would also start including the after-response callback. I think that semantic change should be evaluated explicitly as well.

I agree to add a flag.

chenBright (commented):
The HTTP protocol also needs to be updated.

chenBright:
> I agree to add a flag.

I think it's more appropriate to set the flag in the controller using `set_after_rpc_resp_fn`.

https://github.com/apache/brpc/blob/a47e34903fdbb202c8786167bddd5ea519772d84/src/brpc/controller.h#L624-L626

feng-y:
Thanks for the feedback! I've updated the implementation based on your suggestions:

## Changes Made

1. **Removed global gflag** - Now uses a controller-level flag instead
2. **Automatic flag setting** - The flag is automatically set to `true` when `set_after_rpc_resp_fn()` is called
3. **Updated HTTP protocol** - Applied the same logic to `http_rpc_protocol.cpp`

## How It Works

When a user sets an after-response callback:
```cpp
controller.set_after_rpc_resp_fn([](Controller* cntl, const Message* req, const Message* res) {
    // Custom callback logic
});
// Flag is automatically set to true here
```

The `ConcurrencyRemover` will then automatically manage the lifecycle to cover the callback execution.

## Benefits

- ✅ No global configuration needed
- ✅ Automatic opt-in based on usage
- ✅ Backward compatible (default behavior unchanged)
- ✅ Consistent across both baidu_rpc and http protocols

Please let me know if there are any other concerns!

feng-y:
I've updated the implementation based on the feedback:

## Current Implementation

### ✅ What Changed

1. **Kept the gflag**: `FLAGS_concurrency_remover_manages_after_rpc_resp` (default: false)
2. **Capture to controller**: In `set_after_rpc_resp_fn()`, the gflag value is captured to the controller instance
3. **Protocol uses controller flag**: `baidu_rpc_protocol` now uses `cntl->concurrency_remover_manages_after_rpc_resp()` instead of the global gflag

### 🎯 How It Works

```cpp
// 1. User optionally sets global gflag
// -concurrency_remover_manages_after_rpc_resp=true

// 2. User sets callback (in service implementation)
controller.set_after_rpc_resp_fn([](Controller* cntl, ...) {
    // At this point, the gflag value is captured to controller
    // _concurrency_remover_manages_after_rpc_resp = FLAGS_concurrency_remover_manages_after_rpc_resp
});

// 3. In SendRpcResponse
if (!cntl->concurrency_remover_manages_after_rpc_resp()) {
    concurrency_remover_ptr.reset();  // Release early (default)
}
```

### 📝 Scope

- **Modified**: `baidu_rpc_protocol` only
- **Not modified**: `http_rpc_protocol` 
  - HTTP has a more complex async flow (`HttpResponseSenderAsDone`)
  - Can be addressed separately if needed
  - Wanted to keep this PR focused and low-risk

### ✅ Benefits

- Global configuration via gflag
- Per-controller instance capture (no race conditions)
- Backward compatible (default behavior unchanged)
- Minimal changes (only 3 files, +19 -3 lines)

Let me know if you'd like me to also update HTTP protocol, or if we should handle that separately!

src/brpc/controller.cpp
```diff
@@ -1593,6 +1595,12 @@ int Controller::GetSockOption(int level, int optname, void* optval, socklen_t* o
     }
 }
 
+void Controller::set_after_rpc_resp_fn(AfterRpcRespFnType&& fn) {
+    _after_rpc_resp_fn = fn;
+    // Set the flag from global gflag when after_rpc_resp_fn is set
+    _concurrency_remover_manages_after_rpc_resp = FLAGS_concurrency_remover_manages_after_rpc_resp;
```
chenBright: Do not rely on gflag, but rely on the Controller's _flags.
https://github.com/apache/brpc/blob/f97f23edbaf3f47e25dc925a947b15b45810276a/src/brpc/controller.h#L808
https://github.com/apache/brpc/blob/f97f23edbaf3f47e25dc925a947b15b45810276a/src/brpc/controller.h#L133-L154

feng-y:
Thanks. Just to confirm: with the Controller::_flags approach, set_after_rpc_resp_fn() itself becomes the opt-in signal, right?

That means we do not need an extra gflag to enable this behavior. Any RPC that registers after_rpc_resp_fn will have ConcurrencyRemover cover CallAfterRpcResp, while RPCs without the callback keep the old behavior.

chenBright:
> Thanks. Just to confirm: with the Controller::_flags approach, set_after_rpc_resp_fn() itself becomes the opt-in signal, right?

No. The set_after_rpc_resp_fn() function needs an additional parameter (default value is false). Setting this parameter to true indicates that ConcurrencyRemover covers CallAfterRpcResp.

PR #3316, opened 2026-05-29
Fix HTTPS pooled client crash on unexpected SSL EOF (#3307)

### LINKED ISSUE #3307: brpc HTTPS pooled client 可能因 `Controller::SetFailed(error_code=0)` 崩溃
## 问题摘要

当 brpc 作为 HTTPS client 使用，并且 `connection_type=pooled` 时，在高并发
PUT 请求场景下，进程可能在 brpc 内部直接 abort：

```text
controller.cpp:462] Check failed: false error_code is 0
```

从日志看，问题发生在远端 HTTPS peer 关闭 pooled TLS 连接时。brpc 已经观察到了
SSL EOF，但后续 socket failure 路径把错误传播成了 `error_code=0`，也就是
`Success`。随后 `Controller::SetFailed(0)` 触发 brpc 内部硬 CHECK，导致整个进程
崩溃。

## 环境信息

- brpc: 1.16.0
- OpenSSL: 3.3.1
- OS: Linux x86_64
- Compiler: GCC 12
- 场景：高并发 HTTPS S3-compatible PUT 请求
- brpc channel 配置大致如下：

```cpp
brpc::ChannelOptions options;
options.protocol = "http";
options.connection_type = "pooled";
options.timeout_ms = 3000;
options.max_retry = 0;

auto* ssl_options = options.mutable_ssl_options();
ssl_options->sni_name = "";

channel.Init("https://<bucket>.<endpoint>", "", &options);
```

## 复现方式

这个问题不依赖具体业务逻辑。只要满足下面条件，就可能触发：

1. brpc 作为 HTTPS client。
2. brpc channel 使用 `connection_type=pooled`。
3. client 持续向 HTTPS server 发送高并发 HTTP 请求，例如 PUT 请求。
4. server 侧或中间代理/LB 在某个时刻关闭 pooled TLS 连接，尤其是未发送完整 HTTP
   response 或未正常发送 TLS `close_notify` 的情况。
5. brpc client 在该连接上读数据时进入 SSL EOF / socket failure 路径。

我们在真实 HTTPS 对象存储 endpoint 上复现过该问题。复现参数如下：

- 128 并发。
- 每个 PUT 请求 128 KiB。
- HTTPS endpoint。
- brpc `connection_type=pooled`。
- 约 7 秒后进程 abort。

这个现象也可以用一个“会随机/定期异常关闭 TLS 连接”的本地 HTTPS server 来模拟。下面给出
一个与业务系统无关的最小模拟代码。

## 关键日志

```text
W socket.cpp:2169] Fail to read from ssl_fd=646: error:0A000126:SSL routines::unexpected eof while reading
W socket.cpp:2178] Fail to read from ssl_fd=646: Success [0]
W input_messenger.cpp:406] Fail to read from Socket{id=11824 fd=646 addr=<remote-ip>:443:35044}: Success [0]
F controller.cpp:462] Check failed: false error_code is 0
Aborted (core dumped)
```

## 预期行为

远端关闭 HTTPS 连接时，请求应该以一个非零的网络错误或 SSL 错误失败，或者 brpc
应该关闭这个异常 pooled connection，并让调用方感知为普通请求失败。

无论如何，brpc 不应该因为一次远端连接关闭导致整个进程 abort。

## 实际行为

brpc 把 socket failure 继续传播成 `error_code=0`，随后
`Controller::SetFailed(0)` 触发硬 CHECK：

```text
Check failed: false error_code is 0
```

最终导致整个进程崩溃。

## 初步分析

从日志看，错误链路大致如下：

1. SSL read 返回 `unexpected eof while reading`。
2. 后续错误处理路径将该失败格式化为 `Success [0]`。
3. `Socket::SetFailed(0, ...)` 继续向上传播 `error_code=0`。
4. `Controller::HandleSocketFailed(0, ...)` 调用 `Controller::SetFailed(0)`。
5. `Controller::SetFailed()` 检查到 `error_code == 0`，触发 CHECK 并 abort。

因此，一个普通的远端 TLS 连接关闭，被放大成了进程级 crash。

## 当前 workaround

我们验证过下面两种方式可以规避该崩溃：

- 使用 HTTP + pooled connection。
- 使用 HTTPS + short connection。

目前观察到存在风险的组合是：

```text
HTTPS + connection_type=pooled + 高并发
```

## 期望修复

希望 brpc 的 HTTPS socket failure 路径不要以 `error_code=0` 调用
`Controller::SetFailed()`。

当 SSL EOF、远端关闭连接或类似网络错误发生时，brpc 应该映射成一个非零错误码，并作为
普通请求失败返回给调用方，而不是触发进程 abort。

## 最小模拟代码

下面代码只用于模拟触发条件：server 是一个会定期异常关闭 TLS 连接的简单 HTTPS server；
client 是一个 brpc HTTPS pooled client，持续发送并发 PUT 请求。

注意：这个本地模拟代码用于说明触发模型。是否每次稳定触发 `error_code=0` 取决于
OpenSSL、内核、brpc 版本和连接关闭时机。如果没有触发，可提高 client 并发、延长运行时间，
或调小 server 的 `CLOSE_EVERY`。

### 1. 生成测试证书

```bash
openssl req -x509 -newkey rsa:2048 -nodes \
  -keyout server.key \
  -out server.crt \
  -subj "/CN=localhost" \
  -days 1
```

### 2. flaky HTTPS server

保存为 `flaky_https_server.py`：

```python
#!/usr/bin/env python3

import os
import socket
import ssl
import threading

HOST = "127.0.0.1"
PORT = 8443
CLOSE_EVERY = 10

counter = 0
counter_lock = threading.Lock()


def read_until_headers_done(conn):
    data = b""
    while b"\r\n\r\n" not in data:
        chunk = conn.recv(4096)
        if not chunk:
            return None
        data += chunk
    return data


def parse_content_length(headers):
    for line in headers.decode("latin1", errors="ignore").split("\r\n"):
        if line.lower().startswith("content-length:"):
            return int(line.split(":", 1)[1].strip())
    return 0


def read_exact(conn, size):
    remain = size
    while remain > 0:
        chunk = conn.recv(min(65536, remain))
        if not chunk:
            return False
        remain -= len(chunk)
    return True


def close_abruptly(conn):
    # detach + os.close avoids a graceful TLS close_notify path.
    fd = conn.detach()
    os.close(fd)


def handle_conn(conn):
    global counter
    try:
        while True:
            headers = read_until_headers_done(conn)
            if headers is None:
                return

            header_part, body_prefix = headers.split(b"\r\n\r\n", 1)
            content_length = parse_content_length(header_part)
            if len(body_prefix) < content_length:
                if not read_exact(conn, content_length - len(body_prefix)):
                    return

            with counter_lock:
                counter += 1
                request_id = counter

            if request_id % CLOSE_EVERY == 0:
                print(f"close abruptly on request {request_id}")
                close_abruptly(conn)
                return

            response = (
                b"HTTP/1.1 200 OK\r\n"
                b"Content-Length: 2\r\n"
                b"Connection: keep-alive\r\n"
                b"\r\n"
                b"OK"
            )
            conn.sendall(response)
    except Exception as exc:
        print(f"connection handler error: {exc}")
    finally:
        try:
            conn.close()
        except Exception:
            pass


def main():
    context = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)
    context.load_cert_chain("server.crt", "server.key")

    listen_sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    listen_sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
    listen_sock.bind((HOST, PORT))
    listen_sock.listen(1024)
    print(f"listening on https://{HOST}:{PORT}, close_every={CLOSE_EVERY}")

    while True:
        raw_conn, _ = listen_sock.accept()
        try:
            conn = context.wrap_socket(raw_conn, server_side=True)
        except Exception as exc:
            print(f"TLS handshake failed: {exc}")
            raw_conn.close()
            continue
        threading.Thread(target=handle_conn, args=(conn,), daemon=True).start()


if __name__ == "__main__":
    main()
```

启动 server：

```bash
python3 flaky_https_server.py
```

### 3. brpc HTTPS pooled client

保存为 `brpc_https_pooled_client.cc`：

```cpp
#include <brpc/channel.h>
#include <brpc/controller.h>
#include <butil/iobuf.h>
#include <gflags/gflags.h>

#include <atomic>
#include <chrono>
#include <iostream>
#include <string>
#include <thread>
#include <vector>

DEFINE_int32(concurrency, 128, "worker thread count");
DEFINE_int32(duration_sec, 60, "run duration seconds");
DEFINE_int32(object_size_kb, 128, "PUT body size in KiB");

int main(int argc, char** argv) {
  google::ParseCommandLineFlags(&argc, &argv, true);

  brpc::Channel channel;
  brpc::ChannelOptions options;
  options.protocol = "http";
  options.connection_type = "pooled";
  options.timeout_ms = 3000;
  options.max_retry = 0;

  brpc::ChannelSSLOptions* ssl_options = options.mutable_ssl_options();
  ssl_options->sni_name = "localhost";

  if (channel.Init("https://127.0.0.1:8443", "", &options) != 0) {
    std::cerr << "failed to init brpc channel\n";
    return 1;
  }

  std::string payload(static_cast<size_t>(FLAGS_object_size_kb) * 1024, 'x');
  std::atomic<int64_t> seq{0};
  std::atomic<int64_t> ok{0};
  std::atomic<int64_t> failed{0};

  auto deadline =
      std::chrono::steady_clock::now() + std::chrono::seconds(FLAGS_duration_sec);

  std::vector<std::thread> workers;
  for (int i = 0; i < FLAGS_concurrency; ++i) {
    workers.emplace_back([&] {
      while (std::chrono::steady_clock::now() < deadline) {
        int64_t id = seq.fetch_add(1);

        brpc::Controller cntl;
        cntl.http_request().set_method(brpc::HTTP_METHOD_PUT);
        cntl.http_request().uri() = "/object-" + std::to_string(id);
        cntl.http_request().set_content_type("application/octet-stream");
        cntl.request_attachment().append(payload);

        channel.CallMethod(nullptr, &cntl, nullptr, nullptr, nullptr);
        if (cntl.Failed()) {
          failed.fetch_add(1);
          // If the brpc bug is triggered, the process aborts before this line
          // can handle the request as a normal failure.
          std::cerr << "request failed: code=" << cntl.ErrorCode()
                    << " text=" << cntl.ErrorText() << "\n";
        } else {
          ok.fetch_add(1);
        }
      }
    });
  }

  for (auto& worker : workers) {
    worker.join();
  }

  std::cout << "ok=" << ok.load() << " failed=" << failed.load() << "\n";
  return 0;
}
```

编译方式取决于本地 brpc 安装方式，示例：

```bash
g++ -std=c++17 brpc_https_pooled_client.cc -o brpc_https_pooled_client \
  -lbrpc -lgflags -lssl -lcrypto -lprotobuf -lpthread
```

运行：

```bash
./brpc_https_pooled_client \
  --concurrency=128 \
  --duration_sec=60 \
  --object_size_kb=128
```

如果问题被触发，预期会看到类似：

```text
Fail to read from ssl_fd=...: error:...:SSL routines::unexpected eof while reading
Fail to read from ssl_fd=...: Success [0]
Controller::SetFailed(...): Check failed: false error_code is 0
Aborted (core dumped)
```

Felix-Gong:
## Summary

- Fix crash when HTTPS pooled connection encounters unexpected SSL EOF (error_code=0)
- Properly handle SSL_ERROR_SSL in socket.cpp DoRead() to return ESSL instead of 0
- Discovered during RISC-V porting and integration testing

## Root Cause

When OpenSSL 3.x detects unexpected EOF (peer closed without close_notify), SSL_read returns 0 with SSL_ERROR_SSL. The code didn't return -1, causing error_code=0 to propagate to Controller::SetFailed() which triggers CHECK(false).

## Fix

In the default branch of DoRead(), return -1 with errno=ESSL when SSL errors are detected, instead of falling through and returning nr.

Fixes #3307

### DIFF
--- a/src/brpc/socket.cpp
+++ b/src/brpc/socket.cpp
@@ -2133,7 +2133,14 @@ ssize_t Socket::DoRead(size_t size_hint) {
     default: {
         const unsigned long e = ERR_get_error();
         if (nr == 0) {
-            // Socket EOF or SSL session EOF
+            if (ssl_error != SSL_ERROR_ZERO_RETURN) {
+                // Unexpected EOF without proper SSL shutdown (close_notify)
+                LOG(WARNING) << "Fail to read from ssl_fd=" << fd()
+                             << ": unexpected ssl_error=" << ssl_error;
+                errno = ESSL;
+                return -1;
+            }
+            // Clean SSL shutdown (close_notify received)
         } else if (e != 0) {
             LOG(WARNING) << "Fail to read from ssl_fd=" << fd()
                          << ": " << SSLError(e);
--- a/test/brpc_ssl_unittest.cpp
+++ b/test/brpc_ssl_unittest.cpp
@@ -497,3 +497,79 @@ TEST_F(SSLTest, ssl_perf) {
     close(clifd);
     close(servfd);
 }
+
+struct AbruptCloseArgs { int listenfd; };
+
+static void* abrupt_close_server(void* arg) {
+    AbruptCloseArgs* a = (AbruptCloseArgs*)arg;
+    int connfd = accept(a->listenfd, NULL, NULL);
+    if (connfd < 0) return NULL;
+    SSL_CTX* ctx = brpc::CreateServerSSLContext(
+        "cert1.crt", "cert1.key", brpc::SSLOptions(), NULL, NULL);
+    SSL* ssl = brpc::CreateSSLSession(ctx, 0, connfd, true);
+    if (ssl) { SSL_do_handshake(ssl); SSL_free(ssl); }
+    close(connfd);
+    return NULL;
+}
+
+TEST_F(SSLTest, ssl_unexpected_eof) {
+    // Verify that Socket::DoRead() returns -1 with errno=ESSL when the
+    // remote side closes the TCP connection without sending close_notify.
+    // Without the fix, DoRead() returns 0, causing error_code=0 to
+    // propagate to Controller::SetFailed() which triggers CHECK(false).
+
+    const int port = 5962;
+    butil::EndPoint ep(butil::IP_ANY, port);
+    butil::fd_guard listenfd(butil::tcp_listen(ep));
+    ASSERT_GT(listenfd, 0);
+
+    AbruptCloseArgs server_args = { listenfd };
+    pthread_t server_tid;
+    ASSERT_EQ(0, pthread_create(&server_tid, NULL, abrupt_close_server,
+                                &server_args));
+
+    brpc::Protocol dummy_protocol = {
+        brpc::policy::ParseRpcMessage, brpc::SerializeRequestDefault,
+        brpc::policy::PackRpcRequest, NULL, ProcessResponse,
+        NULL, NULL, NULL, brpc::CONNECTION_TYPE_ALL, "ssl_ut_eof"
+    };
+    ASSERT_EQ(0, RegisterProtocol((brpc::ProtocolType)31, dummy_protocol));
+
+    brpc::InputMessageHandler dummy_handler = {
+        dummy_protocol.parse, dummy_protocol.process_response,
+        NULL, NULL, dummy_protocol.name
+    };
+    brpc::InputMessenger messenger;
+    ASSERT_EQ(0, messenger.AddHandler(dummy_handler));
+
+    brpc::SocketOptions socket_options;
+    butil::EndPoint server_ep(butil::IP_ANY, port);
+    socket_options.remote_side = server_ep;
+    socket_options.connect_on_create = true;
+    // Do NOT set on_edge_triggered_events — we will call DoRead manually.
+    socket_options.user = &messenger;
+
+    brpc::ChannelSSLOptions ssl_options;
+    SSL_CTX* raw_ctx = brpc::CreateClientSSLContext(ssl_options);
+    ASSERT_NE(nullptr, raw_ctx);
+    std::shared_ptr<brpc::SocketSSLContext> ssl_ctx =
+        std::make_shared<brpc::SocketSSLContext>();
+    ssl_ctx->raw_ctx = raw_ctx;
+    socket_options.initial_ssl_ctx = ssl_ctx;
+
+    brpc::SocketId socket_id;
+    ASSERT_EQ(0, brpc::Socket::Create(socket_options, &socket_id));
+    brpc::SocketUniquePtr ptr;
+    ASSERT_EQ(0, brpc::Socket::Address(socket_id, &ptr));
+
+    // Wait for server to close the connection without close_notify.
+    pthread_join(server_tid, NULL);
+    usleep(50000);
+
+    // DoRead should detect the unexpected EOF and return -1 with errno=ESSL.
+    ssize_t nr = ptr->DoRead(1024);
+    EXPECT_EQ(-1, nr);
+    EXPECT_EQ(brpc::ESSL, errno);
+
+    ptr->SetFailed();
+}


src/brpc/socket.cpp
```diff
@@ -2133,11 +2133,19 @@ ssize_t Socket::DoRead(size_t size_hint) {
...
+                             << ": unexpected ssl_error=" << ssl_error;
+                errno = ESSL;
+                return -1;
+            }
+            // Clean SSL shutdown (close_notify received)
         } else if (e != 0) {
             LOG(WARNING) << "Fail to read from ssl_fd=" << fd()
                          << ": " << SSLError(e);
             errno = ESSL;
+            return -1;
```
wwbmmm: Why modify this branch?

src/brpc/socket.cpp
```diff
@@ -2146,6 +2154,10 @@ ssize_t Socket::DoRead(size_t size_hint) {
                                    BIO_fd_non_fatal_error(saved_errno) != 0 ||
                                   nr < 0;
             PLOG_IF(WARNING, is_fatal_error) << "Fail to read from ssl_fd=" << fd();
+            if (is_fatal_error) {
```
wwbmmm: Why modify this branch?

src/brpc/socket.cpp
Felix-Gong: You're right. No changes are needed for this branch — the real fix is only in the if (nr == 0) branch where the original code fell through and returned 0 without distinguishing unexpected SSL EOF from a clean shutdown. I'll revert this change.

src/brpc/socket.cpp
Felix-Gong: This branch is changing errno from saved_errno to ESSL, which is a minor semantic improvement but not a bug fix.

wwbmmm: If you change this behavior, you'd better check the callers' code and make sure they don't depend on the value of `errno`.

wwbmmm:
BTW, Can you add some unit tests to verify this fix?

src/brpc/socket.cpp
Felix-Gong: The caller only checks errno == EAGAIN and errno == EINTR for retry logic. ESSL matches neither, so control flow is unchanged — it still enters SetFailed().
The only difference is the error message. But I'll revert this change to keep the fix minimal.

Felix-Gong:
> BTW, Can you add some unit tests to verify this fix?

I'll think about how to add a test that covers the full DoRead() path for the unexpected SSL EOF scenario, and submit it later.

Felix-Gong:
> BTW, Can you add some unit tests to verify this fix?

@wwbmmm I have added unit test `ssl_unexpected_eof` in PR #3321 to verify this fix. The test creates an SSL server that abruptly closes the connection without `close_notify`, then
  verifies `DoRead()` returns `-1` with `errno=ESSL`. CI of #3321 will fail until this fix is merged; after merging, rebase #3321 and CI should pass.

chenBright:
I think you should add the unit test `ssl_unexpected_eof` to this PR, rather than creating a new PR.

Felix-Gong:
> I think you should add the unit test `ssl_unexpected_eof` to this PR, rather than creating a new PR.

I'd prefer to keep the test in a separate PR (#3321) because the fix and the test are independent changes. This way the fix can be reviewed and merged first, and the test PR can be rebased to verify CI passes. If you think it's better to combine them, I can merge them into one PR.

chenBright:
> Add unit test ssl_unexpected_eof to verify the fix in https://github.com/apache/brpc/pull/3316

I think it would be more appropriate to merge them into one PR.

Felix-Gong:
Done. The unit test `ssl_unexpected_eof` has been added to this PR (commit dc09eaab). Please review.

wwbmmm:
LGTM

chenBright (commented):
LGTM

PR #3320, opened 2026-06-01
fix: guard SO_BINDTODEVICE

daeho-ro:

Issue Number: resolve 

Problem Summary: MacOS 14 doesn't have `SO_BINDTODEVICE` and so guard it only if it is defined.

Changed: Check the `SO_BINDTODEVICE` is defined or not.

Side effects:
- Performance effects:

---
### Check List:


wwbmmm:
LGTM

chenBright (commented):
LGTM

PR #3321, opened 2026-06-01
test: add unit test for unexpected SSL EOF handling

Felix-Gong:
## Summary
- Add unit test `ssl_unexpected_eof` to verify the fix in #3316
- Test verifies that `Socket::DoRead()` returns `-1` with `errno=ESSL` when the remote side closes the TCP connection without sending `close_notify`

## Note
- Depends on #3316 being merged first (CI will fail until then)
- Verified on x86 and RISC-V, all SSL tests pass


Felix-Gong:
Merged test into #3316 as requested.

PR #2093, opened 2023-01-17
Add Mysql Protocol

yanglimingcn:
Add Mysql Protocol, support text protocol, transaction, support prepare statement.

issue #209

### What problem does this PR solve?

Issue Number:

Problem Summary:

### What is changed and the side effects?

Changed:

Side effects:
- Performance effects(性能影响):

- Breaking backward compatibility(向后兼容性): 

---
### Check List:
- Please make sure your changes are compilable(请确保你的更改可以通过编译).
- When providing us with a new feature, it is best to add related tests(如果你向我们增加一个新的功能, 请添加相关测试).
- Please follow [Contributor Covenant Code of Conduct](../../master/CODE_OF_CONDUCT.md).(请遵循贡献者准则).

PR #3330, opened 2026-06-04
Add MySQL client protocol (text protocol, transactions, prepared statements) on clean-room auth (#2093)

rajvarun77:
Implements a MySQL client protocol for brpc built on a **clean-room** authentication codec (no GPL lineage), addressing #2093.

### Scope
- **Clean-room connection-phase auth** — `mysql_native_password` and `caching_sha2_password` (fast-auth + full-auth RSA / secure-transport cleartext), derived from the public MySQL protocol documentation.
- **COM_QUERY text protocol** — OK / ERR / EOF / text result-set parsing.
- **Interactive transactions** via connection affinity (reserve/use a pooled connection; SINGLE rejected, auto-rollback on drop).
- **Prepared statements** (binary protocol) with per-connection `stmt_id` caching and transparent re-prepare when a statement lands on a fresh connection.

### Shared-core footprint (kept redis-like / minimal)
The only generic `Controller`/`Socket` additions are: a socket reserve/use hook and `fd_version` (ABA guard) for transaction/prepared-statement affinity, an auth empty-write guard for MySQL's server-greets-first handshake, and protocol registration. **No protocol-specific type lives on the shared `Controller`** (the prepared-statement pointer rides a generic opaque per-RPC slot).

### Tests
Clean-room integration tests (transactions, prepared statements, pooled-connection concurrency, connection-type boundary) run against a self-spawned `mysqld`; **19/19 passing locally**. Concurrency coverage includes per-transaction connection pinning under overlap, abort/auto-rollback under contention, and prepared re-prepare when a connection is stolen by a transaction.


rajvarun77:
@chenBright @wwbmmm — marking this ready for review.

Full MySQL client protocol for brpc — COM_QUERY text protocol, interactive transactions, and prepared statements — built on a **clean-room** authentication codec (no GPL lineage), addressing #2093.

- **Auth (clean-room):** `mysql_native_password` + `caching_sha2_password` (fast-auth + full-auth RSA / secure-transport cleartext), derived from the public MySQL protocol docs.
- **Minimal shared-core footprint (redis-like):** the only generic `Controller`/`Socket` additions are a socket reserve/use hook + `fd_version` (ABA guard) for transaction/prepared-statement affinity, an auth empty-write guard for the server-greets-first handshake, and protocol registration. No protocol-specific type lives on the shared `Controller`.
- **Tests:** clean-room integration tests (transactions, prepared statements, pooled-connection concurrency, connection-type, binary TIME/DATETIME) run against a self-spawned `mysqld` — 20/20 passing locally; a standalone 30-case end-to-end run was independently verified.

The clean-room auth codec is also up as a smaller, focused PR (#3310, Stage 1a) if you'd prefer to review that foundation first.

src/brpc/policy/mysql/mysql.h
```diff
@@ -0,0 +1,244 @@
...
+    bool AddParam(float p);
+    bool AddParam(double p);
+    bool AddParam(const butil::StringPiece& p);
+
+    // True if previous command failed.
+    bool has_error() const {
+        return _has_error;
+    }
+
+    const MysqlTransaction* get_tx() const {
```
wwbmmm: simple getter, just use `tx()`

src/brpc/policy/mysql/mysql.h
```diff
@@ -0,0 +1,244 @@
...
+        return _tx;
+    }
+
+    MysqlStatementStub* get_stmt() const {
```
wwbmmm: simple getter, just use `stmt()`

src/brpc/socket.cpp
```diff
@@ -578,6 +579,9 @@ int Socket::ResetFileDescriptor(int fd) {
     _avg_msg_size = 0;
     // MUST store `_fd' before adding itself into epoll device to avoid
     // race conditions with the callback function inside epoll
+    static butil::atomic<uint64_t> BAIDU_CACHELINE_ALIGNMENT fd_version(0);
+    _fd_version.store(fd_version.fetch_add(1, butil::memory_order_relaxed),
```
wwbmmm, line 583: It seems that this _fd_version doesn't need to be atomic, since the static fd_version is atomic.

src/brpc/controller.h
```diff
@@ -915,6 +925,16 @@ friend void policy::ProcessThriftRequest(InputMessageBase*);
     // Defined at both sides
     StreamSettings *_remote_stream_settings;
 
+    // Whether/how to reserve the sending socket after the RPC (mysql transactions).
+    BindSockAction _bind_sock_action;
```
wwbmmm: I think this can be a flag in `Controller::_flags`, no need to add a new member.

src/brpc/policy/mysql/mysql_protocol.cpp
```diff
@@ -0,0 +1,536 @@
...
+    if (request->GetDescriptor() != MysqlRequest::descriptor()) {
+        return cntl->SetFailed(EREQUEST, "The request is not a MysqlRequest");
+    }
+    const MysqlRequest* rr = (const MysqlRequest*)request;
+    // We work around SerializeTo of pb which is just a placeholder.
+    if (!rr->SerializeTo(buf)) {
+        return cntl->SetFailed(EREQUEST, "Fail to serialize MysqlRequest");
+    }
+    // mysql doesn't use pipelined_count to verify the end of a response; instead we
+    // reuse it as a MysqlStmtType tag so the parse function knows which reply shape
```
wwbmmm, line 468: This is confusing. I think we can add some alias function to `ControllerPriveAccessor` to make it more clear.

src/brpc/socket.cpp
rajvarun77: Thanks — you're right that the *uniqueness* of the value comes from the static `fd_version` atomic. But the per-`Socket` `_fd_version` still needs to be atomic for a different reason: it is written and read from different threads with no lock between them.

- Writer: `ResetFileDescriptor()` stores `_fd_version` on the reconnect path (`CheckConnectedAndKeepWrite` -> `ResetFileDescriptor`), with no socket mutex held.
- Reader: the mysql prepared-statement path reads `socket->fd_version()` on the RPC bthread (`MysqlStatement::StatementId`). It obtains the socket via `Socket::Address()`, which only gates on the `SocketId` version — and `ResetFileDescriptor` does not bump that version — so `Address()` succeeds even while the fd is being reset. The read can therefore overlap the write on the same Socket.

A concurrent non-atomic read/write on the same object is a data race (UB), and a 64-bit load is not tear-free on 32-bit targets, so the member has to be atomic. This mirrors `_fd` itself, which is a `butil::atomic<int>` written in the same `ResetFileDescriptor` and read lock-free (`memory_order_relaxed`) all over the IO path. I kept `_fd_version` on the same relaxed-atomic pattern. Happy to add a brief comment to that effect if it helps.

src/brpc/policy/mysql/mysql_common.h
```diff
@@ -0,0 +1,422 @@
...
+#include <map>
+#include "butil/logging.h"  // LOG()
+
+namespace brpc {
+// Msql Collation
+extern const char* MysqlDefaultCollation;
+extern const char* MysqlBinaryCollation;
+const std::map<std::string, uint8_t> MysqlCollations = {
+    {"big5_chinese_ci", 1},
+    {"latin2_czech_cs", 2},
```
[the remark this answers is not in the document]
rajvarun77: Addressed in the current revision — `MysqlCollations` is declared `extern` in `mysql_common.h` and defined once in `mysql_common.cpp`, so there is no in-header definition / ODR issue.

src/brpc/policy/mysql/mysql.cpp
```diff
@@ -0,0 +1,530 @@
...
+    _cached_size_ = 0;
+}
+
+size_t MysqlResponse::ByteSizeLong() const {
+    return _cached_size_;
+}
+
+void MysqlResponse::MergeFrom(const MysqlResponse& from) {
+    CHECK_NE(&from, this);
+}
```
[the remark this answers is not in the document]
rajvarun77: Fixed — `MysqlResponse` is now explicitly non-copyable (deleted copy ctor + assignment), and `MergeFrom` is a hard `CHECK`-failure instead of a silent no-op, so an accidental copy/CopyFrom is caught rather than silently dropping parsed replies.

test/mysql/brpc_mysql_connection_type_unittest.cpp
```diff
@@ -0,0 +1,373 @@
...
+//
+//   * PreparedStatementUnderShortMustError (PRIMARY):
+//       build a SHORT channel, prepare "SELECT ? AS v", bind one INT param,
+//       CallMethod.  Must ERROR (cntl.Failed() OR reply(0).is_error()); must
+//       NOT return a correct result set; must NOT crash.  Looped a few times.
+//
+//   * PlainQueryUnderShortMustSucceed (POSITIVE CONTROL):
+//       same SHORT channel; a stateless COM_QUERY "SELECT 7 AS v" must SUCCEED
+//       and return 7.  Proves SHORT is fine for stateless queries; only the
+//       connection-scoped prepared-statement handle breaks under SHORT.
```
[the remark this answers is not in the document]
rajvarun77: Addressed — the file-header comment now describes `PreparedStatementUnderShortRePreparesAndSucceeds` (success via transparent re-prepare on each new short connection), matching the actual test.

docs/cn/mysql_client.md
```diff
@@ -0,0 +1,556 @@
+[MySQL](https://www.mysql.com/)是著名的开源的关系型数据库，为了使用户更快捷地访问mysql并充分利用bthread的并发能力，brpc直接支持mysql协议。示例程序：[example/mysql_c++](https://github.com/brpc/brpc/tree/master/example/mysql_c++/)
+
+**注意**：只支持MySQL 4.1 及之后的版本的文本协议，支持事务，不支持Prepared statement。目前支持的鉴权方式为mysql_native_password，使用事务的时候不支持single模式。
```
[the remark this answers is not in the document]
rajvarun77: Addressed — the overview now states that prepared statements are supported, consistent with the Prepared Statement section.

src/brpc/policy/mysql/mysql_statement.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+#include <vector>
+#include <gflags/gflags.h>
+#include "brpc/socket.h"
+#include "brpc/policy/mysql/mysql_statement.h"
+
+namespace brpc {
+DEFINE_int32(mysql_statment_map_size,
+             100,
+             "Mysql statement map size, usually equal to max bthread number");
```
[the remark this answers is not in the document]
rajvarun77: Fixed — the flag is spelled `mysql_statement_map_size` in the current revision (DEFINE + all uses).

src/brpc/policy/mysql/mysql.h
rajvarun77: Done — renamed `get_tx()` to `tx()` and updated the callers.

src/brpc/policy/mysql/mysql.h
rajvarun77: Done — renamed `get_stmt()` to `stmt()` and updated the callers.

src/brpc/controller.h
rajvarun77: Done — the `BindSockAction` is now stored in two bits of `Controller::_flags` (`FLAGS_BIND_SOCK_RESERVE`/`FLAGS_BIND_SOCK_USE`) via `set_bind_sock_action()`/`bind_sock_action()`, instead of a dedicated member. No behavior change.

src/brpc/policy/mysql/mysql_protocol.cpp
rajvarun77: Done — added `ControllerPrivateAccessor::set_mysql_statement_type()` as a clearly named alias over the `pipelined_count` slot the mysql protocol reuses, and the protocol now calls that instead of `set_pipelined_count()` directly.

rajvarun77:
@wwbmmm @chenBright — pushed a revision addressing the review feedback and fixing the integration-test setup. Summary:

**Review comments**
- Stored the mysql-transaction `BindSockAction` in two `Controller::_flags` bits instead of a dedicated member.
- Made `MysqlResponse` explicitly non-copyable and turned the previously no-op `MergeFrom` into a hard `CHECK`-failure (so an accidental copy is caught instead of silently dropping replies).
- Renamed `get_tx()`/`get_stmt()` to `tx()`/`stmt()`.
- Added a clearly named `ControllerPrivateAccessor::set_mysql_statement_type()` alias over the `pipelined_count` slot the mysql protocol reuses.
- The `MysqlCollations` ODR note, the gflag spelling, and the two doc/comment mismatches were already addressed in the current revision.
- On `_fd_version`: I left a reply explaining why it needs to stay atomic — it is written in `ResetFileDescriptor` (reconnect path, no lock) and read lock-free from another thread via `fd_version()` in the prepared-statement path, so a non-atomic field would be a data race. It mirrors how `_fd` itself is handled. Happy to discuss further.

**Integration tests now run against a real MySQL server**
- The suites previously did not exercise a live server in CI: a legacy test pointed at an external host and hard-failed, and the make/bazel wiring built it as a single globbed target. I removed the legacy test, switched the `test/mysql/*` suites to the redis-style self-spawning + skip-if-absent pattern, build one cc_test per file, and merged master so the CI image installs `mysql-server`. The suites now spawn a local throwaway mysqld and run.

**Local verification**
I built and ran all seven mysql suites locally against a live MySQL 9.6 server: 110 tests, 0 skipped, 0 failed — covering `caching_sha2_password` auth, the text and binary (prepared-statement) protocols, transactions, and connection pooling. (MySQL 9.x removed `mysql_native_password`, so this also exercises the caching_sha2 path end-to-end.)

CI is running on the latest revision. Could you take another look when you have a chance? Thanks for the reviews.

rajvarun77:
@wwbmmm @chenBright — heads-up on CI: the two `*-unittest-with-bazel` lanes show one failure, and it is an unrelated flake, not a regression from this PR.

- The failing target is `//test:brpc_channel_unittest`, and within it a single timing-sensitive case, `ChannelTest.backup_request`, hung and blew the 300s test timeout (the timeout stack trace points at `ChannelTest::TestBackupRequest()`). It is a timeout, not an assertion failure.
- The **same commit passes** that test in the make lanes (`clang-unittest` and `clang-unittest-asan` are green), and **all seven `//test:mysql/*` targets pass** in the bazel lanes.
- It is independent of this change: the new flags use `_flags` bits 14/15 while backup-request uses `FLAGS_BACKUP_REQUEST` (bit 5) — no collision — and the backup-request path never touches `bind_sock_action()`.

A re-run of the failed jobs should clear it (I cannot trigger re-runs as an outside contributor).

This PR is ready for another look whenever you have time: all review threads are addressed (I left one reply on `_fd_version` explaining why it stays atomic), the mysql integration suites now run against a real CI-provisioned mysqld, and I verified all seven suites locally against a live MySQL 9.6 (110 tests, 0 skipped, 0 failed). Thanks!

rajvarun77:
Moved all seven MySQL unittests from `test/mysql/` into `test/` (following the review on #3310), so the existing `brpc_*_unittest` glob picks them up. `test/Makefile`, `test/CMakeLists.txt`, and `test/run_tests.sh` revert to master with no mysql-specific lines. In `test/BUILD.bazel` the files join the `brpc_unittests` glob; the five tests that fork a real `mysqld` keep their `["external", "local"]` tags by folding those entries into that target's `per_test_tags` (rather than a separate mysql target).

All seven tests compile and run green in the unittest jobs — auth handshake 28 / packet 27 / scramble 35, plus connection-type, pool-concurrency, prepared, and txn integration. Any remaining red is a pre-existing unrelated flake (`brpc_channel_unittest`/`brpc_load_balancer_unittest` timeouts, `SocketTest`, `EventDispatcherTest` under ASan), not this change.

rajvarun77:
Merged the latest master (which now includes the merged #3310) into this branch and resolved the add/add conflict on `test/brpc_mysql_auth_handshake_unittest.cpp` by keeping master's version — it is identical to this branch's copy plus the explicit `main()` that #3310 landed (parses the `-mysql_*` flags for the running-server mode). The branch is mergeable again and CI is re-running.

@wwbmmm @yanglimingcn — ready for another look when you have a moment.

rajvarun77:
This is ready for review and merge. The branch is up to date with master (the #3310 merge conflict is resolved) and `mergeable`.

The only failing checks are the three bazel unittest jobs, all on the **same unrelated** target `//test:brpc_channel_unittest` (78/79 pass in each):

- `clang-unittest-with-bazel` / `gcc-unittest-with-bazel` — `brpc_channel_unittest` TIMEOUT at 300s
- `clang-unittest-bazel-with-babylon-and-new-pb` — `ChannelTest.backup_request_policy` failed

This is a pre-existing timing flake, not related to this PR's diff. The failing assertion is `test/brpc_channel_unittest.cpp:2208` — a 100 ms hang-guard on socket teardown — which overshot by <1 ms (100,896 µs) under bazel's parallel execution + 300 s per-test cap on the 2-vCPU runner. The same test passes in the serial make/cmake unittest jobs (`clang-unittest` and `clang-unittest-asan` are green), and the MySQL tests this PR touches all run and pass there (auth handshake 28 / packet 27 / scramble 35).

@chenBright @wwbmmm @yanglimingcn — could you take a look and merge when you have a moment? Happy to open a separate follow-up for the `brpc_channel_unittest` flake/timeout if useful.

src/brpc/socket.cpp
rajvarun77: @wwbmmm could you please take a look.

rajvarun77:
Restructured for easier review: the generic `Controller`/`Socket` connection-affinity primitives (and the `ChannelTest.backup_request` uninitialized-`bind_sock_action` fix) are split into a small standalone PR #3348. This PR is now **stacked on top of #3348** and contains only the MySQL-specific code (protocol codec, registration, `PROTOCOL_MYSQL`, accessor wrappers, tests, docs).

The diff here will show the core changes too until #3348 merges into master, after which it auto-reduces to MySQL-only. No functional change from the previous revision — the combined tree is byte-identical.

rajvarun77:
Update: reverted the split — the core/MySQL separation isn't warranted for a one-line fix. #3330 is again the single self-contained PR (MySQL protocol + the generic affinity primitives), and the `ChannelTest.backup_request` fix (uninitialized `bind_sock_action` in the `Call(Call*)` copy constructor) is folded directly into it. #3348 is closed. No functional change.

rajvarun77:
@wwbmmm @yanglimingcn @chenBright — this is green on all lanes and ready for another look.

The earlier `brpc_channel_unittest` failure that I'd thought was an unrelated flake turned out to be a real bug this PR introduced: the new `bind_sock_action` member of `Controller::Call` was not default-initialized, so the backup/retry call (built via the `Call` copy constructor) read an indeterminate value in `Call::OnComplete`, which branches on it before the normal pool-return / `SetFailed` path. When the garbage happened to match `BIND_SOCK_RESERVE`/`BIND_SOCK_USE`, the backup call's socket was diverted instead of returned, leaving the in-flight RPC unconcluded and hanging `ChannelTest.backup_request` (nondeterministically, hence the flaky appearance). Fixed by giving the member an in-class default initializer (`BIND_SOCK_NONE`) so every `Call` construction path is safe; a backup/retry never inherits transaction affinity. `ChannelTest.backup_request` now passes in under a second.

All prior review comments are addressed. The one remaining open thread — whether per-`Socket` `_fd_version` needs to be atomic — has a reply explaining why: it is written on the reconnect path with no socket lock held and read from another thread, so the relaxed atomic is there to make that cross-thread access race-free, independent of the global counter's uniqueness. Happy to adjust if you'd prefer a different framing. Could you take another pass when you have a moment?

rajvarun77:
@chenBright — could you review this PR when you have a chance? It's green on all CI lanes and mergeable. All earlier review threads are addressed, and the previous `brpc_channel_unittest` failure is fixed (it was an uninitialized `bind_sock_action` in the `Call` copy constructor, now default-initialized). Would value your eyes on the MySQL protocol codec and the core `Controller`/`Socket` affinity primitives. Thanks!

src/brpc/controller.h
```diff
@@ -762,6 +776,10 @@ friend void policy::ProcessThriftRequest(InputMessageBase*);
         // CONNECTION_TYPE_SINGLE. Otherwise, it may be a temporary
         // socket fetched from socket pool
         SocketUniquePtr sending_sock;
+        // In-class default so every Call init path (default ctor, Reset(), and
+        // the Call(Call*) copy ctor used for backup/retry) is NONE unless set
+        // explicitly. A backup/retry never inherits transaction affinity.
+        BindSockAction bind_sock_action = BIND_SOCK_NONE;
```
chenBright: `bind_sock_action` needs to be set in `ResetPods()`.

chenBright:
Okay, I'll review it when I have time.

chenBright:
The MySQL protocol was originally developed by @yanglimingcn, who is most familiar with it. @yanglimingcn Please review it when you have time.

wwbmmm:
LGTM

src/brpc/socket.h
```diff
@@ -840,6 +843,9 @@ friend class TransportFactory;
     butil::atomic<int> _fd;  // -1 when not connected.
     int _tos;                // Type of service which is actually only 8bits.
     int64_t _reset_fd_real_us; // When _fd was reset, in microseconds.
+    // ABA/version counter; written on fd reset and read via fd_version() from
+    // other threads, so use relaxed atomics to avoid a data race.
+    butil::atomic<uint64_t> _fd_version;  // _fd_version, used only for mysql now.
```
chenBright: What is `_fd_version` used for?

rajvarun77: Used to track when the underlying physical connection changes. For the new MySQL protocol implementation, we need to detect socket reconnections so we know exactly when to safely reinitialize the MySQL prepared statements.

src/brpc/controller.h
rajvarun77: Thanks for the pointer I will get it to it by eow and then resolve this comment.

rajvarun77: Done in `530e1124`: `Call::bind_sock_action` is now initialized explicitly in every constructor (incl. the backup/retry copy ctor), not via an in-class default. That missing init was the cause of the backup/retry hang. CI green.

rajvarun77:
@wwbmmm @yanglimingcn @chenBright — addressed the `ResetPods`/`bind_sock_action` comment in `530e1124`; all 15 CI lanes green (incl. the bazel lanes that flaked on `ChannelTest.backup_request`). PTAL and merge if it looks good. Thanks!

rajvarun77:
@wwbmmm @yanglimingcn @chenBright — gentle follow-up: this has been green on all CI lanes for a couple of days (the `bind_sock_action`/`ResetPods` comment is addressed in `530e1124`). Could you take another look and merge if it looks good? Thanks!

rajvarun77:
@chenBright — checking in on the path to merge here. It has been three weeks since the handoff to @yanglimingcn (Jun 16) with no activity, and the PR has been merge-ready throughout: 15/15 CI lanes green, all review threads resolved (your `ResetPods`/`bind_sock_action` comment addressed in `530e1124`), and @wwbmmm's LGTM. Would @wwbmmm's approval be sufficient to merge, or is @yanglimingcn's review a hard requirement?

@yanglimingcn — happy to walk through any part of the implementation (clean-room auth handshake, prepared-statement/transaction connection affinity, binary protocol) if that would make the review easier. And if you don't have bandwidth right now, no problem at all — just let us know how you'd like to proceed. Thanks!

src/brpc/policy/mysql/mysql.cpp
```diff
@@ -0,0 +1,537 @@
...
+#include "butil/logging.h"
+#include "brpc/controller.h"
+#include "brpc/policy/mysql/mysql.h"
+#include "brpc/policy/mysql/mysql_common.h"
+
+namespace brpc {
+
+DEFINE_int32(mysql_multi_replies_size, 10, "multi replies size in one MysqlResponse");
+
+// ===================================================================
```
chenBright: Please remove redundant comments.

src/brpc/policy/mysql/mysql.cpp
```diff
@@ -0,0 +1,537 @@
...
+        os << std::hex << (unsigned)buf;
+    }
+}
+
+std::ostream& operator<<(std::ostream& os, const MysqlRequest& r) {
+    r.Print(os);
+    return os;
+}
+
+// ===================================================================
```
chenBright: Please remove redundant comments.

src/brpc/policy/mysql/mysql.cpp
```diff
@@ -0,0 +1,537 @@
...
+
+#ifndef _MSC_VER
+#endif  // !_MSC_VER
```
chenBright: Please remove unused code.

src/brpc/policy/mysql/mysql.cpp
```diff
@@ -0,0 +1,537 @@
...
+    if (other != this) {
+        _first_reply.Swap(other->_first_reply);
+        std::swap(_other_replies, other->_other_replies);
+        _arena.swap(other->_arena);
+        std::swap(_nreply, other->_nreply);
+        std::swap(_cached_size_, other->_cached_size_);
+    }
+}
+
+// ===================================================================
```
chenBright: Please remove redundant comments.

src/brpc/controller.cpp
```diff
@@ -297,6 +297,9 @@ void Controller::ResetPods() {
     _request_streams.clear();
     _response_streams.clear();
     _remote_stream_settings = NULL;
+    set_bind_sock_action(BIND_SOCK_NONE);
```
chenBright: Since `_flags = 0` has already been set in line 244, there's no need to call `set_bind_sock_action` again.

src/brpc/controller.cpp
```diff
@@ -297,6 +297,9 @@ void Controller::ResetPods() {
...
+    _bind_sock.reset();
```
chenBright: `_bind_sock.reset();` should be called in `ResetNonPods()`.

src/brpc/policy/mysql/mysql.cpp
rajvarun77: Removed — also stripped the same separators from the four mysql test files.

src/brpc/policy/mysql/mysql.cpp
rajvarun77: Removed.

src/brpc/policy/mysql/mysql.cpp
rajvarun77: Removed.

src/brpc/policy/mysql/mysql.cpp
rajvarun77: Removed.

src/brpc/controller.cpp
rajvarun77: Right — the _flags = 0 above already clears those bits. Removed.

src/brpc/controller.cpp
rajvarun77: Done — moved to ResetNonPods().

chenBright (approved):
LGTM

PR #3331, opened 2026-06-05
bthread: support per-tag CPU affinity in --cpu_set

yanglimingcn:
Previously --cpu_set accepted a single CPU list (e.g. "0-3,5,7") that was applied uniformly to all bthread worker threads regardless of their bthread_tag.

This change extends the flag to accept a per-tag format:

  --cpu_set="0:0-3,5,7;1:6-9,4"

where each semicolon-separated segment is "<tag>:<cpu-list>".  Tags not mentioned in the string get no CPU binding.  The legacy single-list format continues to work unchanged and binds all tags to the same set.

Implementation:
- Replace _cpus (vector<unsigned>) with _tag_cpus (vector<vector<unsigned>>), sized to FLAGS_task_group_ntags in the constructor.
- Add parse_one_cpuset() (static helper) for the existing range-list parsing logic; parse_cpuset() now dispatches between legacy and per-tag formats based on the presence of ':' or ';'.
- worker_thread() looks up _tag_cpus[tag] and round-robins over the per-tag CPU list using the global _next_worker_id counter.

Issue Number: resolve 

Problem Summary:

Changed:

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/bthread/task_control.cpp
+++ b/src/bthread/task_control.cpp
@@ -48,8 +48,13 @@ DEFINE_int32(task_group_ntags, 1, "TaskGroup will be grouped by number ntags");
 DEFINE_bool(task_group_set_worker_name, true,
             "Whether to set the name of the worker thread");
 DEFINE_string(cpu_set, "",
-              "Set of CPUs to which cores are bound. "
-              "for example, 0-3,5,7; default: disable");
+              "Set of CPUs to which worker threads are bound. "
+              "Two formats are supported:\n"
+              "  Legacy (bind all tags to one set): \"0-3,5,7\"\n"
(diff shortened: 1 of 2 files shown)


chenBright (commented):
LGTM

src/bthread/task_control.cpp
```diff
@@ -82,7 +87,7 @@ void run_tagged_worker_startfn(bthread_tag_t tag) {
 
 struct WorkerThreadArgs {
     WorkerThreadArgs(TaskControl* _c, bthread_tag_t _t) : c(_c), tag(_t) {}
-    TaskControl* c;
+    TaskControl*  c;
```
wwbmmm: duplicate space

wwbmmm:
LGTM

PR #3024, opened 2025-07-10
add inline redis protocol support

thweetkomputer:
### What problem does this PR solve?

Issue Number:

Problem Summary:

Add inline redis protocol support. Many redis benchmark tools use inline protocol, such as `redis_benchmark`, `dfly_bench`, although inline protocol is not guaranteed to be support for a redis-compatible database.

### What is changed and the side effects?

Changed:

Side effects:
- Performance effects:

There will be no overhead for commands following normal protocol (RESP2).

- Breaking backward compatibility: 

No compatiblity issues should be introduced unless someone replies on brpc to reject inline commands.

---
### Check List:
- Please make sure your changes are compilable.
- When providing us with a new feature, it is best to add related tests.
- Please follow [Contributor Covenant Code of Conduct](https://github.com/apache/brpc/blob/master/CODE_OF_CONDUCT.md).

PR #3338, opened 2026-06-13
Fix inline redis protocol consuming HTTP/2 connection preface (#3109)

### LINKED ISSUE #3109: GRPC client error: connection closed before server preface received
**Describe the bug**
When using a GRPC client to call a brpc server with redis protocol enabled, the client will failed with `code = Unavailable desc = connection closed before server preface received`.

Introduced by #3024 , the Redis protocol check is too loose, and it parses the GRPC protocol as the Redis protocol.

**To Reproduce**

This is the code to reproduce the the issue.
``` c++
#include <brpc/server.h>

class RedisService : public brpc::RedisService {
};

int main(int argc, char* argv[]) {
    google::ParseCommandLineFlags(&argc, &argv, true);
    brpc::Server server;
    brpc::ServerOptions options;
    RedisService redis_service;
    options.redis_service = &redis_service;
    if (server.Start(7447, &options) != 0) {
        LOG(ERROR) << "Failed to start dummy server";
        return 1;
    }
    server.RunUntilAskedToQuit();
    return 0;
}
```

**Expected behavior**

The GRPC client failed with `code = Unimplemented desc = [10.234.44.157:4003][E1002]Fail to find method on /service.method"

**Versions**
OS: ubuntu 20.04 kernel: 5.15.0
Compiler: clang-12
brpc: 1.14.1
protobuf: 27.3

rajvarun77:

Issue Number: resolve #3109

Problem Summary:

Inline redis protocol support (#3024) accepts any alpha-leading line as an inline command. The HTTP/2 client connection preface (`PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n`) is alpha-leading, so `RedisCommandParser::ConsumeImpl` parsed it as an inline command (first token `PRI`) and returned `PARSE_OK` instead of `PARSE_ERROR_TRY_OTHERS`.

As a result, protocol auto-detection in `InputMessenger` never fell through to the HTTP/2 handler, and a gRPC client calling a brpc server with `redis_service` enabled failed with:

```
code = Unavailable desc = connection closed before server preface received
```

Changed:

- In the inline branch of `RedisCommandParser::ConsumeImpl` (`src/brpc/redis_command.cpp`), detect the HTTP/2 connection preface — both a full match and a not-yet-complete prefix — and return `PARSE_ERROR_TRY_OTHERS`, leaving the buffer untouched so detection falls through to the HTTP/2 protocol.
- The check is gated on the first byte (`'P'`) to avoid overhead on normal inline commands.

Why match the preface rather than blacklist command names: some HTTP methods are also valid redis commands (e.g. `GET`), so a name-based blacklist is inherently ambiguous. The HTTP/2 preface is a fixed magic string that no redis command begins with, so matching it is unambiguous and has no false positives (no redis command starts with `PR…`).

- Added `RedisTest.inline_does_not_eat_h2_preface` (`test/brpc_redis_unittest.cpp`) covering: the full preface → `TRY_OTHERS`; a partial prefix → `TRY_OTHERS`; and a genuine inline command sharing the leading byte (`PING`) → `PARSE_OK`.

Side effects:

- Performance effects: negligible — one extra first-byte comparison per inline command, and a short `memcmp` only when it is `'P'`.

### DIFF
--- a/src/brpc/redis_command.cpp
+++ b/src/brpc/redis_command.cpp
@@ -408,6 +408,23 @@ RedisCommandConsumeState RedisCommandParser::ConsumeImpl(butil::IOBuf& buf,
             *err = PARSE_ERROR_TRY_OTHERS;
             return CONSUME_STATE_ERROR;
         }
+        // The HTTP/2 connection preface ("PRI * HTTP/2.0\r\n...") is alpha-leading
+        // and would otherwise be consumed as an inline redis command (first token
+        // "PRI"), preventing protocol auto-detection from falling through to
+        // HTTP/2 (e.g. gRPC clients). Defer to other protocols when the input
+        // matches the preface, either fully or as a not-yet-complete prefix. No
+        // valid redis command begins with these bytes, so this is unambiguous.
+        // See issue #3109.
+        static const char h2_preface[] = "PRI * HTTP/2.0\r\n";
+        const size_t h2_preface_len = sizeof(h2_preface) - 1;
+        if (*pfc == h2_preface[0]) {
+            char head[h2_preface_len];
+            const size_t n = buf.copy_to(head, h2_preface_len);
+            if (memcmp(head, h2_preface, n) == 0) {
+                *err = PARSE_ERROR_TRY_OTHERS;
+                return CONSUME_STATE_ERROR;
+            }
+        }
         const size_t buf_size = buf.size();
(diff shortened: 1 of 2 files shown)


wwbmmm:
LGTM

chenBright (approved):
LGTM

PR #3341, opened 2026-06-14
Add IOBuf<->std::iostream adapters for zero-copy I/O

chenBright:

Issue Number: resolve 

Problem Summary:

`IOBuf` integrates with protobuf via `IOBufAsZeroCopy{Input,Output}Stream`, but
there is no direct bridge to the `std::iostream` hierarchy. As a result, any
library that consumes/produces `std::istream&` / `std::ostream&` — most notably
`nlohmann::json` — cannot read from or write to an `IOBuf` without an
intermediate `std::string`:

```cpp
auto j = nlohmann::json::parse(req->request_attachment().to_string());
```

This is an obstacle to adopting nlohmann::json in bRPC.

Changed:

  Introduce four classes in `src/butil/iobuf.{h,cpp}` that bridge IOBuf and
  the standard iostream hierarchy:

  - `IOBufAsInputStreamBuf` — read-only `std::streambuf` over an IOBuf,
  walking the backing blocks via the public `backing_block()` API. Overrides
  `underflow`, `xsgetn`(bulk memcpy across block boundaries to avoid the
  default per-byte sbumpc loop), and `showmanyc` (saturating sum of
  remaining block sizes).
  - `IOBufInputStream` — `std::istream` view that wires the streambuf above.
  - `IOBufAsOutputStreamBuf` — append-only `std::streambuf` that delegates to
  the existing `IOBufAsZeroCopyOutputStream` for block allocation and the
  three-branch `BackUp` (including the shared-block degradation that protects
  IOBufs sharing blocks from being corrupted by an in-flight write). Overrides
  `overflow`, `xsputn` (block-sized memcpy instead of per-byte sputc),
  and `sync` (forwards to shrink()).
  - `IOBufOutputStream` — `std::ostream` view that wires the streambuf above;
  accepts an optional block_size to allocate dedicated blocks instead of
  drawing from the per-thread TLS pool.

  Usage with nlohmann::json:

```c++
  // Parse: zero string copy
  butil::IOBufInputStream in(request_body);
  auto j = nlohmann::json::parse(in);

  // Serialize: bytes flow straight into IOBuf blocks
  butil::IOBufOutputStream os(response_body);
  os << j;
```

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -44,6 +44,7 @@ bazel_dep(name = 'thrift', version = '0.21.0', repo_name = 'org_apache_thrift')
 
 # test only
 bazel_dep(name = "gperftools", version = "2.18.1", dev_dependency = True)
+bazel_dep(name = "nlohmann_json", version = "3.12.0", dev_dependency = True)
 bazel_dep(name = 'googletest', version = '1.14.0.bcr.1', repo_name = 'com_google_googletest', dev_dependency = True)
 bazel_dep(name = 'hedron_compile_commands', dev_dependency = True)
 git_override(
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -279,6 +279,24 @@ http_archive(
     urls = ["https://archive.apache.org/dist/thrift/0.15.0/thrift-0.15.0.tar.gz"],
 )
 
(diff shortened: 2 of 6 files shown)


wwbmmm:
LGTM

PR #3344, opened 2026-06-14
Limit nested container depth in Redis reply and AMF parsers

wwbmmm:

Issue Number: null

Problem Summary:

Redis replies and AMF payloads can contain nested container values. The parsers previously bounded per-container allocation sizes, but did not bound nesting depth. Very deep inputs could therefore consume excessive call stack while parsing.

Changed:

- Add `redis_max_reply_depth` to limit nested Redis array replies.
- Add `amf_max_depth` to limit nested AMF objects and arrays.
- Add regression tests for over-limit nested Redis replies and AMF strict arrays.

Side effects:
- Performance effects: negligible; parsing adds one integer depth check per nested container.
- Breaking backward compatibility: inputs with Redis reply or AMF nesting deeper than the configured limit are rejected.

---
### Check List:

### DIFF
--- a/src/brpc/amf.cpp
+++ b/src/brpc/amf.cpp
@@ -22,8 +22,20 @@
 #include "butil/find_cstr.h"
 #include "brpc/log.h"
 #include "brpc/amf.h"
+#include "gflags/gflags.h"
 
 namespace brpc {
+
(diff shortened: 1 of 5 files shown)


chenBright (approved):
LGTM

PR #3346, opened 2026-06-14
use -fno-access-control instead of -Dprivate=public for tests

### LINKED ISSUE #3345: -Dprivate=public in test CMakeLists.txt breaks std::any compilation with C++17
## Description

When building BRPC tests with C++17 (GCC 12+), compilation fails with:

```
error: 'struct std::any::_Manager_internal<_Tp>' redeclared with different access
```

The root cause is `-Dprivate=public -Dprotected=public` in `test/CMakeLists.txt` line 56, which is used to access private members in tests. This macro rewrites `private` to `public` in all headers, including standard library headers like `<any>`.

## Steps to Reproduce

```bash
# On GCC 12+ with C++17
g++ -std=c++17 -Dprivate=public -Dprotected=public -c test.cpp
# where test.cpp transitively includes <any> (e.g. via gtest-port.h)
```

The `<any>` header is pulled in by gtest's `gtest-port.h:2312`:
```cpp
#if __has_include(<any>) && __cplusplus >= 201703L
#include <any>
#endif
```

## Environment

- GCC 12+ (tested with GCC 13.3 on x86, GCC 15.1 on RISC-V)
- C++17 or later
- Any architecture (x86, ARM, RISC-V)

## Root Cause

`test/CMakeLists.txt` line 56:
```cmake
set(CMAKE_CPP_FLAGS "${CMAKE_CPP_FLAGS} ... -Dprivate=public -Dprotected=public -include ${PROJECT_SOURCE_DIR}/test/sstream_workaround.h")
```

The existing `sstream_workaround.h` handles the same issue for `<sstream>` by including it before the macro takes effect, but does not protect `<any>`.

## Why CI Does Not Catch This

BRPC's CMakeLists.txt hardcodes `-std=c++11` (line 155/158), so `__cplusplus` is `201103L` and gtest skips the `#include <any>`. The bug only manifests when building with C++17 explicitly or when the compiler defaults to C++17 (e.g. GCC 14+).

Felix-Gong:
## Summary

Replace `-Dprivate=public -Dprotected=public` macro hack with `-fno-access-control` compiler flag in CMake and Makefile test builds. The macro approach caused redeclaration errors with standard library headers (`<sstream>`, `<any>`, etc.) that use private access specifiers internally.

`-fno-access-control` is a GCC/Clang compiler flag that cleanly disables access control checking without affecting header parsing. The Bazel build already uses this flag at `test/BUILD.bazel:26`, so this makes CMake and Makefile builds consistent with Bazel.

Fixes #3345

## Changes

- `test/CMakeLists.txt`: Replace `-Dprivate=public -Dprotected=public -include sstream_workaround.h` with `-fno-access-control`
- `test/Makefile`: Same change
- `test/sstream_workaround.h`: Deleted (no longer referenced)

## Test plan

- Local x86 tests: 72/75 passed (3 failures are pre-existing: bthread_dispatcher_unittest epoll issue, test_butil StackTraceTest + EndPointTest)
- All core bthread and brpc tests pass


chenBright (commented):
I think using `-fno-access-control`, like in `test/BUILD.bazel`, would be more reasonable.

https://github.com/apache/brpc/blob/172589604ffbb5c970b9bcd3c1df0441a9614dd6/test/BUILD.bazel#L26

test/sstream_workaround.h
```diff
@@ -21,13 +21,22 @@
...
+// Similarly, std::any in C++17 uses private access specifiers internally that
+// would break with -Dprivate=public. Include it here before the macro takes effect.
 
 #ifdef private
 # undef private
 # include <sstream>
+# if __cplusplus >= 201703L
+#  include <any>
+# endif
 # define private public
```
chenBright: Using `-fno-access-control` instead of `# define private public`?

Felix-Gong:
Good point! `-fno-access-control` is indeed a cleaner approach than `#define private public`. It avoids the redeclaration issues with standard library headers entirely.

I'll update the PR to:
1. Replace `-Dprivate=public` with `-fno-access-control` in CMake build flags
2. Simplify `sstream_workaround.h` (the `#ifdef private` guard becomes unnecessary)

The Bazel build already uses `-fno-access-control` at [test/BUILD.bazel:26](https://github.com/apache/brpc/blob/172589604ffbb5c970b9bcd3c1df0441a9614dd6/test/BUILD.bazel#L26), so this makes CMake consistent with Bazel.

chenBright:
test/Makefile also needs to be updated.

https://github.com/apache/brpc/blob/172589604ffbb5c970b9bcd3c1df0441a9614dd6/test/Makefile#L21

test/sstream_workaround.h
```diff
@@ -18,16 +18,12 @@
...
-// __xfer_bufptrs’ redeclared with different access"
-
-#ifdef private
-# undef private
-# include <sstream>
-# define private public
-#else
-# include <sstream>
-#endif
+// Previously, tests used -Dprivate=public to access private members.
```
chenBright: The file test/sstream_workaround.h can be deleted, right?

chenBright:
These all need to be deleted.
https://github.com/apache/brpc/blob/f04e0e06051c21e3b83ca883c8b68a1a60f35784/test/BUILD.bazel#L146-L151
https://github.com/apache/brpc/blob/f04e0e06051c21e3b83ca883c8b68a1a60f35784/test/BUILD.bazel#L185
https://github.com/apache/brpc/blob/f04e0e06051c21e3b83ca883c8b68a1a60f35784/test/BUILD.bazel#L198
https://github.com/apache/brpc/blob/f04e0e06051c21e3b83ca883c8b68a1a60f35784/test/BUILD.bazel#L213
https://github.com/apache/brpc/blob/f04e0e06051c21e3b83ca883c8b68a1a60f35784/test/BUILD.bazel#L242

Felix-Gong:
> These all need to be deleted.

Yes, I'm removing all sstream_workaround references from test/BUILD.bazel now. The fix will be pushed shortly.

PR #3347, opened 2026-06-15
Make _state atomic to prevent concurrent _read_buf mutation on TCP fallback

chenBright:

Issue Number: resolve 

Problem Summary:

`RdmaEndpoint::_state` is a plain (non-atomic) enum, but it is written
by the RDMA handshake bthread and read concurrently by the
event dispatcher bthread in `OnNewDataFromTcp`. This is a data race, 
and  it can let the two threads concurrently mutate `_socket->_read_buf` 
and corrupt an `IOBuf`.

```
time   bthread A (OnNewDataFromTcp)                    bthread B (ProcessHandshakeAtServer)
────   ─────────────────────────                       ─────────────────────────────────────
 T0                                                    Program order:
                                                         s->_read_buf.append(magic, ...)
                                                         ep->_state = FALLBACK_TCP

       ═════════════════════════════ Actual order after CPU reordering ══════════════════════

 T1                                                    ep->_state = FALLBACK_TCP
                                                         ← hoisted: this store takes effect FIRST
                                                         (_nevent is still 0,
                                                          _read_buf still has no magic in it)

 T2   (new bytes arrive on the fd,
        epoll edge-trigger fires)
       Socket::StartInputEvent
         _nevent.fetch_add(acq_rel): 0 → 1, == 0
         → spawn bthread A: OnEdge → OnNewDataFromTcp

 T3   read ep->_state (plain read, no acquire)
       => observes FALLBACK_TCP ✓ (T1 is visible)

 T4   else if (_state == FALLBACK_TCP) ✓
       → InputMessenger::OnNewMessages(m)
         └─ DoRead →
              _read_buf.append_from_file_descriptor(fd, ...)
         ⚠ A is now WRITING into _read_buf

 T5                                                    s->_read_buf.append(magic, MAGIC_STR_LEN)
                                                         ← reordered: this append happens NOW
                                                         ⚠⚠⚠ A and B are concurrently mutating
                                                              the SAME IOBuf object!

                                                       Consequences:
                                                         • IOBuf internal linked list /
                                                           _ref_array / _data pointer / refcount
                                                           are mutated concurrently
                                                         • memory corruption, use-after-free,
                                                           refcount drift, crash
                                                         • or — more insidious — the bytes
                                                           appear to be accepted but the
                                                           internal state of _read_buf is
                                                           already corrupted

```

Changed:

Make `_state` a `butil::atomic<State>`.

- Terminal-state stores use `memory_order_release` and the corresponding loads use 
   `memory_order_acquire`, so that everything published before a terminal state becomes 
   visible to the reader:
   - the magic bytes appended back into _socket->_read_buf before FALLBACK_TCP;
   - the RDMA window/resource setup done before ESTABLISHED.
- All other (non-terminal) handshake transitions use `memory_order_relaxed`.

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/brpc/rdma/rdma_endpoint.cpp
+++ b/src/brpc/rdma/rdma_endpoint.cpp
@@ -161,7 +161,7 @@ RdmaEndpoint::~RdmaEndpoint() {
 void RdmaEndpoint::Reset() {
     DeallocateResources();
 
-    _state = UNINIT;
+    _state.store(UNINIT, butil::memory_order_relaxed);
     _resource = NULL;
     _send_cq_events = 0;
     _recv_cq_events = 0;
@@ -195,7 +195,8 @@ void RdmaConnect::StartConnect(const Socket* socket,
         return;
     }
     if (!IsRdmaAvailable()) {
-        rdma_transport->_rdma_ep->_state = RdmaEndpoint::FALLBACK_TCP;
+        rdma_transport->_rdma_ep->_state.store(RdmaEndpoint::FALLBACK_TCP,
+                                               butil::memory_order_relaxed);
         rdma_transport->_rdma_state = RdmaTransport::RDMA_OFF;
         done(0, data);
         return;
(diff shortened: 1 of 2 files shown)


yanglimingcn:
LGTM

PR #3348, opened 2026-06-15
Add socket reserve/use connection affinity + fd_version primitives

rajvarun77:
Generic `Controller`/`Socket` plumbing that lets a protocol pin a pooled/short connection across multiple RPCs (connection affinity). This is the shared substrate extracted from the MySQL client PR #3330 so it can be reviewed independently; #3330 is now stacked on top of this branch and contains only the MySQL-specific code.

### Controller
- `BindSockAction {NONE,RESERVE,USE}` encoded in two `Controller::_flags` bits (`FLAGS_BIND_SOCK_RESERVE`/`USE`) via `set_bind_sock_action()`/`bind_sock_action()` — no dedicated member added to the hot `Controller`.
- `_bind_sock` holds a socket reserved by a previous RPC. `EndRPC` propagates the action to the current `Call`; `Call::OnComplete` reserves the socket (on success) instead of returning it to the pool / failing it; `IssueRPC` reuses the reserved socket when the action is `USE`.
- `_session_data`: opaque per-RPC slot a protocol codec may use to carry typed state between serialize/pack/parse (not owned by `Controller`).

### Socket
- `_fd_version` (relaxed atomic, bumped on every `ResetFileDescriptor`) + `fd_version()` accessor — an ABA guard so an affinity holder can detect that a reserved fd was reconnected underneath it. Atomic because it is written on the reconnect path (no socket lock held) and read from other threads.
- `Socket::Write` accepts an empty buffer when `WriteOptions::auth_flags` is set, for server-greets-first handshakes that send their first bytes from the connection-phase handler rather than from `data`.

### Backup/retry fix
Also fixes an uninitialized read this machinery would otherwise introduce: the `Call(Call*)` copy constructor used for backup requests and retries did not initialize `bind_sock_action`, so `OnComplete` could read indeterminate bits and (when they matched `RESERVE`/`USE`) divert the backup call's socket away from the pool-return path, hanging the RPC (manifested as `ChannelTest.backup_request` timing out). Initialized to `BIND_SOCK_NONE`, matching `Call::Reset()`; a backup/retry never inherits affinity.

No protocol uses these primitives in this PR; the first consumer is MySQL (#3330).

### DIFF
--- a/src/brpc/controller.cpp
+++ b/src/brpc/controller.cpp
@@ -297,6 +297,9 @@ void Controller::ResetPods() {
     _request_streams.clear();
     _response_streams.clear();
     _remote_stream_settings = NULL;
+    set_bind_sock_action(BIND_SOCK_NONE);
+    _bind_sock.reset();
+    _session_data = NULL;
     _auth_flags = 0;
     _rpc_received_us = 0;
 }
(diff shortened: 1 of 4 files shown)


rajvarun77:
This is the shared-core substrate extracted from #3330 (MySQL client protocol), which is now stacked on top of this PR. Reviewing this first keeps the generic `Controller`/`Socket` change small and self-contained.

rajvarun77:
Reverting the split — folding the backup/retry copy-constructor fix back into #3330 directly; a separate core PR isn't warranted for a one-line initialization fix.

PR #3353, opened 2026-06-21
Limit AMF string and array parsing

wwbmmm:

Issue Number: null

Problem Summary:

AMF parsing accepted declared string lengths and array counts directly from the input stream. Malformed RTMP/AMF payloads could make the parser reserve excessive memory or spend a long time walking impossible element counts before discovering that the stream was incomplete.

Changed:

- Add configurable AMF string and array limits.
- Read AMF string bodies incrementally instead of resizing the destination string to the declared length up front.
- Reject ECMA array and strict array counts above the configured limit before parsing their elements.
- Add RTMP AMF unit tests for truncated long strings and oversized array counts.

Side effects:
- Performance effects: AMF string parsing now copies through a small stack buffer, avoiding large zero-fill allocations for malformed payloads.

- Breaking backward compatibility: AMF payloads whose string length or array count exceed the new defaults will be rejected. The defaults are 64 MiB per string and 1,048,576 array elements.

---
### Check List:

### DIFF
--- a/src/brpc/amf.cpp
+++ b/src/brpc/amf.cpp
@@ -27,11 +27,35 @@
 namespace brpc {
 
 DEFINE_int32(amf_max_depth, 128, "Maximum nesting depth for AMF objects and arrays");
+DEFINE_int32(amf_max_string_size, 64 * 1024 * 1024,
+             "Maximum byte size for AMF strings");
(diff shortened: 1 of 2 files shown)


chenBright (approved):
LGTM

PR #3354, opened 2026-06-21
Fix empty Redis command array parsing

wwbmmm:

Issue Number: null

Problem Summary:

The Redis command parser accepted an empty RESP array as the start of a command
and kept parsing subsequent bulk strings as array elements. Since an empty array
has no command name or arguments, this input should be rejected by the parser.

Changed:

- Reject empty RESP command arrays in `RedisCommandParser`.
- Replace the internal array index `CHECK` with an explicit parse error before
  writing parsed bulk strings into the argument vector.
- Add a parser regression test for empty RESP arrays followed by bulk strings.

Side effects:
- Performance effects: None expected.

- Breaking backward compatibility: Empty RESP arrays are now rejected as invalid
  Redis commands.

---
### Check List:


chenBright (approved):
LGTM

PR #3294, opened 2026-05-12
fix(selective_channel): isolate backup responses and add race regress…

altman08:
…ion test

### What problem does this PR solve?

Issue Number: resolve 

Problem Summary:
In SelectiveChannel backup-request flows, two concurrent subcalls may race on response lifecycle and produce inconsistent response data (e.g. unexpected repeated field sizes). In worse timing windows, this can lead to memory corruption and crash during protobuf cleanup (observed in RepeatedPtrFieldBase::DestroyProtos while clearing sub-responses).

### What is changed and the side effects?

Changed:

Side effects:
- Performance effects:

- Breaking backward compatibility: 

---
### Check List:
- Please make sure your changes are compilable.
- When providing us with a new feature, it is best to add related tests.
- Please follow [Contributor Covenant Code of Conduct](https://github.com/apache/brpc/blob/master/CODE_OF_CONDUCT.md).

PR #3359, opened 2026-06-25
Fix late selective_channel retry after EndRPC

### LINKED ISSUE #3358: selective_channel: late SubDone may re-enter retry/backup after _lb.reset() and crash on null ChannelBalancer
**Describe the bug**
We hit a crash in `selective_channel` on a path where a late `SubDone::Run()` re-enters the retry/backup flow after `Controller::_lb` has already been reset.

This report is self-contained on purpose because the original investigation notes are on an internal link that is not publicly accessible.

The crash is not a use-after-free of `Controller`, `Sender`, `SelectiveChannel`, or `SubDone` themselves. The immediate problem is that `_main_cntl->_lb` becomes null, while a late callback still goes into `IssueRPC()` and dereferences the null `ChannelBalancer*`.

### Stack shape

The stack we observed is effectively:

```text
SubDone::Run
  -> Controller::OnVersionedRPCReturned
    -> Controller::IssueRPC
      -> schan::Sender::IssueRPC
        -> static_cast<ChannelBalancer*>(_main_cntl->_lb.get())
          -> ChannelBalancer::SelectChannel(this=0x0)
            -> SharedLoadBalancer::SelectServer(this=0x0)
```

**To Reproduce**
Why this looks like a real race in current master
Based on current master behavior in controller.cpp and selective_channel.cpp:
Controller::EndRPC() resets _lb before _done->Run()
Controller::OnVersionedRPCReturned() may directly call IssueRPC() when it decides to continue retry/backup
Controller::IssueRPC() checks _sender, but does not check _lb
schan::Sender::IssueRPC() directly reads _main_cntl->_lb.get() and uses it as ChannelBalancer*

This creates a race window:
One sub-RPC finishes and drives the main RPC into EndRPC()
EndRPC() does _lb.reset()
Another sub-RPC finishes slightly later
Its SubDone::Run() still enters OnVersionedRPCReturned()
Retry/backup is still considered
IssueRPC() is called
Sender::IssueRPC() reads _main_cntl->_lb.get() == nullptr
Null dereference follows

**Trigger conditions**
The crash becomes much easier to hit when all of the following are true:
all sub-channels fail to establish connections
backup_request_ms > 0
max_retry > 0
multiple sub-RPCs fail or time out nearly at the same time
In our production case this happened under a network failure window where all candidate peers became unreachable, while backup requests and retries were both enabled.

**Related issues / PRs checked**
I checked the existing public reports and found related-but-different items:
#2454: DynamicPartitionChannel thrift协议使用backup request产生coredump
#1656: Crashed when retry many times
PR #3294: fix(selective_channel): isolate backup responses and add race regression test
Those appear to be different failure modes. In particular, PR #3294 looks like a backup response isolation fix, not a fix for _lb lifetime vs late SubDone retry/backup re-entry.

**Suggested fixes**
The lowest-risk defensive fix seems to be:
In schan::Sender::IssueRPC(), check _main_cntl->_lb.get() immediately after loading it, and fail gracefully if it is null

**Possible additional fixes**
In Controller::OnVersionedRPCReturned(), short-circuit retry/backup if _lb == nullptr
Reconsider whether _lb.reset() in Controller::EndRPC() should happen later, e.g. after _done->Run() or together with final teardown

**Expected behavior**
Late SubDone callbacks should not be able to re-enter retry/backup in a way that dereferences a null ChannelBalancer.

**Versions**
OS:rocky 9.1
Compiler:gcc 12
brpc: 1.17.0
protobuf: 

hjwsm1989:

Issue Number: #3358

`SelectiveChannel` has a use-after-teardown race: a late `SubDone::Run()` can
re-enter the retry/backup path **after** the main RPC has already run `EndRPC()`.

`EndRPC()` resets the controller's load balancer (`_lb`), but a sub-call that
finishes slightly later still flows through:

```
SubDone::Run
  -> Controller::OnVersionedRPCReturned()
     -> Controller::IssueRPC()
        -> schan::Sender::IssueRPC()   // reads _main_cntl->_lb, now null
```

so it retries on partially torn-down state — the null-balancer crash in #3358
(`SharedLoadBalancer::SelectServer(this=0x0)`). This is a distinct race from #3294.

Trigger: `backup_request_ms > 0`, `max_retry > 0`, and several sub-RPCs ending
almost simultaneously (timeout / connection failure) so a sub-call's callback
lands just after the main RPC's `EndRPC()`.

Changed:
- `EndRPC()` sets a new `FLAGS_ENDING_RPC` flag at its very start.
- `OnVersionedRPCReturned()` checks `is_ending_rpc()` and ignores late `SubDone`
  callbacks instead of re-entering retry/backup.
- `schan::Sender::IssueRPC()` keeps a defensive null check on the balancer as a
  last-resort guard at the `selective_channel` boundary.

Test:
Added a regression test using `SelectiveChannel` with backup request + retry that
lets the main RPC time out first and delayed sub-calls finish afterward —
reproducing the late-callback window and asserting it no longer re-enters
retry/backup. Re-ran the related selective / backup-request tests.

Side effects:
- Performance: none. The guard is a single flag check on the already-cold
  teardown path; the hot path is unchanged.

### Cluster validation

Also validated on a 3-node, production-like cluster issuing periodic
multi-address meta RPCs over `SelectiveChannel` (`backup_request_ms` set,
`max_retry=3`, short timeout). With the target on a remote node, one-way egress
delay (`tc netem delay 150ms`) was injected while the per-RPC timeout stayed at
50ms, so the main RPC enters `EndRPC()` while backup/retry sub-RPCs are still in
flight and their `SubDone::Run()` lands afterward.

The fix is silent by design (a late `SubDone` just hits the `is_ending_rpc()`
guard and is dropped), so to confirm the window was actually exercised I
temporarily added a `LOG(WARNING)` at the guard. Over a 4-minute injection it
fired **2137 times** with **0 crashes and 0 new coredumps**, and the cluster
recovered automatically once the delay was cleared. Without this fix each of
those late `SubDone`s would have re-entered retry/backup and dereferenced the
already-reset `_lb`. (The log was temporary instrumentation, not part of this PR.)

### Check List:
- [x] Changes are compilable.
- [x] Added a regression test for the new behavior.
- [x] Follows the Contributor Covenant Code of Conduct.

Fixes: #3358

### DIFF
--- a/src/brpc/selective_channel.cpp
+++ b/src/brpc/selective_channel.cpp
@@ -135,6 +135,7 @@ public:
     Sender(Controller* cntl,
            const google::protobuf::Message* request,
            google::protobuf::Message* response,
+           const butil::intrusive_ptr<SharedLoadBalancer>& lb,
            google::protobuf::Closure* user_done);
     ~Sender() { Clear(); }
     int IssueRPC(int64_t start_realtime_us);
@@ -148,6 +149,7 @@ private:
     Controller* _main_cntl;
     const google::protobuf::Message* _request;
     google::protobuf::Message* _response;
+    butil::intrusive_ptr<SharedLoadBalancer> _lb;
     google::protobuf::Closure* _user_done;
     short _nfree;
     short _nalloc;
@@ -293,10 +295,12 @@ void ChannelBalancer::Describe(std::ostream& os,
 Sender::Sender(Controller* cntl,
                const google::protobuf::Message* request,
                google::protobuf::Message* response,
+               const butil::intrusive_ptr<SharedLoadBalancer>& lb,
                google::protobuf::Closure* user_done)
     : _main_cntl(cntl)
     , _request(request)
     , _response(response)
+    , _lb(lb)
     , _user_done(user_done)
     , _nfree(0)
     , _nalloc(0)
@@ -306,14 +310,20 @@ Sender::Sender(Controller* cntl,
 
 int Sender::IssueRPC(int64_t start_realtime_us) {
     _main_cntl->_current_call.need_feedback = false;
+    ChannelBalancer* balancer =
+        static_cast<ChannelBalancer*>(_lb.get());
+    if (balancer == NULL) {
+        _main_cntl->SetFailed(ECANCELED,
+                              "SelectiveChannel balancer is unavailable");
+        return -1;
+    }
     LoadBalancer::SelectIn sel_in = { start_realtime_us,
                                       true,
                                       _main_cntl->has_request_code(),
                                       _main_cntl->_request_code,
                                       _main_cntl->_accessed };
     ChannelBalancer::SelectOut sel_out;
-    const int rc = static_cast<ChannelBalancer*>(_main_cntl->_lb.get())
-        ->SelectChannel(sel_in, &sel_out);
+    const int rc = balancer->SelectChannel(sel_in, &sel_out);
     if (rc != 0) {
         _main_cntl->SetFailed(rc, "Fail to select channel, %s", berror(rc));
         return -1;
@@ -424,6 +434,7 @@ void Sender::Clear() {
     _alloc_resources[1] = Resource();
     const CallId cid = _main_cntl->call_id();
     _main_cntl = NULL;
+    _lb.reset(NULL);
     if (_user_done) {
         _user_done->Run();
     }
@@ -570,8 +581,15 @@ void SelectiveChannel::CallMethod(
     if (!initialized()) {
         cntl->SetFailed(EINVAL, "SelectiveChannel=%p is not initialized yet",
                         this);
+        // This is a branch only entered by wrongly-used RPC, just call done
+        // in-place. See comments in channel.cpp on deadlock concerns.
+        if (user_done) {
+            user_done->Run();
+        }
+        return;
     }
-    schan::Sender* sndr = new schan::Sender(cntl, request, response, user_done);
+    schan::Sender* sndr =
+        new schan::Sender(cntl, request, response, _chan._lb, user_done);
     cntl->_sender = sndr;
     cntl->add_flag(Controller::FLAGS_DESTROY_CID_IN_DONE);
     const CallId cid = cntl->call_id();
--- a/src/brpc/controller.cpp
+++ b/src/brpc/controller.cpp
@@ -625,6 +625,16 @@ void Controller::OnVersionedRPCReturned(const CompletionInfo& info,
         return;
     }
 
+    if (is_ending_rpc()) {
+        // SelectiveChannel may still deliver late SubDone callbacks after the
+        // main RPC has entered EndRPC(). Ignore those callbacks instead of
+        // letting them re-enter retry/backup on partially torn-down state.
+        _error_code = saved_error;
+        response_attachment().clear();
+        CHECK_EQ(0, bthread_id_unlock(info.id));
+        return;
+    }
+
     if ((!_error_code && _retry_policy == NULL) ||
         _current_call.nretry >= _max_retry) {
         goto END_OF_RPC;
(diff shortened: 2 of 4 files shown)


hjwsm1989:
@chenBright hello, please help review

src/brpc/selective_channel.cpp
```diff
@@ -306,14 +306,20 @@ Sender::Sender(Controller* cntl,
 
 int Sender::IssueRPC(int64_t start_realtime_us) {
     _main_cntl->_current_call.need_feedback = false;
+    ChannelBalancer* balancer =
+        static_cast<ChannelBalancer*>(_main_cntl->_lb.get());
+    if (balancer == NULL) {
+        _main_cntl->SetFailed(ECANCELED,
+                              "SelectiveChannel balancer is unavailable");
+        return -1;
+    }
```
chenBright, line 319: The problem seems unresolved. The process still crashes even if the balancer is deleted by main_cntl after the if statement.

hjwsm1989: @chenBright thanks for review, The previous null check was not enough because it did not keep the balancer alive after `_lb.get()`.

I updated the patch so `Sender` holds its own `intrusive_ptr<SharedLoadBalancer>` captured from `SelectiveChannel` at creation time, and `IssueRPC()` uses that retained reference. This prevents `main_cntl->_lb.reset()` in `EndRPC()` from freeing the balancer while `IssueRPC()` is still using it.

The `FLAGS_ENDING_RPC` guard is kept to prevent late `SubDone` from re-entering retry / backup after the main RPC starts ending.

src/brpc/selective_channel.cpp
```diff
@@ -571,7 +581,8 @@ void SelectiveChannel::CallMethod(
         cntl->SetFailed(EINVAL, "SelectiveChannel=%p is not initialized yet",
                         this);
     }
-    schan::Sender* sndr = new schan::Sender(cntl, request, response, user_done);
+    schan::Sender* sndr =
```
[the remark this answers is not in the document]
hjwsm1989: Addressed. `SelectiveChannel::CallMethod()` now returns immediately when the channel is not initialized, and runs `done` in-place if provided, matching the behavior of `PartitionChannel`.

This avoids allocating `Sender` or dispatching into the inner channel with an unavailable balancer. I also added a regression test covering both sync and async uninitialized calls.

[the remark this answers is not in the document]
yanglimingcn: Since lb already belongs to a Sender, then in Sender::IssueRPC, _main_cntl->_lb.get() should be changed to _lb.get(), otherwise this part looks a bit confusing.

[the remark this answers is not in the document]
hjwsm1989: @yanglimingcn This has already been addressed in the latest patch. `Sender::IssueRPC()` now uses the `Sender`-owned `_lb.get()` and calls `SelectChannel()` through the local `balancer` pointer, instead of reading `_main_cntl->_lb.get()` directly.

hjwsm1989:
@chenBright @yanglimingcn comments updated，thanks for another review

src/brpc/selective_channel.cpp
```diff
@@ -135,6 +135,7 @@ friend class SubDone;
     Sender(Controller* cntl,
            const google::protobuf::Message* request,
            google::protobuf::Message* response,
+           SharedLoadBalancer* lb,
```
chenBright: The type of `lb` should be butil::intrusive_ptr<SharedLoadBalancer>, and the caller should pass in `_chan._lb`.

hjwsm1989: @chenBright Agreed. I changed the Sender constructor to take `const butil::intrusive_ptr<SharedLoadBalancer>&` and pass `_chan._lb` directly from `SelectiveChannel::CallMethod()`.

`Sender` still keeps its own `_lb` copy, and `IssueRPC()` continues to use `_lb.get()`. This makes the ownership/lifetime intent explicit.

chenBright (approved):
LGTM

yanglimingcn:
LGTM

PR #3363, opened 2026-06-28
Limit Redis inline command allocation

wwbmmm:

Issue Number: null

Problem Summary:

Redis inline command parsing copied the full input buffer before checking
whether a complete line had arrived. A long pending inline command could make
the parser repeatedly allocate and copy buffered data while waiting for CRLF.

Changed:
- Scan Redis inline commands for CRLF before allocating parser arena memory.
- Apply `redis_max_allocation_size` to the current inline command line.
- Copy only the current inline command line instead of the whole input buffer.
- Add regression tests for an oversized inline command without a trailing CRLF
  and for pipelined inline input with a large incomplete next command.

Side effects:
- Performance effects: Incomplete inline commands no longer allocate parser
  arena memory while waiting for CRLF, and complete inline commands only copy
  the current line.
- Breaking backward compatibility: Inline Redis commands longer than
  `redis_max_allocation_size` are now rejected.

---
### Check List:

### DIFF
--- a/src/brpc/redis_command.cpp
+++ b/src/brpc/redis_command.cpp
@@ -451,6 +451,14 @@ RedisCommandConsumeState RedisCommandParser::ConsumeImpl(butil::IOBuf& buf,
                 *err = PARSE_ERROR_NOT_ENOUGH_DATA;
                 return CONSUME_STATE_ERROR;
             }
+            if (buf.size() == max_inline_size + 1) {
+                char last_char = '\0';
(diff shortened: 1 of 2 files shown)


wwbmmm:
The CI failure seems unrelated to this code change. I have retried for 3 times, each time failed in different place. I think we can ignore this failure.

chenBright (approved):
LGTM

PR #3367, opened 2026-07-02
Add Power-of-Two-Choices Peak-EWMA load balancer (p2c)

### LINKED ISSUE #3340: [Feature] Add Power-of-Two-Choices Peak-EWMA load balancer (p2c) — O(1) tail-latency-aware policy

brpc has several client-side load balancers (`rr`, `wrr`, `random`, `wr`, `la`, the consistent-hash family, `_dynpart`), but only one of them is latency-aware: `la` (locality-aware). `la` maintains a doubly-buffered **weight tree over all servers** with weights driven by inverse average latency, and samples a single point on the cumulative-weight line.

brpc currently has **no Power-of-Two-Choices (P2C) sampling load balancer** — neither Envoy-style least-request (P2C over active-request-biased weights) nor Finagle-style Peak-EWMA. P2C is the single most widely deployed tail-latency balancer in modern RPC stacks and service meshes (Finagle/Twitter `p2cPeakEwma`, linkerd, Envoy `LEAST_REQUEST`), and it is absent from brpc.

This matters in two regimes that `rr`/`random` ignore and where `la`'s averaging window lags:
- a **single slow/degraded backend** (GC pause, noisy neighbor, cold cache) — should be shed within one sample, not after an averaging window;
- a **heterogeneous fleet** (mixed CPU generations/capacities) — selection should continuously bias toward faster, less-loaded nodes.

Add a new client-side load balancer, registered as **`p2c`** (alias `p2c_ewma`): **Power-of-Two-Choices with Peak-EWMA latency scoring**.

Algorithm (per request):
1. Sample **two distinct random backends** from the server list (`fast_rand_less_than`, as `rr`/`random` already do).
2. Route to the one with the **lower load score**:
   - `score = ewma_us * (inflight + 1) / max(weight, 1)`
   - `ewma_us` is a **peak-sensitive EWMA of round-trip latency** per backend: an upward latency spike sets the score to the spike value immediately; downward recovery decays with a configurable time constant `TAU` (Finagle default decay ≈ 10s). This is what sheds a stalled backend within one sample.
   - `inflight` is the backend's outstanding-request count (incremented at select, decremented at completion) — the Envoy `LEAST_REQUEST` `(active+1)` term.
   - dividing by the static naming-service `weight` makes it **weighted P2C** (degrades to unweighted when weights are equal).
3. On RPC completion, update the chosen backend's EWMA and decrement its in-flight; failures/timeouts inflate the latency (`max(measured, timeout)`) so a failing node is avoided — mirroring `la`'s error handling.

Properties: **O(1) selection** (two score evaluations regardless of fleet size, vs `la`'s O(log N) tree walk); spike-reactive via the peak term; herding-resistant via two-sample comparison.

**Interface feasibility — no brpc core change required.** Every hook P2C-EWMA needs already exists and is exercised by `la`:
- `src/brpc/load_balancer.h` — `SelectOut::need_feedback` (load_balancer.h:50), `virtual void Feedback(const CallInfo&)` (load_balancer.h:100), and `CallInfo { begin_time_us; server_id; error_code; controller; }` (load_balancer.h:53-61) giving per-call latency, which backend, success/failure, and timeout.
- `src/brpc/controller.cpp:873-876` — the RPC completion path actually invokes the hook:
  ```cpp
  if (need_feedback && c->_lb) {
      const LoadBalancer::CallInfo info =
          { begin_time_us, peer_id, error_code, c };
      c->_lb->Feedback(info);
  }
  ```
  and `controller.cpp:1120` propagates `sel_out.need_feedback` into the call.
- Membership lives in `DoublyBufferedData<Servers, TLS>` (lock-free reads, copy-on-write modify) exactly as `round_robin_load_balancer.cpp`. Mutable per-node load state (`inflight`, `ewma_us`, `stamp_us`) is heap-allocated once per backend and referenced by stable pointer — the same technique `locality_aware_load_balancer` already uses for its per-node atomics.
- brpc's `Socket` has no general app-level in-flight counter, so — like `la` — the LB owns its in-flight counters. Fully feasible.

**Files to add / modify**
- Add `src/brpc/policy/p2c_ewma_load_balancer.{h,cpp}`.
- Modify `src/brpc/global.cpp`: include the header, add a `P2CEwmaLoadBalancer` member to `GlobalExtensions`, and register it next to the existing balancers — `LoadBalancerExtension()->RegisterOrDie("p2c", &g_ext->p2c_ewma_lb);`.
- Add `test/brpc_p2c_ewma_load_balancer_unittest.cpp` (modeled on the `la`/`rr` LB unit tests), registered in all three build systems (`test/CMakeLists.txt`, `test/BUILD.bazel`, `test/Makefile`).
- Add a `p2c` row to `docs/cn/lb.md` / `docs/en/lb.md`.

**Performance / testing plan (multi-protocol, multi-config)**

A cross-protocol benchmark matrix to prove the gain is real on actual brpc-supported backends, not just a synthetic echo server.
- **Protocols:** baidu_std (protobuf RPC, controllable-sleep echo server — clean baseline), **Memcached** (`MemcacheRequest` — tiny ops, very high QPS, exposes selection overhead/herding), **Redis** (`RedisRequest` — single-threaded backend, high load sensitivity), **MySQL** (variable query cost — biggest tail surface), HTTP. Integration backends spawn-or-skip using the existing `brpc_redis_unittest.cpp` `system("which ...")` precedent, with one node acting as the **degraded** backend.
- **Configs:** fleet size N ∈ {4, 16, 64, 256}; homogeneous vs heterogeneous latency; degradation injected (none / 1 node +50ms / 10% of fleet +20ms / transient spike-then-recover); offered load (low / knee / saturating); equal vs 1:2:4 naming-service weights; pooled vs single connection.
- **Functional tests (must pass first):** add/remove/batch, single-server, `ExcludedServers` exclusion, feedback-driven score shift, in-flight inc/dec across select→feedback (incl. error/timeout path), weighted 1:2:4 split, and concurrency under TSan/ASan.
- **Metrics:** p50/p90/p99/p999/max latency; throughput at fixed SLO and saturating; per-node traffic share over time (divert latency after a spike); request-count variance (fairness/herding); selection CPU cost per op as N grows (P2C O(1) vs `la` O(log N)).
- **Headline result to validate:** tail-latency reduction and throughput-at-SLO uplift of `p2c` over `rr`/`random`/`la` on the heterogeneous and degraded configs, with selection cost no worse than `rr`.

- **Extend `la`** — different algorithm (global weight tree, averaging window); does not give O(1) sampling or peak spike-reactivity. Keeping both gives users a real choice.
- **Envoy-style least-request only** (`weight / (active+1)`) — a strict subset; P2C-EWMA folds the `(inflight+1)` active-request term *and* the peak-latency term into one score, so it covers least-request as the equal-latency special case.
- **WRR + ORCA load reports (gRPC gRFC A58)** — requires a server-reported load backchannel brpc does not have; out of scope for a client-only change.

**Additional context**

References (primary sources):
- Envoy — Supported load balancers (LEAST_REQUEST = P2C, `weight = lb_weight / (active+1)^bias`, "resistance to herding"): https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/load_balancers.html
- Envoy — LEAST_REQUEST can pick the busier host (P2C rationale): https://github.com/envoyproxy/envoy/issues/14859
- Finagle — Aperture Load Balancers (`p2cPeakEwma`; Peak-EWMA = moving average over RTT highly sensitive to peaks, weighted by outstanding requests): https://twitter.github.io/finagle/guide/ApertureLoadBalancers.html
- Finagle — `Balancers` API (`p2cPeakEwma` factory): https://twitter.github.io/finagle/docs/com/twitter/finagle/loadbalancer/Balancers$.html
- linkerd — "Beyond Round Robin: Load Balancing for Latency" (Peak-EWMA; p99.9 wins over RR/least-loaded): https://linkerd.io/2016/03/16/beyond-round-robin-load-balancing-for-latency/
- gRPC — gRFC A58 Client-Side Weighted Round Robin (ORCA load reports): https://github.com/grpc/proposal/blob/master/A58-client-side-weighted-round-robin-lb-policy.md

I'm happy to implement this end-to-end (balancer + unit tests in all three build systems + docs + the benchmark matrix). Would the maintainers be open to a `p2c` policy, and is a GitHub issue the right place to scope it, or would you prefer a `[DISCUSS]` thread on `dev@brpc.apache.org` first?

cc @zyearn (load-balancer policy owner / author of `la`) — would value your read on whether a `p2c` policy fits brpc.

rajvarun77:

Issue Number: resolve #3340

Problem Summary: brpc has no Power-of-Two-Choices load balancer — the most widely deployed tail-latency-aware policy (Envoy LEAST_REQUEST, Finagle/linkerd Peak-EWMA). A single degraded backend keeps receiving 25% of `rr` traffic until humans intervene; `la`'s averaging window reacts slower than one observation.

Changed: New `p2c` policy (`src/brpc/policy/p2c_ewma_load_balancer.{h,cpp}`): each selection samples two random servers (`p2c:choices=N` widens; evaluates all when N ≥ cluster size, with random tie-breaking to avoid herding) and routes to the lower `peak_ewma_latency_us * (inflight + 1) / weight`. Latency spikes replace the average immediately; recovery decays over `tau_ms` (default 10s); failures are punished with at least the RPC timeout. Uses only existing hooks (`SelectServer`/`Feedback`, `DoublyBufferedData` membership like `rr`); per-node stats are stable pointers shared by both buffers as in `la`. Registered in `global.cpp`; 10 unit tests; `p2c` docs in `docs/{cn,en}/client.md`.

Benchmark (rpc_press, 4 echo backends, slow = +5ms `bthread_usleep`, qps=4000, 15s, 50 threads, 3 reps averaged, zero errors; p99/p999 µs):

| scenario | `la` | `p2c` (best-of-2) | `p2c:choices=4` | `rr` |
|---|---|---|---|---|
| all healthy | 450 / 1042 | 418 / 768 | 435 / 1136 | 567 / 12833 |
| 1 slow of 4 | 384 / 771 | 2451 / 6460 | 399 / 545 | 6440 / 6511 |
| 2 slow of 4 | 1045 / 2709 | 6458 / 6779 | 4284 / 6486 | 6474 / 6542 |

Slow-node traffic share (1 slow of 4): `rr` 25%, `p2c` 1.3%, `p2c:choices=4` 0.6%, `la` 0.0% — consistent with the numbers on #3340.

Side effects:
- Performance effects: none on existing policies; `p2c` selection is O(1) (two score evaluations).
- Breaking backward compatibility: no.

---
### Check List:
- Compiles in CMake/Bazel/Makefile (sources and tests are picked up by the existing globs); `brpc_p2c_ewma_load_balancer_unittest` passes 10/10 and `brpc_load_balancer_unittest` passes 16/16.

cc @chenBright (thanks for the green light on #3340!) @zyearn

### DIFF
--- a/src/brpc/policy/p2c_ewma_load_balancer.cpp
+++ b/src/brpc/policy/p2c_ewma_load_balancer.cpp
@@ -163,7 +163,10 @@ int P2CEwmaLoadBalancer::SelectServer(const SelectIn& in, SelectOut* out) {
     if (n == 0) {
         return ENODATA;
     }
-    const int64_t now_us = butil::gettimeofday_us();
+    // Reuse the caller-provided timestamp to avoid an extra clock read per
+    // selection; only Channel::CheckHealth passes 0.
+    const int64_t now_us = in.begin_time_us > 0
+        ? in.begin_time_us : butil::gettimeofday_us();
 
     const ServerInfo* best = NULL;
     double best_score = 0;
--- a/src/brpc/policy/p2c_ewma_load_balancer.cpp
+++ b/src/brpc/policy/p2c_ewma_load_balancer.cpp
@@ -17,6 +17,7 @@
 
 
 #include <cmath>                                       // std::exp
+#include <gflags/gflags.h>
 #include "butil/fast_rand.h"                           // fast_rand_less_than
 #include "butil/time.h"                                // gettimeofday_us
 #include "butil/string_splitter.h"                     // KeyValuePairsSplitter
@@ -28,11 +29,22 @@
 namespace brpc {
 namespace policy {
 
+DEFINE_uint32(p2c_default_choices, 2,
+              "Default number of servers sampled per selection in p2c, "
+              "overridable per channel with the `choices' parameter");
+DEFINE_int64(p2c_default_tau_ms, 10000,
+             "Default decay time(ms) of the peak-EWMA latency in p2c "
+             "(Finagle's PeakEwma default), overridable per channel with "
+             "the `tau_ms' parameter");
+DEFINE_int64(p2c_max_punish_ms, 30000,
+             "Cap(ms) on the punished latency recorded for a failed call in "
+             "p2c, bounding how long a persistently failing server takes to "
+             "recover after it turns healthy. 0 means no cap");
+
 namespace {
-const uint32_t DEFAULT_CHOICES = 2;
+// Upper bound of the `choices' parameter. Compile-time because it sizes the
+// sampling array on SelectServer's stack.
 const uint32_t MAX_CHOICES = 64;
-// Finagle's PeakEwma default decay time.
-const int64_t DEFAULT_TAU_US = 10 * 1000000L;
 // Floor of the latency term so that in-flight counts break ties between
 // servers that have no latency observation yet.
 const double MIN_LATENCY_TERM_US = 1.0;
@@ -51,8 +63,9 @@ uint32_t WeightOfTag(const std::string& tag) {
 }  // namespace
 
 P2CEwmaLoadBalancer::P2CEwmaLoadBalancer()
-    : _choices(DEFAULT_CHOICES)
-    , _tau_us(DEFAULT_TAU_US) {}
+    // Clamp so that broken flag values can not disable comparison or decay.
+    : _choices(std::min(std::max(FLAGS_p2c_default_choices, 2u), MAX_CHOICES))
+    , _tau_us(std::max<int64_t>(FLAGS_p2c_default_tau_ms, 1) * 1000L) {}
 
 bool P2CEwmaLoadBalancer::Add(Servers& bg, const Servers& fg,
                               const ServerId& id) {
@@ -277,6 +290,13 @@ void P2CEwmaLoadBalancer::Feedback(const CallInfo& info) {
         latency_us = std::max(latency_us, timeout_us);
         latency_us = std::max(
             latency_us, 2 * stat->ewma_us.load(butil::memory_order_relaxed));
+        // Cap the punishment: consecutive failures double the EWMA each
+        // time, which otherwise grows without bound and delays recovery
+        // after the server turns healthy.
+        const int64_t max_punish_us = FLAGS_p2c_max_punish_ms * 1000L;
+        if (max_punish_us > 0 && latency_us > max_punish_us) {
+            latency_us = max_punish_us;
+        }
     }
 
     BAIDU_SCOPED_LOCK(stat->update_mutex);
--- a/src/brpc/policy/p2c_ewma_load_balancer.cpp
+++ b/src/brpc/policy/p2c_ewma_load_balancer.cpp
@@ -1,0 +1,366 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+
+#include <cmath>                                       // std::exp
+#include "butil/fast_rand.h"                           // fast_rand_less_than
+#include "butil/time.h"                                // gettimeofday_us
+#include "butil/string_splitter.h"                     // KeyValuePairsSplitter
+#include "butil/strings/string_number_conversions.h"   // StringToUint
+#include "brpc/socket.h"
+#include "brpc/controller.h"
+#include "brpc/policy/p2c_ewma_load_balancer.h"
+
+namespace brpc {
+namespace policy {
+
+namespace {
+const uint32_t DEFAULT_CHOICES = 2;
+const uint32_t MAX_CHOICES = 64;
+// Finagle's PeakEwma default decay time.
+const int64_t DEFAULT_TAU_US = 10 * 1000000L;
+// Floor of the latency term so that in-flight counts break ties between
+// servers that have no latency observation yet.
+const double MIN_LATENCY_TERM_US = 1.0;
+
+uint32_t WeightOfTag(const std::string& tag) {
+    if (tag.empty()) {
+        return 1;
+    }
+    uint32_t weight = 0;
+    if (!butil::StringToUint(tag, &weight) || weight == 0) {
+        LOG(WARNING) << "Invalid weight tag=`" << tag << "', use weight=1";
+        return 1;
+    }
+    return weight;
+}
+}  // namespace
+
+P2CEwmaLoadBalancer::P2CEwmaLoadBalancer()
+    : _choices(DEFAULT_CHOICES)
+    , _tau_us(DEFAULT_TAU_US) {}
+
+bool P2CEwmaLoadBalancer::Add(Servers& bg, const Servers& fg,
+                              const ServerId& id) {
+    if (bg.server_list.capacity() < 128) {
+        bg.server_list.reserve(128);
+    }
+    if (bg.server_map.seek(id.id) != NULL) {
+        return false;
+    }
+    ServerInfo info = { id.id, WeightOfTag(id.tag), NULL };
+    const size_t* pindex = fg.server_map.seek(id.id);
+    if (pindex == NULL) {
+        // Both buffers do not have the server. Create the stat structure
+        // which will be shared by both buffers.
+        info.stat = std::make_shared<NodeStat>();
+    } else {
+        // Already added to the other buffer, share its stat.
+        info.stat = fg.server_list[*pindex].stat;
+    }
+    bg.server_map[id.id] = bg.server_list.size();
+    bg.server_list.push_back(info);
+    return true;
+}
+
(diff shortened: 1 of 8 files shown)


src/brpc/policy/p2c_ewma_load_balancer.cpp
```diff
@@ -0,0 +1,366 @@
...
+int P2CEwmaLoadBalancer::SelectServer(const SelectIn& in, SelectOut* out) {
+    butil::DoublyBufferedData<Servers>::ScopedPtr s;
+    if (_db_servers.Read(&s) != 0) {
+        return ENOMEM;
+    }
+    const size_t n = s->server_list.size();
+    if (n == 0) {
+        return ENODATA;
+    }
+    const int64_t now_us = butil::gettimeofday_us();
```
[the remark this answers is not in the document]
rajvarun77: Good catch — now reusing `in.begin_time_us` when non-zero, falling back to `gettimeofday_us()` only for call sites that pass 0 (e.g. `Channel::CheckHealth`). Fixed in 1894ef1b.

src/brpc/policy/p2c_ewma_load_balancer.cpp
```diff
@@ -0,0 +1,369 @@
...
+namespace policy {
+
+namespace {
+const uint32_t DEFAULT_CHOICES = 2;
+const uint32_t MAX_CHOICES = 64;
+// Finagle's PeakEwma default decay time.
+const int64_t DEFAULT_TAU_US = 10 * 1000000L;
+// Floor of the latency term so that in-flight counts break ties between
+// servers that have no latency observation yet.
+const double MIN_LATENCY_TERM_US = 1.0;
```
chenBright: These variables should be made configurable using gflag.

src/brpc/policy/p2c_ewma_load_balancer.h
```diff
@@ -0,0 +1,100 @@
...
+    struct Servers {
+        std::vector<ServerInfo> server_list;
+        // Maps SocketId to index in server_list.
+        butil::FlatMap<SocketId, size_t> server_map;
+
+        Servers() {
+            if (server_map.init(1024, 70) != 0) {
+                LOG(WARNING) << "Fail to init server_map";
+            }
+        }
```
chenBright: No need to initialize the map.

src/brpc/policy/p2c_ewma_load_balancer.cpp
```diff
@@ -0,0 +1,369 @@
...
+        // time skews, ignore the sample.
+        return;
+    }
+    if (info.error_code != 0) {
+        // Punish failures with at least the timeout(if any) and twice the
+        // current average, so that a failing server keeps losing comparisons
+        // even when it fails fast.
+        const int64_t timeout_us = info.controller->timeout_ms() * 1000L;
+        latency_us = std::max(latency_us, timeout_us);
+        latency_us = std::max(
```
chenBright, line 291: It seems that persistent RPC errors will cause ewma_us to keep growing. Could this lead to slow recovery?

src/brpc/policy/p2c_ewma_load_balancer.cpp
```diff
@@ -0,0 +1,369 @@
...
+            latency_us, 2 * stat->ewma_us.load(butil::memory_order_relaxed));
+    }
+
+    BAIDU_SCOPED_LOCK(stat->update_mutex);
```
chenBright, line 302: Please test the overhead of the lock.

src/brpc/policy/p2c_ewma_load_balancer.cpp
rajvarun77: Done — the choices and tau_ms defaults are now gflags (`p2c_default_choices`, `p2c_default_tau_ms`), still overridable per channel via the lb parameter string. `MAX_CHOICES` stays compile-time since it sizes the on-stack sampling array, and `MIN_LATENCY_TERM_US` is only a tie-breaking epsilon — happy to make those flags too if you'd prefer.

src/brpc/policy/p2c_ewma_load_balancer.h
rajvarun77: Done — removed the `init()`; `server_map` now relies on FlatMap's small-map auto-initialization. (The explicit init was copied from LALB's older pattern.)

src/brpc/policy/p2c_ewma_load_balancer.cpp
rajvarun77: Good catch — this was real. The 2x punishment compounds under persistent errors, and since every feedback refreshes the decay stamp, no decay happens while the errors continue, so recovery time grew with outage length. Added `FLAGS_p2c_max_punish_ms` (default 30s, 0 = no cap) capping the punished sample, which bounds recovery to roughly `tau * ln(cap / healthy latency)` — a few tau in practice. New regression test `error_punish_is_capped` feeds 64 consecutive errors and asserts the EWMA stays within the cap.

src/brpc/policy/p2c_ewma_load_balancer.cpp
rajvarun77: Measured with the new `feedback_lock_overhead` test, all threads feeding back to a single server so its `update_mutex` sees worst-case contention: 265 ns/op at 1 thread, ~700 ns/op at 4 threads (about 5.7M feedbacks/s on one node). The mutex is per-server and only taken on the feedback path — selection stays lock-free — and real traffic spreads feedbacks across all servers, so per-node contention stays far below this worst case.

rajvarun77:
@chenBright thanks for the review — all four comments are addressed in the latest commit: choices/tau defaults are now gflags, error punishment is capped by `p2c_max_punish_ms` so recovery is bounded, the FlatMap init is removed, and lock-overhead numbers are in the thread (265 ns/op uncontended, ~700 ns/op worst case). CI is green — PTAL, and happy to iterate further.

wwbmmm:
LGTM

chenBright (approved):
LGTM

PR #3371, opened 2026-07-04
fix out-of-bounds read parsing AVC SPS in ParseSPS

ubeddulla:

Issue Number: N/A

Problem Summary:

`AVCDecoderConfigurationRecord::Create` copies the sequence header into a buffer that is not NUL-terminated (`DEFINE_SMALL_ARRAY`), then calls `ParseSPS(buf.data() + 2, sps_length)`. `ParseSPS` takes a `butil::StringPiece`, so the bare `const char*` is turned into one through the implicit `StringPiece(const char*)` constructor, which runs `strlen`. When a publisher sends an AVC sequence header whose SPS body has no zero byte, that `strlen` walks off the end of the copied record and reads out of bounds. The path is reachable from untrusted RTMP video input.

Changed:

Pass `buf.substr(2, sps_length)` so `ParseSPS` receives an explicitly sized view and no `strlen` runs. It is the same idiom already used a few lines below when pushing into `sps_list`. I also added a regression test that feeds a zero-free SPS body, which trips ASan on the old code and stays clean after the fix.

Side effects:
- Performance effects: none.

- Breaking backward compatibility: none. The reads inside `ParseSPS` were already bounded by `sps_length`, so valid sequence headers parse exactly as before.

---
### Check List:

### DIFF
--- a/src/brpc/rtmp.cpp
+++ b/src/brpc/rtmp.cpp
@@ -612,7 +612,7 @@ butil::Status AVCDecoderConfigurationRecord::Create(const void* data, size_t len
             return butil::Status(EINVAL, "Not enough data to decode SPS");
         }
         if (sps_length > 0) {
-            butil::Status st = ParseSPS(buf.data() + 2, sps_length);
+            butil::Status st = ParseSPS(buf.substr(2, sps_length), sps_length);
             if (!st.ok()) {
                 return st;
             }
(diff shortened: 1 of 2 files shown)


wasphin:
LGTM

chenBright (approved):
LGTM

PR #3377, opened 2026-07-08
build(cmake): Modernize target-scoped CMake configuration

darion-yaphet:

  Issue Number: N/A

  Problem Summary:
  The CMake build files still rely on legacy global include directories, compile flags, and duplicated per-example configuration. This makes target behavior harder to
  reason about and causes standalone examples/tests to drift from the main build configuration.

  Changed:
  - Modernize top-level, `src`, `tools`, and `test` CMake targets with target-scoped configuration.
  - Add shared `brpc_common_config` for common compile definitions, include paths, options, and features.
  - Refactor standalone example CMake files to use `example/cmake/BrpcExample.cmake`.
  - Update the gtest download helper to use CMake `3.16...3.28`.
  - Ensure downloaded gtest headers take priority over Homebrew's C++17-only gtest headers during unit-test builds.

  Side effects:
  - Standalone examples now consistently require CMake 3.16 or newer, matching the main project requirement.
  - `DOWNLOAD_GTEST=ON` creates a copied gtest include directory in the test build tree to avoid header conflicts with system-installed gtest.

  - Performance effects:
  No runtime performance impact expected. This is a build-system-only change.

  No source/API compatibility break expected. Users building standalone examples with CMake older than 3.16 will need to upgrade CMake.

  ---
  ### Check List:
  - [ ] When providing us with a new feature, it is best to add related tests.

  ### Verification

  Passed:
  - `cmake -S . -B build -DOPENSSL_ROOT_DIR=/opt/homebrew/opt/openssl@3`
  - `cmake --build build -j6`
  - `cmake -S . -B build-tests -DOPENSSL_ROOT_DIR=/opt/homebrew/opt/openssl@3 -DBUILD_UNIT_TESTS=ON -DDOWNLOAD_GTEST=ON`
  - `cmake --build build-tests -j6`
  - Standalone example smoke builds for `echo_c++`, `http_c++`, and `memcache_c++`

  Known macOS arm64 test issues:
  - `test_butil` fails in `StackTraceTest.find_symbol`
  - `bthread_fd_unittest` fails in `FDTest.ping_pong`
  - `rdma_performance` standalone example was not tested because `ibverbs` is unavailable on macOS

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.16...3.28)
 project(brpc C CXX)
 
 option(WITH_GLOG "With glog" OFF)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,7 +51,7 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
     if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 4.8)
         message(FATAL_ERROR "GCC is too old, please install a newer version supporting C++11")
     endif()
-elseif(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+elseif(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     # require at least clang 3.3
     if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 3.3)
         message(FATAL_ERROR "Clang is too old, please install a newer version supporting C++11")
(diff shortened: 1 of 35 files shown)


CMakeLists.txt
```diff
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-cmake_minimum_required(VERSION 2.8.12)
```
chenBright, line 18: The upgraded version seems too high, is it necessary?

darion-yaphet: > The upgraded version seems too high, is it necessary?

I understand the concern about compatibility, but CMake 3.16 is actually not very radical. CMake 2.x was quite an old era. The main line has moved to 4.x now, and CMake 3 is the mainstream baseline for many years past. 

The main benefits of upgrading to CMake 3 are not "keeping up with the latest trends", but rather the ability to use modern CMake syntax: target-based configuration, target_compile_features, target_link_options, generator expressions, better dependency discovery and cross-platform behavior. These features can avoid the pollution of global CFLAGS/LDFLAGS, allowing includes, macro definitions, and link parameters to be propagated by target. This makes the project easier to maintain and less likely to fail due to differences in the local environment. 

Overall, 3.16 represents a relatively conservative modern baseline, not a radical upgrade.

chenBright (approved):
LGTM

arrowd:
FreeBSD is going to jump to CMake 4 in the coming months, so it'd be great to get a new release of brpc containing this change. Otherwise, the brpc package we have will get broken.

chenBright:
> FreeBSD is going to jump to CMake 4 in the coming months, so it'd be great to get a new release of brpc containing this change. Otherwise, the brpc package we have will get broken.

@arrowd We plan to release a new version next month.

PR #3382, opened 2026-07-12
Enforce body size limits in protocol parsers

wwbmmm:

Issue Number: null

Problem Summary:

Some HTTP, memcache, and couchbase parsing paths did not consistently enforce
the global `max_body_size` setting. Non-progressive HTTP messages could buffer
more data than configured, while memcache and couchbase parsers could wait for
an oversized body declared in a response header.

Changed:

- Track the accumulated body size of non-progressive HTTP messages and reject
  data before appending beyond `max_body_size`.
- Return HTTP 413 for oversized HTTP/1 requests and close the connection after
  writing the response.
- Reset only the oversized HTTP/2 stream with `ENHANCE_YOUR_CALM`.
- Reject oversized memcache and couchbase responses immediately after parsing
  their fixed-size headers.
- Add regression tests for fixed-length HTTP, chunked HTTP, memcache, and
  couchbase messages.

Side effects:

- Performance effects: Each non-progressive HTTP body block adds one
  overflow-safe size comparison and addition. Memcache and couchbase add one
  comparison per response. No additional scan, allocation, or copy is added.
- Breaking backward compatibility: Messages larger than `max_body_size` are
  now rejected as documented. Progressive HTTP reading remains unchanged so
  existing bounded-memory streaming transfers can exceed this setting.

### Check List:

- [x] The changes are compilable.
- [x] Regression tests are included.
- [x] The affected test suites pass.

### DIFF
--- a/src/brpc/details/http_message.cpp
+++ b/src/brpc/details/http_message.cpp
@@ -39,6 +39,7 @@ DEFINE_bool(http_verbose, false,
 DEFINE_int32(http_verbose_max_body_length, 512,
              "[DEBUG] Max body length printed when -http_verbose is on");
 DECLARE_int64(socket_max_unwritten_bytes);
+DECLARE_uint64(max_body_size);
 
 // Implement callbacks for http parser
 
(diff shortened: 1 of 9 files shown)


chenBright (approved):
LGTM

PR #3384, opened 2026-07-13
Reclaim unscheduled timer tasks instead of holding slots until run_time

### LINKED ISSUE #3383: TimerThread: unscheduled tasks are not reclaimed until run_time, wasting memory with large timeouts
**Describe the bug**

When we set a very large timeout, the memory usage is always too high.
Image

bthread::TimerThread only reclaims a task's pooled Task slot when the task is finally popped from its internal min-heap at its run_time. unschedule() merely marks the task as removed (bumps its version)
  and intentionally does not return the slot — reclamation is deferred to the timer thread. As a result, a task that is cancelled early still occupies its slot for a long time, and the live Task count
  grows to roughly qps * timeout even though almost all of those tasks were unscheduled long ago.

**To Reproduce**

  1. Start a TimerThread.
  2. Repeatedly schedule() tasks with a large abstime (e.g. seconds_from_now(3600)) and unschedule() them shortly after — mimicking RPC deadline timers that are almost always cancelled before firing.
  3. Watch the number of allocated Task objects (ResourcePool<Task>) / process RSS.

**Expected behavior**

**Versions**
OS:
Compiler:
brpc:
protobuf:

stdpain:

Issue Number: #3383 

Problem Summary:

 bthread::TimerThread reclaims a task's pooled Task slot only when the task is popped from the internal min-heap at its run_time. unschedule() just marks the task removed and does not free the slot. So a
  cancelled timer keeps its slot until run_time, and the live Task count grows to ~ qps * timeout even though almost all of those tasks were unscheduled long ago. Two retention paths:

  1. Heap retention — a task pulled into the heap is only checked for cancellation once (via try_delete() at consume time). If unscheduled afterwards, it lingers in the heap until run_time.
  2. Bucket accumulation — the timer thread only consumes buckets when it wakes, and schedule() wakes it only for an earlier task. If every pending task is far in the future, scheduled-then-unscheduled
  tasks pile up in the buckets, unconsumed and unreclaimed, until that far deadline.

  This is significant for RPC deadline timers with large timeouts (e.g. 100k qps × 300s ≈ ~1.9 GB held mostly by already-cancelled timers).

Changed:

  - Heap sweep: in TimerThread::run(), after consuming buckets, drop unscheduled tasks from the heap. Triggered by a timer-thread-local heuristic (heap ≥ brpc_timer_heap_sweep_min_size, default 4096, and
  roughly doubled since the last sweep), so the unschedule() hot path is untouched (no new shared atomics). Amortized O(1) per task.
  - Periodic wakeup: cap the sleep at brpc_timer_max_wakeup_interval_ms (default 0 = disabled, legacy behavior) so the thread periodically wakes to drain buckets / sweep the heap even when all deadlines
  are far-future. An empty heap still sleeps until woken by schedule().
  - Added pending_task_count() and allocated_task_count() for observability.
  - Added unit tests for both the heap-retention and bucket-accumulation cases (each verified to fail with the corresponding fix disabled).

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/bthread/timer_thread.cpp
+++ b/src/bthread/timer_thread.cpp
@@ -34,6 +34,31 @@ namespace bthread {
 
 DEFINE_uint32(brpc_timer_num_buckets, 13, "brpc timer num buckets");
 
+// Tasks unscheduled after being pulled into the timer thread's min-heap are
+// only recycled when popped at their run_time, which can be far in the future
+// for large timeouts. To bound the memory they occupy (~ qps * timeout), the
+// timer thread periodically sweeps the heap and drops unscheduled tasks. The
+// sweep only kicks in once the heap grows beyond this size, so that small
+// heaps (where the retained memory is negligible) never pay the O(N) cost.
+DEFINE_uint32(brpc_timer_heap_sweep_min_size, 4096,
+              "The timer thread sweeps unscheduled tasks out of its internal "
+              "heap only when the heap has at least this many tasks");
+
+// The timer thread only consumes buckets and reclaims unscheduled tasks when
+// it wakes up, which normally happens at the nearest task's run_time. If every
+// pending task has a far-future run_time (e.g. minutes away), the thread would
+// sleep that whole time while newly scheduled-then-unscheduled tasks pile up
+// in the buckets, occupying pooled slots for the entire duration. Capping the
+// sleep makes the thread wake up periodically to drain the buckets and sweep
+// the heap, bounding that latency regardless of the timeout distribution.
+// 0 (the default) disables the cap: sleep until the nearest run_time, the
+// legacy behavior. Set it to a positive value to bound reclaim latency when
+// tasks may have far-future run_times.
+DEFINE_uint32(brpc_timer_max_wakeup_interval_ms, 0,
+              "The timer thread wakes up at least this often (in milliseconds) "
+              "to reclaim unscheduled tasks even when all pending tasks are far "
+              "in the future; 0 means no periodic wakeup");
+
 // Defined in task_control.cpp
 void run_worker_startfn();
 
@@ -132,6 +157,7 @@ TimerThread::TimerThread()
     , _buckets(NULL)
     , _nearest_run_time(std::numeric_limits<int64_t>::max())
     , _nsignals(0)
+    , _npending(0)
     , _thread(0) {
 }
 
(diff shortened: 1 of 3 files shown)


test/bthread_timer_thread_unittest.cpp
```diff
@@ -254,4 +256,113 @@ TEST(TimerThreadTest, schedule_and_unschedule_in_task) {
...
+static void noop_routine(void*) {}
+
+// Tasks that are unscheduled after being pulled into the timer thread's
+// internal heap must not stay there (occupying a pooled Task slot) until
+// their run_time. With large timeouts that would let memory grow ~ qps *
+// timeout. The timer thread should sweep them out and keep the heap bounded.
+TEST(TimerThreadTest, sweep_unscheduled_tasks_in_heap) {
+    // Lower the sweep threshold so the test doesn't need a huge heap.
+    GFLAGS_NAMESPACE::SetCommandLineOption(
+        "brpc_timer_heap_sweep_min_size", "512");
```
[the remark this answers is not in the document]
stdpain: UT update

src/bthread/timer_thread.cpp
```diff
@@ -408,6 +466,12 @@ void TimerThread::run() {
             BT_VLOG << "pull again, tasks=" << tasks.size();
             continue;
         }
```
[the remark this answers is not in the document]
stdpain: fixed

src/bthread/timer_thread.h
```diff
@@ -84,7 +84,22 @@ class TimerThread {
     // Get identifier of internal pthread.
     // Returns (pthread_t)0 if start() is not called yet.
     pthread_t thread_id() const { return _thread; }
-    
+
+    // Number of tasks currently buffered in the timer thread's internal
+    // min-heap (pulled out of buckets but not run yet). Exposed mainly for
+    // tests and observability. Note that tasks unscheduled after being pulled
+    // into the heap are counted until the timer thread sweeps them away.
+    int64_t pending_task_count() const {
```
wwbmmm: You don't need to add this public interface, brpc test code can assess private members (through the `-fno-access-control` compiler flag)

src/bthread/timer_thread.h
```diff
@@ -84,7 +84,22 @@ class TimerThread {
...
+        return _npending.load(butil::memory_order_relaxed);
+    }
+
+    // Total number of internal Task objects the process-wide pool has ever
+    // constructed (a high-water mark that never shrinks). Shared by all
+    // TimerThread instances. Exposed for tests/observability: if unscheduled
+    // tasks are reclaimed promptly, freed slots get reused and this stays
+    // bounded instead of growing with the number of scheduled tasks.
+    static size_t allocated_task_count();
```
wwbmmm: Same as above

stdpain: fixed

src/bthread/timer_thread.h
stdpain: fixed

wwbmmm:
LGTM

PR #3387, opened 2026-07-14
skip header fields containing CR/LF in http serialization

ubeddulla:

Issue Number:

Problem Summary:
`MakeRawHttpRequest` and `MakeRawHttpResponse` write every header field-name and field-value straight onto the wire (`os << it->first << ": " << it->second << CRLF`), and emit `Content-Type` from `content_type()` the same way. When one of those values carries a raw CR or LF (for example a value that an application forwards from an untrusted request), it closes the current header line and the following bytes are parsed as extra header fields or a body, i.e. HTTP request/response splitting. The inbound `http_parser` already refuses these control bytes in header fields, but the outbound serializers trusted them.

Changed:
1. Add `HeaderHasCRLF()` and, in both serializers, skip any header whose name or value contains CR or LF (a warning is logged).
2. Apply the same check to the `Content-Type` line in both functions.

Side effects:
- Performance effects: one `find_first_of` scan per emitted header; negligible.
- Breaking backward compatibility: none for well-formed messages. A header field carrying a raw CR/LF is already malformed under RFC 7230 and is now dropped instead of being split onto the wire.

### DIFF
--- a/src/brpc/details/http_message.cpp
+++ b/src/brpc/details/http_message.cpp
@@ -578,6 +578,16 @@ std::ostream& operator<<(std::ostream& os, const http_parser& parser) {
 
 #define BRPC_CRLF "\r\n"
 
+// A header field-name or field-value carrying a raw CR or LF lets whoever
+// controls it close the current line and inject extra header fields (or a
+// body) into the serialized message, i.e. HTTP request/response splitting.
+// The inbound parser already refuses these bytes; the outbound path drops
+// such fields so a value forwarded from an untrusted source can't smuggle
+// headers.
+static bool HeaderHasCRLF(const std::string& s) {
+    return s.find_first_of("\r\n") != std::string::npos;
+}
+
 // Request format
 // Request       = Request-Line              ; Section 5.1
 //                 *(( general-header        ; Section 4.5
@@ -644,12 +654,17 @@ void MakeRawHttpRequest(butil::IOBuf* request,
         }
         os << BRPC_CRLF;
     }
-    if (!h->content_type().empty()) {
+    if (!h->content_type().empty() && !HeaderHasCRLF(h->content_type())) {
         os << "Content-Type: " << h->content_type()
            << BRPC_CRLF;
     }
     for (HttpHeader::HeaderIterator it = h->HeaderBegin();
          it != h->HeaderEnd(); ++it) {
+        if (HeaderHasCRLF(it->first) || HeaderHasCRLF(it->second)) {
+            LOG(ERROR) << "Skip header `" << it->first
+                       << "' containing CR/LF to avoid injection";
+            continue;
+        }
         os << it->first << ": " << it->second << BRPC_CRLF;
     }
     if (h->GetHeader("Accept") == NULL) {
@@ -733,12 +748,18 @@ void MakeRawHttpResponse(butil::IOBuf* response,
             }
         }
     }
-    if (!is_invalid_content && !h->content_type().empty()) {
+    if (!is_invalid_content && !h->content_type().empty() &&
+        !HeaderHasCRLF(h->content_type())) {
         os << "Content-Type: " << h->content_type()
            << BRPC_CRLF;
     }
     for (HttpHeader::HeaderIterator it = h->HeaderBegin();
          it != h->HeaderEnd(); ++it) {
+        if (HeaderHasCRLF(it->first) || HeaderHasCRLF(it->second)) {
+            LOG(ERROR) << "Skip header `" << it->first
+                       << "' containing CR/LF to avoid injection";
+            continue;
+        }
         os << it->first << ": " << it->second << BRPC_CRLF;
     }
     os << BRPC_CRLF;  // CRLF before content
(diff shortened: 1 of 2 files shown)


src/brpc/details/http_message.cpp
```diff
@@ -644,12 +654,17 @@ void MakeRawHttpRequest(butil::IOBuf* request,
...
     }
-    if (!h->content_type().empty()) {
+    if (!h->content_type().empty() && !HeaderHasCRLF(h->content_type())) {
         os << "Content-Type: " << h->content_type()
            << BRPC_CRLF;
     }
     for (HttpHeader::HeaderIterator it = h->HeaderBegin();
          it != h->HeaderEnd(); ++it) {
+        if (HeaderHasCRLF(it->first) || HeaderHasCRLF(it->second)) {
+            LOG(ERROR) << "Skip header `" << it->first
```
wwbmmm: This should be warning, not error

src/brpc/details/http_message.cpp
```diff
@@ -578,6 +578,16 @@ std::ostream& operator<<(std::ostream& os, const http_parser& parser) {
...
 #define BRPC_CRLF "\r\n"
 
+// A header field-name or field-value carrying a raw CR or LF lets whoever
+// controls it close the current line and inject extra header fields (or a
+// body) into the serialized message, i.e. HTTP request/response splitting.
+// The inbound parser already refuses these bytes; the outbound path drops
+// such fields so a value forwarded from an untrusted source can't smuggle
+// headers.
+static bool HeaderHasCRLF(const std::string& s) {
+    return s.find_first_of("\r\n") != std::string::npos;
```
wwbmmm: This may affect performance, since MakeRawHttpRequest/Response are hot paths. You'd better add a flag to switch this on/off.

ubeddulla: Good point. I put it behind -http_check_outbound_header_crlf (default on). The scan is a single find_first_of over the header name/value, so it's cheap, but anyone who never forwards untrusted header values can turn it off.

src/brpc/details/http_message.cpp
ubeddulla: Done, switched both to LOG(WARNING). Also wrapped the logged name in butil::ToPrintable so a CR/LF in the name can't inject newlines into the log.

ubeddulla:
Pushed an update covering the review:

- Gated the CR/LF check behind `-http_check_outbound_header_crlf` (default on) so it can be turned off on hot paths that never forward untrusted header values.
- Changed the skip logs from ERROR to WARNING, and run the header name through `butil::ToPrintable` so a CR/LF in the name can't inject newlines into the log.
- Content-Type now logs the same warning when skipped instead of dropping silently, in both request and response paths.
- Added an explicit Content-Type CR/LF test for request and response, each with its own IOBuf.

wwbmmm:
LGTM

src/brpc/details/http_message.cpp
```diff
@@ -734,11 +761,22 @@ void MakeRawHttpResponse(butil::IOBuf* response,
...
+                         << "' containing CR/LF to avoid injection";
+        } else {
+            os << "Content-Type: " << h->content_type()
+               << BRPC_CRLF;
+        }
     }
     for (HttpHeader::HeaderIterator it = h->HeaderBegin();
          it != h->HeaderEnd(); ++it) {
+        if (HeaderHasCRLF(it->first) || HeaderHasCRLF(it->second)) {
+            LOG(WARNING) << "Skip header `" << butil::ToPrintable(it->first)
```
chenBright, line 777: `it->second` also need to be logged.

ubeddulla: Done, the value is logged too now (also through ToPrintable so a CR/LF in it stays on one line).

test/brpc_http_message_unittest.cpp
```diff
@@ -703,6 +703,31 @@ TEST(HttpMessageTest, serialize_http_response) {
...
+    ASSERT_EQ(std::string::npos, request_str.find("Injected: 1")) << request_str;
+
+    brpc::HttpHeader res_header;
+    res_header.SetHeader("X-Evil", "a\r\nInjected: 1");
+    butil::IOBuf response;
+    content.append("data");
+    MakeRawHttpResponse(&response, &res_header, &content);
+    std::string response_str = response.to_string();
+    ASSERT_EQ(std::string::npos, response_str.find("Injected: 1")) << response_str;
+}
```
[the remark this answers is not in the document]
ubeddulla: Cleaned this up as well - the header test now uses its own IOBuf for the request and for the response instead of reusing one across both calls.

ubeddulla:
Small follow-up: the header CR/LF test was still reusing one IOBuf for both MakeRawHttpRequest and MakeRawHttpResponse, so each call now gets its own. That's every review point covered as far as I can tell - happy to squash the three commits if you'd rather have this as one.

ubeddulla:
any update?

chenBright (approved):
LGTM

PR #3393, opened 2026-07-17
bthread: use BAIDU_GET_VOLATILE_THREAD_LOCAL to access tls_unique_use…

altman08:
…r_ptr in sched_to

tls_unique_user_ptr was read directly without going through the BAIDU_GET_VOLATILE_THREAD_LOCAL macro in TaskGroup::sched_to, while the paired write later in the same function correctly used BAIDU_SET_VOLATILE_THREAD_LOCAL. sched_to contains jump_stack, a suspend-point where GCC (aarch64) and Clang may incorrectly cache the address of a thread_local variable, which is exactly the hazard BAIDU_GET_VOLATILE_THREAD_LOCAL is designed to avoid. Fix the read to use the macro for consistency and correctness.

Issue Number: resolve 

Problem Summary:

Changed:

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/bthread/task_group.cpp
+++ b/src/bthread/task_group.cpp
@@ -817,7 +817,7 @@ void TaskGroup::sched_to(TaskGroup** pg, TaskMeta* next_meta) {
 #endif
     // Save errno so that errno is bthread-specific.
     int saved_errno = errno;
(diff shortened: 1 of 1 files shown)


yanglimingcn:
LGTM

PR #3394, opened 2026-07-17
bthread: convert ADD/SUB_TLS_PTHREAD_LOCK_COUNT macros to inline function

altman08:
…tions

Replace the function-like macros ADD_TLS_PTHREAD_LOCK_COUNT and SUB_TLS_PTHREAD_LOCK_COUNT with inline functions
AddTlsPthreadLockCount() and SubTlsPthreadLockCount() (with matching no-op definitions in the BRPC_DEBUG_BTHREAD_SCHE_SAFETY=0 branch). Macros lack type safety and scoping, so encapsulating the logic in regular functions is clearer and safer while keeping the same behavior. All call sites are updated accordingly.

Issue Number: resolve 

Problem Summary:

Changed:

Side effects:
- Performance effects:

---
### Check List:


yanglimingcn:
LGTM

PR #3395, opened 2026-07-19
read GOAWAY fields before Additional Debug Data in OnGoAway

ubeddulla:

Issue Number: N/A

Problem Summary:

1. RFC 7540 section 6.8 puts Last-Stream-ID and Error Code at the front of the GOAWAY payload with the Additional Debug Data after them, but `OnGoAway` skips `payload_size - 8` bytes first and then reads the trailing 8 bytes, so a peer that attaches debug data gets to choose the Last-Stream-ID we act on.
2. It also keeps the reserved high bit, so a `0xff` debug byte makes `last_stream_id` negative, and `_goaway_stream_id` then never passes the `>= 0` test in `TryToInsertStream`, leaving the client free to open new streams on a connection the peer has already abandoned.

Changed: moved the two reads ahead of the forward and masked the reserved bit, the way `ConsumeFrameHead` and `OnWindowUpdate` already treat theirs. The new `http2_goaway_with_debug_data` case sends a GOAWAY announcing stream 1 with 8 bytes of debug data and checks `_goaway_stream_id` ends up as 1 instead of -1.

Side effects:
- Performance effects: none, same number of bytes consumed from the iterator.

- Breaking backward compatibility: no. A GOAWAY carrying no debug data parses exactly as before.

### DIFF
--- a/src/brpc/policy/http2_rpc_protocol.cpp
+++ b/src/brpc/policy/http2_rpc_protocol.cpp
@@ -977,10 +977,12 @@ H2ParseResult H2Context::OnGoAway(
         LOG(ERROR) << "Invalid flags=" << h.flags;
         return MakeH2Error(H2_PROTOCOL_ERROR);
     }
+    // Last-Stream-ID and Error Code precede the Additional Debug Data.
+    // The reserved bit of Last-Stream-ID is ignored on receipt.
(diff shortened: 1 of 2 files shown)


wwbmmm:
LGTM

chenBright (approved):
LGTM

PR #3400, opened 2026-07-21
refactor(bvar): expand metric constructor macro

darion-yaphet:
## Summary

  - Remove the `COMMON_VARIABLE_CONSTRUCTOR` macro
  - Replace its uses in the Babylon counter variants of `Adder`, `Maxer`, `Miner`, and `IntRecorder` with explicit constructors
  - Preserve the existing constructor behavior for default construction, `expose(name)`, and `expose_as(prefix, name)`

  ## Motivation

  `COMMON_VARIABLE_CONSTRUCTOR` hid public constructors behind macro expansion, making the code harder to read, search, and navigate with regular C++ tooling.

  This change keeps the API behavior unchanged while making the constructors visible at their declaration sites.

  `using Base::Base` was intentionally avoided because it would expose additional internal base constructors in the Babylon counter path and could broaden the public
  API.

  ## Testing

  - `cmake --build build --target brpc-static -j6`
  - `git diff --check`
  - `rg COMMON_VARIABLE_CONSTRUCTOR . -g '!build/**'`

  ## Notes

  The `WITH_BABYLON_COUNTER` path was not compiled locally because the available Bazel setup is missing the repository-required Bazel `7.2.1` binary.

### DIFF
--- a/src/bvar/recorder.h
+++ b/src/bvar/recorder.h
@@ -302,7 +302,13 @@ public:
     typedef detail::MinusStat InvOp;
     typedef detail::ReducerSampler<IntRecorder, value_type, Op, InvOp> sampler_type;
 
-    COMMON_VARIABLE_CONSTRUCTOR(IntRecorder);
+    IntRecorder() = default;
+    IntRecorder(const butil::StringPiece& name) {
+        this->expose(name);
+    }
+    IntRecorder(const butil::StringPiece& prefix, const butil::StringPiece& name) {
+        this->expose_as(prefix, name);
+    }
 
     DISALLOW_COPY_AND_MOVE(IntRecorder);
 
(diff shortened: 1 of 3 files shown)


chenBright (commented):
I think this is to simplify the code, which is fine, like the DISALLOW_COPY_AND_ASSIGN macro.

wwbmmm:
> I think this is to simplify the code, which is fine, like the DISALLOW_COPY_AND_ASSIGN macro.

Agree

darion-yaphet:
> > I think this is to simplify the code, which is fine, like the DISALLOW_COPY_AND_ASSIGN macro.
> 
> Agree

I agree the macro reduces a few lines, but it hides public constructors, which are part of the API surface and important for readability/tooling. Since these constructors are simple and type-specific, spelling them out makes the public interface clearer.

That said, because variable.h is installed as a public header, removing the macro may break downstream users. I can keep a deprecated compatibility definition for one release while removing internal usages, then remove it in a later breaking-change cycle.

PR #3403, opened 2026-07-22
refactor(bvar): replace NULL with nullptr

darion-yaphet:

  Issue Number: resolve N/A

  Problem Summary:

  `src/bvar` still used the legacy `NULL` macro for pointer null values. Since bRPC already builds with at least C++11, using `nullptr` is clearer and type-safe,  especially in overloaded calls and pointer comparisons.

  Changed:

  - Replaced `NULL` with `nullptr` across `src/bvar` headers and implementation files.
  - Updated related comments that referred to returning `NULL`.
  - Kept the cleanup scoped to bvar sources only.

  Side effects:

  - Performance effects: None expected. This is a source-level modernization with no intended runtime behavior change.

  - Breaking backward compatibility: No. `nullptr` is available under the project’s C++11 baseline.

  ---
  ### Check List:
    - Verified with `cmake --build build --target brpc-static -j6`.
    - Verified with `git diff --check`.
    - Verified no `NULL` tokens remain in `src/bvar` via `rg '\bNULL\b' src/bvar -g '*.{h,cpp}'`.
    - Not applicable; this PR does not add a feature or change behavior.

### DIFF
--- a/src/bvar/collector.cpp
+++ b/src/bvar/collector.cpp
@@ -43,10 +43,10 @@ BAIDU_CASSERT(!(COLLECTOR_SAMPLING_BASE & (COLLECTOR_SAMPLING_BASE - 1)),
 // Combine two circular linked list into one.
 struct CombineCollected {
     void operator()(Collected* & s1, Collected* s2) const {
-        if (s2 == NULL) {
+        if (s2 == nullptr) {
             return;
         }
-        if (s1 == NULL) {
+        if (s1 == nullptr) {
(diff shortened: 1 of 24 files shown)


chenBright:
Nice job!

I previously submitted a PR (#2786) to refactor all NULL values ​​in the project into nullptr. However, the change was too significant, so it wasn't merged.

I think it could be split into multiple PRs for the refactoring.

chenBright (approved):
LGTM

PR #3406, opened 2026-07-24
Fix race when falling back from RDMA to TCP

### LINKED ISSUE #3405: [RDMA] Data race and partial state publication during TCP fallback
## Description

On current `master` (`e0444287`), the RDMA handshake state is accessed
concurrently but is stored in a non-atomic field:

```cpp
// rdma_endpoint.h
State _state;
```

`RdmaEndpoint::_state` is read or written by multiple execution contexts,
including:

- the RDMA handshake bthread;
- the TCP event handling path in `OnNewDataFromTcp()`;
- the RDMA completion path in `HandleCompletion()`.

This constitutes a C++ data race.

There is also a dual-state publication issue when RDMA is unavailable.
`RdmaConnect::StartConnect()` currently publishes the two related states in
this order:

```cpp
rdma_transport->_rdma_ep->_state = RdmaEndpoint::FALLBACK_TCP;
rdma_transport->_rdma_state = RdmaTransport::RDMA_OFF;
```

The endpoint state is used by `OnNewDataFromTcp()` to decide whether normal TCP
message processing can begin:

```cpp
if (ep->_state == FALLBACK_TCP) {
    InputMessenger::OnNewMessages(m);
    return;
}
```

A concurrent TCP event can observe `FALLBACK_TCP` before `_rdma_state` has been
updated to `RDMA_OFF`. This exposes a partially published fallback state to the
TCP processing path.

## Race sequence

```mermaid
sequenceDiagram
    participant C as Connection/StartConnect bthread
    participant E as TCP event bthread
    participant M as InputMessenger

    Note over C: RDMA is unavailable

    C->>C: ep->_state = FALLBACK_TCP
    Note over C,E: C may be descheduled here

    E->>E: OnNewDataFromTcp(socket)
    E->>E: observe ep->_state == FALLBACK_TCP
    E->>M: InputMessenger::OnNewMessages(socket)
    Note over E,M: rdma_transport->_rdma_state<br/>may still be RDMA_UNKNOWN or RDMA_ON

    C->>C: rdma_transport->_rdma_state = RDMA_OFF
    Note over C: The socket-level state is published too late
```

Equivalent thread interleaving:

```text
StartConnect bthread                    TCP event bthread
--------------------                    -----------------
ep->_state = FALLBACK_TCP;

             < context switch >

                                        OnNewDataFromTcp(socket);

                                        if (ep->_state == FALLBACK_TCP) {
                                            InputMessenger::OnNewMessages(
                                                socket);
                                            // _rdma_state may not be RDMA_OFF
                                        }

rdma_transport->_rdma_state = RDMA_OFF;
```

The invalid intermediate state is:

```text
RdmaEndpoint::_state == FALLBACK_TCP
RdmaTransport::_rdma_state != RDMA_OFF
```

## Expected behavior

Once `OnNewDataFromTcp()` observes:

```cpp
ep->_state == FALLBACK_TCP
```

the corresponding transport state must already be visible as:

```cpp
rdma_transport->_rdma_state == RdmaTransport::RDMA_OFF
```

The fallback transition must not expose a partially published state.

## Impact

The exact effect depends on timing and the code consuming `_rdma_state`.
Possible consequences include:

- entering normal TCP message processing while the transport still appears to
  be in an RDMA state;
- inconsistent TCP/RDMA path selection;
- undefined behavior caused by concurrent non-atomic accesses to `_state`.

An older downstream version detected this inconsistency with an invariant
CHECK in its TCP read path. Current upstream `master` no longer contains that
CHECK after the transport refactoring, but the underlying data race and
partial publication window remain.

legionxiong:

Issue Number: #3405

Problem Summary:

`RdmaEndpoint::_state` is accessed concurrently by the RDMA handshake,
TCP event handling, and completion handling paths, but was not atomic.
This causes a C++ data race.

During TCP fallback, `RdmaTransport::_rdma_state` and
`RdmaEndpoint::_state` also form a logical state transition. A TCP event
must not observe `FALLBACK_TCP` before the preceding `RDMA_OFF` update is
visible, otherwise it may start normal TCP message processing with a
partially published transport state.

Changed:

- Change `RdmaEndpoint::_state` to `butil::atomic<State>`.
- Set `RdmaTransport::_rdma_state` to `RDMA_OFF` before publishing
  `RdmaEndpoint::_state` as `FALLBACK_TCP`.
- Publish `FALLBACK_TCP` with `memory_order_release`.
- Load the handshake state with `memory_order_acquire` in
  `OnNewDataFromTcp()`. When it observes `FALLBACK_TCP`, this acquire load
  pairs with the release store and makes the preceding `RDMA_OFF` update
  visible before normal TCP message processing starts.
- Use `memory_order_relaxed` for all other handshake-state transitions and
  observations because they do not publish or consume related data.
- Add comments documenting the acquire/release synchronization and why
  relaxed ordering is sufficient for the other accesses.

Side effects:

- Performance effects:

  The acquire load is limited to the TCP event path, while release stores
  are only used when falling back to TCP. Other handshake-state accesses,
  including the completion hot path, use relaxed ordering to avoid
  unnecessary synchronization overhead.

  None. This change does not modify any public API, wire protocol, or
  externally visible configuration.

### DIFF
--- a/src/brpc/rdma/rdma_endpoint.cpp
+++ b/src/brpc/rdma/rdma_endpoint.cpp
@@ -153,7 +153,7 @@ RdmaEndpoint::~RdmaEndpoint() {
 void RdmaEndpoint::Reset() {
     DeallocateResources();
 
-    _state = UNINIT;
+    _state.store(UNINIT, butil::memory_order_relaxed);
     _handshake_version = 0;
     _outgoing_ece.reset();
     _resource = NULL;
@@ -189,8 +189,9 @@ void RdmaConnect::StartConnect(const Socket* socket,
         return;
     }
     if (!IsRdmaAvailable()) {
-        rdma_transport->_rdma_ep->_state = RdmaEndpoint::FALLBACK_TCP;
         rdma_transport->_rdma_state = RdmaTransport::RDMA_OFF;
+        rdma_transport->_rdma_ep->_state.store(
+            RdmaEndpoint::FALLBACK_TCP, butil::memory_order_release);
         done(0, data);
         return;
     }
@@ -222,16 +223,19 @@ void RdmaEndpoint::OnNewDataFromTcp(Socket* m) {
 
     int progress = Socket::PROGRESS_INIT;
     while (true) {
-        if (ep->_state == UNINIT) {
+        // Pair with release stores of FALLBACK_TCP so RDMA_OFF is visible
+        // before normal TCP message processing starts.
+        const State state = ep->_state.load(butil::memory_order_acquire);
+        if (state == UNINIT) {
             // The connection may be closed or reset before the client starts
             // handshake. This will be handled by client handshake. Ignore here.
-        } else if (ep->_state < ESTABLISHED) {  // during handshake
+        } else if (state < ESTABLISHED) {  // during handshake
             ep->_read_butex->fetch_add(1, butil::memory_order_release);
             bthread::butex_wake(ep->_read_butex);
-        } else if (ep->_state == FALLBACK_TCP){  // handshake finishes
+        } else if (state == FALLBACK_TCP){  // handshake finishes
             InputMessenger::OnNewMessages(m);
             return;
-        } else if (ep->_state == ESTABLISHED) {
+        } else if (state == ESTABLISHED) {
             uint8_t tmp;
             ssize_t nr = read(ep->_socket->fd(), &tmp, 1);
             if (nr == 0) {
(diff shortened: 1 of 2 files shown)


chenBright (changes requested):
When falling back to TCP, the require/release memory order is sufficient, and comments should be added. otherwise, use relaxed ordering.

legionxiong:
> When falling back to TCP, the require/release memory order is sufficient, and comments should be added. otherwise, use relaxed ordering.

Done.

chenBright (approved):
LGTM

zyearn:
LGTM. Merging

PR #3407, opened 2026-07-24
Support checksumming the attachment together with the body

### LINKED ISSUE #3046: Is CRC validation for attachments under consideration?
Hi, I noticed that while brpc supports CRC checks, it currently doesn't validate the attachment field.
In some cases, users put important payloads in attachments, so adding CRC for them could improve data integrity.
I looked into the implementation and it seems not difficult to support.
If this feature is considered useful, I’d be happy to work on a patch.

Thanks!

I suggest extending the existing CRC mechanism with two FLAGS:

1. brpc_checksum: enables CRC validation at the brpc layer. When enabled, only the message will be validated.

2. brpc_checksum_attachment: when enabled alongside brpc_checksum, CRC validation will also include the attachment field.

The checksum for the attachment can be computed by reusing the intermediate CRC state from the message part, so that the existing checksum_value field in the RPC meta can still be used without protocol changes.

yanglimingcn:
Controller::set_request/response_checksum_type() previously only covered the serialized protobuf body; the attachment (if any) was never protected. Add set_request/response_checksum_attachment(bool) so callers can opt the attachment into the same checksum.

- baidu_rpc_meta.proto: add RpcMeta.checksum_with_attachment so the receiver knows whether to fold the attachment into verification. Defaults to false, so old peers that don't understand the field keep verifying against the body only (backward compatible).
- Controller: add the two setters/getters, thread the flag through ClientSettings (Save/ApplyClientSettings) so ParallelChannel/ SelectiveChannel sub-controllers inherit it correctly, and reset it in ResetPods().
- ChecksumIn: add an optional `attachment' field consumed by checksum handlers.
- crc32c_checksum.cpp: extend the crc32c over body then attachment (in that fixed order) when requested.
- baidu_rpc_protocol.cpp: wire checksum_attachment through SerializeRpcMessage/DeserializeRpcMessage and every client/server send/receive path; skip it when progressive attachment reading is enabled since there's no single complete IOBuf to checksum in that case.

Add brpc_checksum_unittest.cpp covering Crc32cCompute/Crc32cVerify directly (including corruption/omission/order sensitivity) and an end-to-end Server/Channel test for both request- and response-side attachment checksums.

Issue Number: resolve #3046

Problem Summary:

Changed:

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/brpc/controller.h
+++ b/src/brpc/controller.h
@@ -145,6 +145,9 @@ friend void policy::ProcessThriftRequest(InputMessageBase*);
     static const uint32_t FLAGS_READ_PROGRESSIVELY = (1 << 3);
     static const uint32_t FLAGS_PROGRESSIVE_READER = (1 << 4);
     static const uint32_t FLAGS_BACKUP_REQUEST = (1 << 5);
+    // Whether set_request_checksum_type()'s checksum also covers the
+    // attachment. See set_request_checksum_attachment().
+    static const uint32_t FLAGS_REQUEST_CHECKSUM_WITH_ATTACHMENT = (1 << 6);
     // Let _done delete the correlation_id, used by combo channels to
     // make lifetime of the correlation_id more flexible.
     static const uint32_t FLAGS_DESTROY_CID_IN_DONE = (1 << 7);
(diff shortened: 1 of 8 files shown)


src/brpc/controller.h
```diff
@@ -897,6 +924,8 @@ friend void policy::ProcessThriftRequest(InputMessageBase*);
     CompressType _response_compress_type;
     ChecksumType _request_checksum_type;
     ChecksumType _response_checksum_type;
+    bool _request_checksum_with_attachment;
+    bool _response_checksum_with_attachment;
```
wwbmmm: Can you store these two flags in controller flags?

chenBright (approved):
LGTM

PR #3408, opened 2026-07-26
Add TLS peer name verification for clients

wwbmmm:

Issue Number: null

Problem Summary:

Client TLS verification validates the certificate chain, but callers cannot
configure the DNS name or IP address expected in the peer certificate. This
makes it possible for a trusted certificate issued for another endpoint to be
accepted when only chain verification is configured.

Changed:

- Add `VerifyOptions::expected_peer_name` for DNS name and IP SAN validation.
- Derive the expected peer name from the host in an HTTPS Channel URL when
  certificate verification is enabled and no explicit peer name is configured.
- Configure OpenSSL peer name verification as part of certificate verification.
- Reject peer name verification when certificate verification is disabled or
  the TLS backend cannot support it.
- Include the verification mode and expected peer name in the channel signature
  to prevent connections with different verification settings from being
  shared.
- Add tests for matching DNS names, mismatched DNS names, IP SAN mismatches, and
  invalid verification configurations.

Side effects:

- Performance effects: one DNS name or IP SAN comparison is added during a TLS
  handshake when `expected_peer_name` is configured.
- Breaking backward compatibility: HTTPS Channels with certificate verification
  enabled now reject certificates that do not match the URL host. Callers that
  intentionally connect through an alias can set `expected_peer_name` explicitly.
  OpenSSL versions older than 1.0.2 and MesaLink reject configurations that
  require peer name verification because they do not provide the required API.

---
### Check List:

- [x] The project builds with `make -j6` in `build`.
- [x] `build/test/brpc_ssl_unittest` passes.
- [x] Related tests cover matching and mismatched peer identities.

### DIFF
--- a/src/brpc/channel.cpp
+++ b/src/brpc/channel.cpp
@@ -128,7 +128,11 @@ static ChannelSignature ComputeChannelSignature(const ChannelOptions& opt) {
             buf.push_back('|');
             buf.append((char*)&verify.verify_depth, sizeof(verify.verify_depth));
             buf.push_back('|');
+            buf.append((char*)&verify.verify_mode, sizeof(verify.verify_mode));
+            buf.push_back('|');
             buf.append(verify.ca_file_path);
+            buf.push_back('|');
+            buf.append(verify.expected_peer_name);
         } else {
             // All disabled ChannelSSLOptions are the same
         }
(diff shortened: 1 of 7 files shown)


zyearn:
My main concern is the older MesaLink/old-OpenSSL backward compatibility break. Otherwise LGTM

Is there anything we can do to avoid it? Maybe a flag so existing users won't be silently broken?

wwbmmm:
> My main concern is the older MesaLink/old-OpenSSL backward compatibility break. Otherwise LGTM
> 
> Is there anything we can do to avoid it? Maybe a flag so existing users won't be silently broken?

You are right. I’ve updated the PR to preserve backward compatibility:

* OpenSSL 1.0.2+ and BoringSSL automatically verify the HTTPS URL host.
* MesaLink and older OpenSSL retain chain-only verification and emit a one-time warning.
* Explicitly setting expected_peer_name still fails on unsupported backends instead of being silently ignored.

PR #3412, opened 2026-07-27
Reject stale CQ callbacks after socket revival

### LINKED ISSUE #3410: [RDMA] SIGSEGV caused by stale PollCq callback
**Describe the bug**

An old `RdmaEndpoint::PollCq` callback can survive an RDMA
`Reset()`/health-check `Revive()` cycle and then access the resources of
the new RDMA generation.

PR #3261 / commit 58fec6fe documented that a queued CQ callback may run
after the main Socket has been reset and revived. The merged fix drains
the old CQ after moving the QP to RESET, but it does not invalidate or
synchronize with a callback that has already copied the non-owning
`RdmaEndpoint*` from the old CQ Socket.

The following interleaving is still possible:

```text
Old CQ Socket callback                 Main Socket / health checker
----------------------                 ----------------------------

PollCq(old_cq_socket)
  ep = old_cq_socket->user()
  [preempted before Address()]         QP failure / SetFailed()
                                       WaitAndReset()
                                       RdmaEndpoint::Reset()
                                         DeallocateResources()
                                           old_cq_socket->_user = NULL
                                           drain old CQ
                                       CheckHealth() / Revive()
                                       AllocateResources()
                                         create new _resource
                                         create new _cq_sid

  Socket::Address(ep->_socket->id())
    succeeds after revival

  cq = ep->_resource->...              New CQ callback may run
    // This is the new CQ
  ibv_poll_cq(cq)
  HandleCompletion(wc)
```

Setting the old CQ Socket's `_user` to `NULL` does not invalidate the
`ep` pointer already copied by a running callback.

`PollCq` verifies that the main Socket is addressable, but does not
verify that:

- `m->id()` is still equal to `ep->_cq_sid`;
- the callback belongs to the current RDMA generation;
- `ep->_resource` is the resource for which the callback was created.

The old and new CQ Socket callbacks can therefore both poll the new CQ
and concurrently call `HandleCompletion()` on the same `RdmaEndpoint`.

Because they belong to different CQ Sockets, Socket-level event
serialization does not serialize their access to the endpoint.
`HandleCompletion()` modifies `_sq_sent`, `_rq_received`, `_sbuf`,
`_rbuf`, and the main Socket's `_read_buf` without endpoint-level
synchronization.

For example, concurrent SEND completions can race on:

```cpp
_sbuf[_sq_sent++].clear();
```

This may result in an out-of-bounds `_sbuf` access or concurrent
`IOBuf::clear()` operations, matching the crash stack reported in
#3252:

```text
butil::IOBuf::clear()
brpc::rdma::RdmaEndpoint::HandleCompletion()
brpc::rdma::RdmaEndpoint::PollCq()
brpc::Socket::ProcessEvent()
```

**To Reproduce**

There is no minimal standalone reproducer yet.

The race can be reproduced deterministically with a test hook:

1. Enable RDMA and Socket health checking.
2. Pause an old CQ callback immediately after reading `m->user()`, but
   before calling `Socket::Address(ep->_socket->id(), ...)`.
3. Force the QP into an error state.
4. Allow `WaitAndReset()`, `RdmaEndpoint::Reset()`, health checking, and
   `Revive()` to complete.
5. Complete a new RDMA handshake so that a new `_resource` and
   `_cq_sid` are installed.
6. Resume the old callback while the new CQ callback is active.
7. Generate SEND/RECV completions or run the test under TSAN.

**Expected behavior**

A CQ callback must never access an `RdmaEndpoint` or `RdmaResource`
belonging to a different Reset/Revive generation.

Reset should either wait for all callbacks of the old generation to
finish, or callbacks should carry lifetime-protected context and reject
themselves when their CQ Socket, resource, or generation is no longer
current.

A complete fix could bind each callback to immutable context such as:

```text
{ main_socket_id, cq_socket_id, resource, generation }
```

After acquiring the main Socket, `PollCq` should validate the CQ Socket,
resource, and generation before accessing `ep->_resource`.
Polling-mode REMOVE operations require equivalent synchronization.

**Versions**
OS: Debian 12
Compiler: g++ 14.2.0
brpc: 1.14; the callback lifecycle described above is also present in master
protobuf: not relevant

Related issue: https://github.com/apache/brpc/issues/3252

Related PR: https://github.com/apache/brpc/pull/3261

CQ-draining fix: https://github.com/apache/brpc/commit/58fec6fe

PR #3261 explicitly considered incrementing a generation in `Reset()`
and validating it in `PollCq`, but the merged implementation chose CQ
draining only.

Draining removes stale CQEs. It does not stop an old callback from
resuming after revival and selecting the newly installed
`ep->_resource`.

legionxiong:

Issue Number: resolves #3410 

Problem Summary:

An RDMA completion-queue callback may remain queued after its endpoint is
reset. If the main socket is later revived with a new CQ, the stale callback
can successfully acquire the revived socket and continue operating on the
endpoint's new generation.

This may cause the stale callback to access invalid or mismatched RDMA
resources in `RdmaEndpoint::PollCq`, potentially resulting in a crash.

Changed:

- Verify that the CQ socket passed to `RdmaEndpoint::PollCq` still matches the
  endpoint's current `_cq_sid`.
- Return immediately when the callback belongs to an older CQ generation.
- Add a regression test covering a stale callback running after the endpoint
  has switched to a new CQ.

Side effects:

- Performance effects: One `SocketId` comparison is added to each `PollCq`
  invocation. The overhead is negligible.

### DIFF
--- a/src/brpc/rdma/rdma_endpoint.cpp
+++ b/src/brpc/rdma/rdma_endpoint.cpp
@@ -1434,6 +1434,11 @@ void RdmaEndpoint::PollCq(Socket* m) {
     if (Socket::Address(ep->_socket->id(), &s) < 0) {
         return;
     }
+    // A queued callback may outlive Reset() and see the main Socket after
+    // it has been revived with another CQ.
+    if (m->id() != ep->_cq_sid) {
+        return;
+    }
     auto* rdma_transport = static_cast<RdmaTransport*>(s->_transport.get());
     CHECK(ep == rdma_transport->_rdma_ep);
 
--- a/test/brpc_rdma_unittest.cpp
+++ b/test/brpc_rdma_unittest.cpp
@@ -186,6 +186,40 @@ private:
     int _saved_handshake_version = 2;
 };
 
+TEST_F(RdmaTest, stale_cq_callback_does_not_poll_new_generation) {
+    SocketOptions main_options;
+    main_options.socket_mode = SOCKET_MODE_RDMA;
+    SocketId main_sid;
+    ASSERT_EQ(0, Socket::Create(main_options, &main_sid));
+
+    SocketUniquePtr main_socket;
+    ASSERT_EQ(0, Socket::Address(main_sid, &main_socket));
+    RdmaTransport* transport =
+        static_cast<RdmaTransport*>(main_socket->_transport.get());
+    rdma::RdmaEndpoint* ep = transport->_rdma_ep;
+
+    SocketOptions cq_options;
+    cq_options.user = ep;
+    SocketId stale_cq_sid;
+    SocketId current_cq_sid;
+    ASSERT_EQ(0, Socket::Create(cq_options, &stale_cq_sid));
+    ASSERT_EQ(0, Socket::Create(cq_options, &current_cq_sid));
+
+    SocketUniquePtr stale_cq_socket;
+    SocketUniquePtr current_cq_socket;
+    ASSERT_EQ(0, Socket::Address(stale_cq_sid, &stale_cq_socket));
+    ASSERT_EQ(0, Socket::Address(current_cq_sid, &current_cq_socket));
+    ep->_cq_sid = current_cq_sid;
+
+    rdma::RdmaEndpoint::PollCq(stale_cq_socket.get());
+
+    stale_cq_socket->_user = NULL;
+    current_cq_socket->_user = NULL;
+    stale_cq_socket->SetFailed();
+    current_cq_socket->SetFailed();
+    main_socket->SetFailed();
+}
+
 TEST_F(RdmaTest, client_close_before_hello_send) {
     StartServer();
 


Huixxi:
This fix is on the right track — it addresses the core race condition described in issue #3410, but it's a targeted patch rather than the more complete solution the issue actually suggests. Here's the breakdown:

**Is the fix correct?**

- The core problem in the issue: a `PollCq` callback belonging to an "old-generation" CQ Socket may resume execution *after* `RdmaEndpoint::Reset()` → health check → `Revive()` → a fresh handshake (which allocates a new `_resource` and a new `_cq_sid`). At that point, the `ep` (`RdmaEndpoint*`) it holds is already "new generation," while `m` (the CQ Socket that triggered the callback) is still "old generation."
- The new check added:

```cpp
if (m->id() != ep->_cq_sid) {
    return;
}
```

  This is exactly one of the key validations the issue suggested — "verify that `m->id()` still equals `ep->_cq_sid`." Since `_cq_sid` is regenerated via `Socket::Create` every time `AllocateResources()` runs (brpc's `SocketId` carries a version number, so even if the fd slot is reused, the id differs), the old-generation CQ Socket's id will necessarily not match the new-generation `_cq_sid`. This correctly rejects the vast majority of stale callbacks and is a reasonable, low-cost fix (one integer comparison).

- The added test `stale_cq_callback_does_not_poll_new_generation` directly constructs the scenario "old CQ socket + mismatched new `_cq_sid`" and verifies `PollCq` returns early, covering the main scenario described in the issue.

**What isn't fully addressed (the issue itself hints at a more complete solution):**

1. **A TOCTOU window still exists**: the check happens once at the start of `PollCq`, but the function later continues to access `ep->_resource` (e.g., `ep->_resource->recv_cq`). If `Reset()` happens right *after* this check passes but *before* `ep->_resource` is dereferenced (setting `_resource` to `NULL`), a crash could still occur. Ideally, per the issue's suggestion, the callback should be bound to immutable context such as `{main_socket_id, cq_socket_id, resource, generation}`, and re-validated right before actually touching `_resource` — not just once at function entry.
2. **No explicit "generation" field was introduced**: the issue text explicitly notes that PR #3261 considered adding a generation counter in `Reset()` and validating it in `PollCq`, but the merged implementation chose CQ draining only. This fix similarly just adds a `_cq_sid` comparison rather than an explicit generation. In practice, `_cq_sid`'s regeneration mechanism likely serves as an effective equivalent to a generation counter (since `SocketId` carries a version), but semantically it's less explicit than a dedicated generation field, and it wouldn't cover edge cases like "the same CQ Socket being reused by later logic" (though the current code doesn't exhibit that pattern).
3. **Only the shared entry point in `PollCq`** (used by both polling and non-polling modes) gets this check; `HandleCompletion` itself still has no secondary validation. If any future call path were to invoke `HandleCompletion` directly, bypassing `PollCq`, the same issue could resurface (currently no such path exists, so this is a latent maintenance risk rather than an active bug).

**Conclusion:**
This PR's change is correct in direction and does fix the specific crash scenario described in issue #3410 (a stale callback incorrectly accessing new-generation resources after socket revival). The change is small, the performance overhead is negligible, and the test specifically validates the fix point. However, it is a "minimally invasive" fix rather than the "complete solution" envisioned in the issue (e.g., finer-grained generation validation, or re-checking right before dereferencing `ep->_resource`). If similar TOCTOU scenarios surface later, further hardening may be needed (e.g., re-checking `ep->_cq_sid` immediately before actually dereferencing `ep->_resource`, or introducing an atomic generation field). I'd suggest flagging this in the review comments, but the fix can reasonably be merged as an incremental improvement in the meantime.

PR #3413, opened 2026-07-27
fix(vars): render /vars as an HTML name-value table

darion-yaphet:
  Issue Number: resolve
  Problem Summary:
  The `/vars` HTML page rendered each bvar as a plain `<p>` line (`name : value`), so names and values were hard to scan when lengths varied. Expandable flot chart
  blocks were also mixed into the same flow, which made the layout less structured.
  Changed:
  - Render `/vars` HTML output as a `gridtable` with `Name` / `Value` columns.
  - Put each variable in a `<tr class="variable|nonplot-variable">` row.
  - Move expandable flot charts into a dedicated `<tr class="detail-row">` under the corresponding variable.
  - Keep non-HTML (`curl` / plain text) `/vars` output unchanged.
  Side effects:
  - Performance effects: None expected. Only the HTML presentation path of `/vars` is changed.
  - Breaking backward compatibility: No. Plain-text `/vars` responses are unchanged; only browser HTML layout is updated.
  ---
  ### Check List:

image

### DIFF
--- a/src/brpc/builtin/vars_service.cpp
+++ b/src/brpc/builtin/vars_service.cpp
@@ -45,6 +45,7 @@ namespace brpc {
 void PutVarsHeading(std::ostream& os, bool expand_all) {
     os << "<script language=\"javascript\" type=\"text/javascript\" src=\"/js/jquery_min\"></script>\n"
         "<script language=\"javascript\" type=\"text/javascript\" src=\"/js/flot_min\"></script>\n"
+       << gridtable_style()
        << TabsHead()
        << "<style type=\"text/css\">\n"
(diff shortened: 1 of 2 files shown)


wwbmmm:
What does it look like when a bvar row expand? Can you give a screenshot?

darion-yaphet:
> What does it look like when a bvar row expand? Can you give a screenshot?

image

darion-yaphet:
Is anyone still following this PR?  @wwbmmm

wwbmmm:
LGTM

chenBright (approved):
LGTM

PR #3418, opened 2026-07-29
bthread: skip signal_task when no worker is waiting

Lin807188583:
## Problem
ready_to_run* unconditionally calls signal_task() on every task-ready event, even when all workers are busy and nobody is parked in wait(). This results in unnecessary futex_wake syscalls that wake no thread, wasting CPU under load.

## Solution
Add per-tag waiter counting via _tagged_waiter_num: ParkingLot::wait() tracks waiters entering/leaving (via parking_lot_waiter_add/sub out-of-line forwarders) BEFORE the state check. ready_to_run* checks has_waiting_workers(tag) before calling signal_task — if no worker of the given tag is parked, the wakeup is skipped (BAIDU_UNLIKELY marks the true branch as cold since under load all workers are busy).

## Race-free design
The waiter_add() is called BEFORE the get_state() check in wait(). This ensures that any concurrent ready_to_run() that calls has_waiting_workers() will see the waiter even if it races between the state check and futex_wait(). If the state already changed (signal was sent), wait() undoes the tracking and returns immediately without sleeping. This eliminates the lost-wakeup window entirely.

## Why it's safe
- Tasks are already enqueued (push_rq) before the signal check, so they will be picked up by the next steal regardless of signaling.
- _tagged_waiter_num uses relaxed atomics (it's a hint, not a synchronization primitive); the actual wake ordering is handled by parking_lot's existing futex mechanism.

## Multi-tag benefit
Per-tag accounting avoids tag-A's ready_to_run() issuing a signal_task() just because tag-B has parked workers.

Tested: bthread_unittest 20/20 on Linux x86 (Ubuntu 24.04/GCC 13) and macOS arm64 (M4).

Issue Number: resolve 

Problem Summary:

Changed:

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/bthread/parking_lot.h
+++ b/src/bthread/parking_lot.h
@@ -24,6 +24,7 @@
 
 #include <gflags/gflags.h>
 #include "butil/atomicops.h"
+#include "bthread/types.h"    // bthread_tag_t
 #include "bthread/sys_futex.h"
 
 namespace bthread {
--- a/src/bthread/parking_lot.h
+++ b/src/bthread/parking_lot.h
@@ -24,22 +24,11 @@
 
 #include <gflags/gflags.h>
 #include "butil/atomicops.h"
-#include "bthread/types.h"    // bthread_tag_t
 #include "bthread/sys_futex.h"
 
 namespace bthread {
 
 DECLARE_bool(parking_lot_no_signal_when_no_waiter);
-
-// Forward declaration
-class TaskControl;
-
-// Out-of-line helpers that forward to TaskControl's per-tag waiter accounting.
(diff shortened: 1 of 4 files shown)


Lin807188583:
确认存在丢唤醒竞态。worker 的"拍快照"和"登记 waiter"不是原子的，producer 卡在这个窗口判断 has_waiting_workers() 为 0 就跳过了 signal_task()，但 _pending_signal 没更新。应该始终更新 _pending_signal，只跳过 futex_wake。现有 --parking_lot_no_signal_when_no_waiter 已实现这个。先关闭

PR #3419, opened 2026-07-30
build(cxx): align all builds on a C++14 baseline

darion-yaphet:

  Issue Number: N/A

  Problem Summary:
  bRPC’s CMake and Make-based build paths used inconsistent C++11/C++14 settings. This establishes C++14 as the common minimum language level.

  Changed:

  - Updated CMake, Make, test, tool, and example build settings to require C++14.
  - Raised compiler checks to GCC 5.0+ and Clang 3.5+.
  - Updated configuration and user-facing documentation.
  - Preserved C++17 for Protobuf 22+ and Bazel.

  Side effects:

  - Performance effects: None expected.
  - Breaking backward compatibility: C++11-only toolchains and build invocations are no longer supported.

  ———

  ### Check List:

  - [x] Changes compile: clean CMake brpc-static build with AppleClang 16 and -std=gnu++14.
  - [x] Verified config_brpc.sh generation, shell syntax, and diff formatting.
  - [ ] Unit test suite not run.
  - [x] No new feature behavior requiring tests.
  - [x] Follows the Contributor Covenant.

### DIFF
--- a/src/butil/type_traits.h
+++ b/src/butil/type_traits.h
@@ -373,11 +373,11 @@ template <typename>
 struct result_of;
 template <typename F, typename... Args>
 struct result_of<F(Args...)> : std::invoke_result<F, Args...> {};
-#elif __cplusplus >= 201103L
+#elif __cplusplus >= 201402L
 template <typename F>
 using result_of = std::result_of<F>;
 #else
-#error Only C++11 or later is supported.
+#error Only C++14 or later is supported.
 #endif // __cplusplus
 
 template <typename F>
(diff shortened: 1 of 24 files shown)


wwbmmm:
LGTM

src/butil/type_traits.h
```diff
@@ -373,11 +373,11 @@ template <typename>
 struct result_of;
 template <typename F, typename... Args>
 struct result_of<F(Args...)> : std::invoke_result<F, Args...> {};
-#elif __cplusplus >= 201103L
+#elif __cplusplus >= 201402L
 template <typename F>
 using result_of = std::result_of<F>;
 #else
-#error Only C++11 or later is supported.
+#error Only C++14 or later is supported.
```
chenBright: I don't think this needs to be modified.

darion-yaphet: Agreed 

The C++14 baseline is already enforced by the build system; std::result_of itself doesn’t need a C++14 gate. Reverted.

chenBright (approved):
LGTM

PR #3421, opened 2026-07-31
fix rpcz root client span lifetime (#3420)

### LINKED ISSUE #3420: rpcz does not record root client spans when Controller only keeps a weak_ptr to Span
**Describe the bug**

When rpcz is enabled in a brpc client process, root client-side RPC spans may not be recorded if the RPC is initiated without an existing local parent span.

The client process can successfully enable `/rpcz`, and rpcz internal service calls may appear, but application-level client RPC spans are missing from the rpcz output.

Root cause: `Controller::_span` is stored as `std::weak_ptr<Span>`. For a root client span, there is no local parent span and therefore no parent `_client_list` holding a strong reference. Once the caller-side temporary `shared_ptr<Span>` goes out of scope, the span can be destroyed before `Controller::SubmitSpan()` submits it to rpcz.

**To Reproduce**

1. Enable rpcz in a brpc client process.
2. Send an outgoing RPC from the client without an existing local parent span.
3. Check the client-side `/rpcz` output.
4. Observe that rpcz internal calls may be visible, but the application-level outgoing client RPC span is missing.

**Expected behavior**

Client-side root RPC spans should remain alive until the RPC finishes and `Controller::SubmitSpan()` has a chance to submit them to rpcz.

The outgoing client RPC should be visible in client-side `/rpcz`.

**Versions**

OS:

Compiler:

brpc:

protobuf:

Nested client spans created under a server span are less likely to hit this issue because the parent span owns child spans through `_client_list`.

A possible fix is to let `Controller` hold the current RPC span with `std::shared_ptr<Span>` until the RPC finishes, `SubmitSpan()` runs, or the Controller is reset. Child-to-parent references can remain weak, so this does not introduce a `shared_ptr` cycle.

lh2debug-2:
Fixes #3420

## Summary

This change keeps the current RPC span alive from Controller until the RPC finishes, SubmitSpan runs, or the Controller is reset. This lets root client spans without a local parent be submitted to rpcz instead of being destroyed after the caller-side temporary shared_ptr goes out of scope.

Child client spans remain linked to their parent through weak local-parent references and parent-owned client lists, so they are still serialized under their parent without introducing shared_ptr cycles.

## Testing

- Built brpc successfully with CMake in /tmp/brpc-community-build-rpcz.
- Ran a small local demo test verifying that a root client span remains alive after the caller-side shared_ptr is released while Controller still owns the span.

Note: the local CMake build used a temporary include-only workaround for an unrelated UBShm timer_mgr.cpp atomic_fetch_add/sub compile issue on GCC 8; no UBShm source change is included in this PR.

### DIFF
--- a/src/brpc/controller.cpp
+++ b/src/brpc/controller.cpp
@@ -183,8 +183,8 @@ static void CreateIgnoreAllRead() { s_ignore_all_read = new IgnoreAllRead; }
 // you don't have to set the fields to initial state after deletion since
 // they'll be set uniformly after this method is called.
 void Controller::ResetNonPods() {
-    if (auto span = _span.lock()) {
-        Span::Submit(span, butil::cpuwide_time_us());
+    if (_span) {
+        Span::Submit(_span, butil::cpuwide_time_us());
     }
     _span.reset();
     _error_text.clear();
@@ -463,9 +463,9 @@ void Controller::SetFailed(const std::string& reason) {
         AppendServerIdentiy();
     }
     _error_text.append(reason);
-    if (auto span = _span.lock()) {
-        span->set_error_code(_error_code);
-        span->Annotate(reason);
+    if (_span) {
+        _span->set_error_code(_error_code);
+        _span->Annotate(reason);
     }
     UpdateResponseHeader(this);
 }
@@ -492,9 +492,9 @@ void Controller::SetFailed(int error_code, const char* reason_fmt, ...) {
     va_start(ap, reason_fmt);
     butil::string_vappendf(&_error_text, reason_fmt, ap);
     va_end(ap);
-    if (auto span = _span.lock()) {
-        span->set_error_code(_error_code);
-        span->AnnotateCStr(_error_text.c_str() + old_size, 0);
+    if (_span) {
+        _span->set_error_code(_error_code);
+        _span->AnnotateCStr(_error_text.c_str() + old_size, 0);
     }
     UpdateResponseHeader(this);
 }
@@ -520,9 +520,9 @@ void Controller::CloseConnection(const char* reason_fmt, ...) {
     va_start(ap, reason_fmt);
     butil::string_vappendf(&_error_text, reason_fmt, ap);
     va_end(ap);
-    if (auto span = _span.lock()) {
-        span->set_error_code(_error_code);
-        span->AnnotateCStr(_error_text.c_str() + old_size, 0);
+    if (_span) {
+        _span->set_error_code(_error_code);
+        _span->AnnotateCStr(_error_text.c_str() + old_size, 0);
     }
     UpdateResponseHeader(this);
 }
@@ -982,9 +982,9 @@ void Controller::EndRPC(const CompletionInfo& info) {
     }
     // RPC finished, now it's safe to release `LoadBalancerWithNaming'
     _lb.reset();
-    if (auto span = _span.lock()) {
-        span->set_ending_cid(info.id);
-        span->set_async(_done);
+    if (_span) {
+        _span->set_ending_cid(info.id);
+        _span->set_async(_done);
         // Submit the span if we're in async RPC. For sync RPC, the span
         // is submitted after Join() to get a more accurate resuming timestamp.
         if (_done) {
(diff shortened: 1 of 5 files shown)


wwbmmm:
There is a scenario like this:

1. Server receive a request and create a local parent span
2. During processing this request, a client start a new async call, create a child span
3. Server send the response before the client return, the local parent span submit, and also submit the child span
4. The client's response return, because the client span's parent is now expired, so the client will submit the span, too.
5. Consequently, the client span is submitted twice.

Is that right?

src/brpc/controller.h
```diff
@@ -877,7 +877,7 @@ friend void policy::ProcessThriftRequest(InputMessageBase*);
 private:
     // NOTE: align and group fields to make Controller as compact as possible.
 
-    std::weak_ptr<Span> _span;
+    std::shared_ptr<Span> _span;
```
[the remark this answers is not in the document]
lh2debug-2: done

lh2debug-2:
> There is a scenario like this:
> 
> 1. Server receive a request and create a local parent span
> 2. During processing this request, a client start a new async call, create a child span
> 3. Server send the response before the client return, the local parent span submit, and also submit the child span
> 4. The client's response return, because the client span's parent is now expired, so the client will submit the span, too.
> 5. Consequently, the client span is submitted twice.
> 
> Is that right?

Thanks for pointing this out.

For the exact case where the parent span is submitted while the async child span is still active, the child is not serialized under the parent, because `SpanDB::Index()` only collects inactive child spans. Later, when the child finishes and its local parent has expired, it is submitted as a standalone client span. So that exact case does not duplicate the child.

The real duplicate risk is a nearby race: the child finishes first, the parent dump serializes it into `parent.client_spans`, then the child later submits itself after the parent expires.

I fixed this by making rpcz collection idempotent. Each `Span` now has a submitted flag. Both paths must claim it before writing:

- standalone submit: `Span::Submit()`
- nested child serialization: `SpanDB::Index()`

| Case | Result |
|---|---|
| Root server span | submitted once as root |
| Root client span | submitted once as root |
| Child span, parent alive | not standalone submitted |
| Child ended, parent dumps first | nested under parent; later standalone submit skipped |
| Parent dumps while child active | not nested; later submitted standalone |

Performance impact should be negligible. The flag is `butil::atomic<bool>` with relaxed ordering, `sizeof(Span)` does not increase, and the CAS is only on rpcz submit/dump paths. In the duplicate case it also reduces LevelDB writes by avoiding the extra standalone child record.

src/brpc/controller.cpp
```diff
@@ -183,7 +183,7 @@ static void CreateIgnoreAllRead() { s_ignore_all_read = new IgnoreAllRead; }
 // you don't have to set the fields to initial state after deletion since
 // they'll be set uniformly after this method is called.
 void Controller::ResetNonPods() {
-    if (auto span = _span.lock()) {
+    if (auto span = _span) {
```
chenBright: Just use `_span` directly please.

chenBright (changes requested):

src/brpc/span.h
```diff
@@ -198,6 +199,11 @@ friend class SpanContainer;
 
     void dump_to_db();
     void submit(int64_t cpuwide_us);
+    bool TryMarkSubmitted() const {
```
chenBright: TryMarkSubmitted -> try_mark_submitted

src/brpc/span.cpp
```diff
@@ -581,13 +586,13 @@ inline int GetSpanDB(butil::intrusive_ptr<SpanDB>* db) {
...
-    // Server spans hold shared_ptr references to their child spans (via _client_list),
-    // ensuring child spans remain alive until the server span is submitted and dumped.
-    // Client spans are not submitted here because their lifetime is managed by their
-    // parent server span.
+    // Submit root spans without a local parent. Server spans and root client
+    // spans are submitted independently; child client spans with a live local
+    // parent are serialized under the parent to avoid duplicate submissions.
     if (span->local_parent().expired()) {
-        span->submit(cpuwide_time_us);
+        if (span->TryMarkSubmitted()) {
```
chenBright:
```c++
if (span->local_parent().expired() && span->TryMarkSubmitted()) {
    span->submit(cpuwide_time_us);
}
```

lh2debug-2: done

src/brpc/span.h
lh2debug-2: done

src/brpc/controller.cpp
lh2debug-2: done

lh2debug-2:
Addressed the review comments and verified the changes with tests.

chenBright (approved):
LGTM

PR #3423, opened 2026-08-02
Signal workers for priority tasks

chenBright:

Issue Number: resolve N/A

Problem Summary:

Tasks placed into a global priority queue did not signal worker threads after being enqueued. If all eligible workers were sleeping, these tasks could remain pending until another event woke a worker.

Changed:

Signal one worker after adding a task to the global priority queue. The signal uses the task's tag so that a worker associated with the correct tag is notified.

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/bthread/task_group.cpp
+++ b/src/bthread/task_group.cpp
@@ -1018,8 +1018,11 @@ void TaskGroup::priority_to_run(void* args_in) {
     if (args->meta->priority_index < 0) {
         return g->push_rq(args->meta->tid);
     }
-    return g->control()->push_ed_priority_queue(
+    g->control()->push_ed_priority_queue(
         args->tag, args->meta->priority_index, args->meta->tid);
+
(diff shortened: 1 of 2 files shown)


wwbmmm:
Will this affect performance?

chenBright:
> Will this affect performance?

This will speed up the EventDispatcher's scheduling, and no obvious side effects were observed.

When FLAGS_enable_bthread_priority_queue is false (the default), bthread_start_urgent from the 
EventDispatcher goes through ready_to_run_in_worker / ready_to_run, which would signals a worker.

When it is true, the EventDispatcher bthread is pushed into the global priority queue and can only be 
picked up by another worker via steal_task(), while idle workers stay parked until signal_task() wakes 
them.  So the signal is necessary to get the EventDispatcher rescheduled quickly, which is exactly the 
purpose of the priority queue.

PR #3426, opened 2026-08-03
fix WeightedRandomizedLoadBalancer skipping the last server

Nas01010101:

`WeightedRandomizedLoadBalancer::SelectServer()` never selects the server that was added last, and over-selects the first one.

`Add()` fills `Server::current_weight_sum` with an *inclusive* prefix sum. `SelectServer()` draws `random_weight` from `butil::fast_rand_less_than(weight_sum)`, i.e. from `[0, weight_sum - 1]`, and then `lower_bound()`s that value against those prefix sums. `lower_bound()` returns the first server whose prefix sum is `>= random_weight`, but a server owns the half-open range `[prefix(i-1), prefix(i))`, so the predicate has to be `> random_weight`.

Two consequences: the first server also serves `random_weight == prefix(0)`, so it gets one slot too many; and the last server is never selected at all, because `random_weight` can never reach `weight_sum`.

With four servers of equal weight the measured distribution is 49.8 / 25.1 / 25.1 / 0.0 percent instead of 25 percent each. The last server only receives traffic through the fallback loop, when every server picked at random happens to be unavailable.

Changed: search for `random_weight + 1`, so `lower_bound()` lands on the first prefix sum strictly greater than `random_weight`. `random_weight + 1` is at most `weight_sum`, which is exactly the last prefix sum, so the iterator is still always valid.

Side effects:
- Performance effects: none, the comparison is unchanged.
- Breaking backward compatibility: no. Traffic shifts towards the configured weights.

The existing `weighted_randomized` test does not catch this. Its servers have weights 3/2/5/10 and it only asserts that every rate is within 0.5x~2x of the expected one. The weight-10 server measures 0.453 before this change and 0.5055 after it, and both are inside that band. The off-by-one predicts that number exactly: the server should own 10 of 20 slots but owns 9, i.e. 0.45.

This PR adds `weighted_randomized_equal_weight`, which uses equal weights so a single misplaced slot is visible, and checks the rates within 0.9x~1.1x. On master the last server is selected 0 times out of 40000 and the test fails; with this change all four land within 0.2476~0.25435.

### Check List:

- `brpc_load_balancer_unittest` passes in full: `[  PASSED  ] 17 tests.`
- The new test was run with `--gtest_repeat=40` without a failure.
- Verified the new test fails without the one-line change and passes with it.

Tested on macOS/arm64 with clang. I was not able to build on Linux/gcc locally; CI will cover that.

### DIFF
--- a/src/brpc/policy/weighted_randomized_load_balancer.cpp
+++ b/src/brpc/policy/weighted_randomized_load_balancer.cpp
@@ -131,13 +131,11 @@ int WeightedRandomizedLoadBalancer::SelectServer(const SelectIn& in, SelectOut* 
     uint64_t weight_sum = s->weight_sum;
     for (size_t i = 0; i < n; ++i) {
         uint64_t random_weight = butil::fast_rand_less_than(weight_sum);
-        // current_weight_sum is an inclusive prefix sum, so server i owns the
-        // half-open range [prefix(i-1), prefix(i)). random_weight belongs to the
-        // first server whose prefix sum is strictly greater than it, which is
-        // lower_bound() of random_weight + 1 rather than of random_weight itself.
-        const Server random_server(0, 0, random_weight + 1);
+        // current_weight_sum is an inclusive prefix sum, so random_weight belongs
+        // to the first server whose prefix sum is strictly greater than it.
+        const Server random_server(0, 0, random_weight);
         const auto& server =
-            std::lower_bound(s->server_list.begin(), s->server_list.end(),
+            std::upper_bound(s->server_list.begin(), s->server_list.end(),
                              random_server, server_compare);
         const SocketId id = server->id;
         if (ExcludedServers::IsExcluded(in.excluded, id)) {
--- a/src/brpc/policy/weighted_randomized_load_balancer.cpp
+++ b/src/brpc/policy/weighted_randomized_load_balancer.cpp
@@ -131,7 +131,11 @@ int WeightedRandomizedLoadBalancer::SelectServer(const SelectIn& in, SelectOut* 
     uint64_t weight_sum = s->weight_sum;
     for (size_t i = 0; i < n; ++i) {
         uint64_t random_weight = butil::fast_rand_less_than(weight_sum);
-        const Server random_server(0, 0, random_weight);
+        // current_weight_sum is an inclusive prefix sum, so server i owns the
+        // half-open range [prefix(i-1), prefix(i)). random_weight belongs to the
+        // first server whose prefix sum is strictly greater than it, which is
+        // lower_bound() of random_weight + 1 rather than of random_weight itself.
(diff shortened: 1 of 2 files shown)


src/brpc/policy/weighted_randomized_load_balancer.cpp
```diff
@@ -131,7 +131,11 @@ int WeightedRandomizedLoadBalancer::SelectServer(const SelectIn& in, SelectOut*
...
         uint64_t random_weight = butil::fast_rand_less_than(weight_sum);
-        const Server random_server(0, 0, random_weight);
+        // current_weight_sum is an inclusive prefix sum, so server i owns the
+        // half-open range [prefix(i-1), prefix(i)). random_weight belongs to the
+        // first server whose prefix sum is strictly greater than it, which is
+        // lower_bound() of random_weight + 1 rather than of random_weight itself.
+        const Server random_server(0, 0, random_weight + 1);
         const auto& server =
             std::lower_bound(s->server_list.begin(), s->server_list.end(),
                              random_server, server_compare);
```
chenBright, line 139: I'd suggest expressing the intent directly with `upper_bound` instead. The whole
point is "find the first prefix sum strictly greater than `random_weight`", and that is
precisely what `upper_bound` means. The `+ 1` form encodes the same thing indirectly, which
is why it needs four lines of comment to explain, and it also makes every reader stop and
re-check whether the increment can overflow.

Nas01010101:
@chenBright switched to `upper_bound(random_weight)`. You are right that it says the thing directly — the comment is down to two lines now and there is no increment left to re-check.

The two forms select the same server for every input. `server_compare` orders by `current_weight_sum`, so `upper_bound` calls it as `random_weight < prefix(i)` and lands on the first prefix strictly greater than `random_weight`, which is what `lower_bound(random_weight + 1)` was computing. I checked it exhaustively over every value of `random_weight` in `[0, weight_sum)` for a set of weight layouts (all-equal, ascending, one dominant server, single server, and a two-server case): 0 disagreements between the two forms, 13 disagreements against the pre-fix `lower_bound(random_weight)`, and each server owns exactly `weight` values of `random_weight`.

On the overflow point — it is moot now, but it could not fire before either: `random_weight` comes from `fast_rand_less_than(weight_sum)` so it is at most `weight_sum - 1`, and `weight_sum` is a `uint64_t` accumulated from `uint32_t` weights.

Also fixed the tolerance comment @Copilot flagged in the test. With `run_times=40000` and `p=0.25`, sigma is `sqrt(n*p*(1-p))` = 86.6, so the 0.9x~1.1x band is about 11 sigma, not more than 20. The comment now shows the arithmetic instead of asserting a number.

chenBright (approved):
LGTM

Nas01010101:
> LGTM

Ty !

PR #3428, opened 2026-08-07
feat: add URMA transport support

zchuango:
About discussion：

> 
> https://github.com/apache/brpc/issues/3167 —— Socket-Transport 架构改造共识
> https://github.com/apache/brpc/discussions/3217 —— 路线 B（URMA-UrmaTransport）与路线 C（双后端）共识
> https://github.com/apache/brpc/issues/3226 —— OBMM/ShmTransport 实现方案
> 本提案对应 https://github.com/apache/brpc/discussions/3217 中的路线 B，并与https://github.com/apache/brpc/issues/3226 的 OBMM-ShmTransport 形成路线 C（双后端协同）。

Issue Number: Related to #3401

Problem Summary:

brpc does not currently provide a remote-memory transport based on the UMDK
URMA API.

This PR adds URMA (Unified Remote Memory Access) as an optional transport in
the existing `Transport` framework. TCP is used for connection establishment
and URMA capability/resource negotiation. After both peers successfully
negotiate URMA support, the data path switches to URMA. When URMA is
unavailable or negotiation does not succeed, the connection falls back to TCP.

The initial implementation supports the `baidu_std` protocol only and is
disabled by default.

Changed:

- Add `SOCKET_MODE_URMA` and `UrmaTransport`, and integrate them with the
  existing transport, socket, and message-processing framework.
- Add `UrmaEndpoint` for URMA resource management, SEND/RECV processing,
  completion handling, and credit-based flow control.
- Add TCP-based URMA capability and resource negotiation, including v2 binary
  and v3 protobuf handshake formats, with TCP fallback.
- Add a registered buffer pool for IOBuf blocks and APIs for explicit user
  memory registration.
- Support completion-event mode and optional busy-polling mode.
- Support the UMDK bonding extension when the corresponding provider extension
  is available.
- Add optional URMA build integration for CMake, Make, and Bazel, together with
  a link-time URMA mock for build and test environments without `liburma` or
  URMA hardware.
- Add URMA unit tests, English and Chinese documentation, and an
  `urma_performance` example.

Side effects:

- Performance effects:
  - URMA is optional and disabled by default. Existing socket modes and their
    default behavior remain unchanged.
  - Enabling URMA maps and registers a configurable IOBuf buffer pool. The
    default configuration is 65,536 buffers of 8 KiB each (512 MiB in total).
  - Busy-polling mode may increase CPU usage because completion pollers
    actively poll the JFC.
  - URMA performance depends on the provider, driver, hardware, and polling
    configuration. This PR does not claim a specific hardware performance
    improvement.

  - No intended breaking changes.
  - `SOCKET_MODE_URMA` and the URMA memory-registration APIs are additive.
    Existing socket modes and their default behavior remain unchanged.

---
### Check List:


wwbmmm:
I'm not familiar with URMA, but it seems that the code in `src/brpc/urma` is similar to `src/brpc/rdma`, is that anything we can reuse?

zchuango:
> I'm not familiar with URMA, but it seems that the code in is similar to , is that anything we can reuse?`src/brpc/urma``src/brpc/rdma`

@wwbmmm Thanks for pointing this out. Yes, the current user-space [liburma](https://atomgit.com/openeuler/umdk/tree/master/src/urma/lib/urma/core) API largely aligns with the RDMA interface. The differences are mostly pushed down to the kernel-space [uburma](https://atomgit.com/openeuler/kernel/tree/OLK-6.6/drivers/ub/urma/uburma) , [ubcore](https://atomgit.com/openeuler/kernel/tree/OLK-6.6/drivers/ub/urma/ubcore) library.
Both URMA and RDMA follow DMA semantics in their programming models; the primary difference is the underlying transport protocol.
For reference, please see:
[UB Bus documentation](https://www.unifiedbus.com/zh)
[URMA open-source software](https://atomgit.com/openeuler/umdk)

zchuango:
The common handshake abstraction is being discussed in:#3432

docs/cn/urma.md
```diff
@@ -0,0 +1,149 @@
...
+## 技术背景
+
+URMA 在 UMDK 支持的设备上提供 verbs 风格接口。当前实现创建可靠消息
+（`URMA_TM_RM`）Jetty，并使用 CTP 传输路径；通过
+`urma_post_jetty_send_wr` 提交发送 WR，通过 `urma_post_jfr_wr` 提交接收
+WR。完成事件既可由 JFC 忙轮询获取，也可通过 JFCE 事件 fd 获取。
+
+## 编译配置
+
+### CMake 编译
```
dwh110, line 20: Bazel‑specific build instructions may be added.

docs/cn/urma.md
```diff
@@ -0,0 +1,149 @@
...
+
+# 编译 urma_performance 示例
+cd example/urma_performance
+cmake -B build
+make -C build -j$(nproc)
+```
+
+`WITH_URMA=ON` 使用上游 UMDK 头文件进行编译。CMake 优先使用系统安装的
+SDK；找不到头文件时，会参照 Mooncake 的 mock 构建方式下载固定版本的
+UMDK，可通过 `DOWNLOAD_URMA_HEADERS=OFF` 禁止下载。找到 `liburma` 时使用
```
dwh110: DOWNLOAD_URMA_HEADERS lacks Bazel build support

src/brpc/urma/mock_urma.cpp
```diff
@@ -0,0 +1,713 @@
...
+urma_eid_info_t mock_eid_info = {
+    .eid = {{0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0A, 0x0B,
+             0x0C, 0x0D, 0x0E, 0x0F, 0x10}},
+    .eid_index = 0};
+
+}  // namespace
+
+extern "C" {
+
+urma_status_t urma_init(urma_init_attr_t *init_attr) {
```
dwh110, line 121: The mock functionality lacks an independent feature switch, which can break normal URMA functionality.

zchuango:
@dwh110 Thanks for the review and suggestion. I have noted this concern and will evaluate the proposed changes. I will update the PR once the necessary modifications are implemented.

docs/cn/urma.md
zchuango: done

docs/cn/urma.md
zchuango: done

src/brpc/urma/mock_urma.cpp
zchuango: @dwh110 done

zchuango:
@wwbmmm @chenBright @dwh110 

I have addressed the main review feedback:
- Added Bazel support for disabling UMDK header downloads.
- Added explicit mock controls for CMake, Make, and Bazel.
- Ensured real liburma builds exclude the mock implementation.
- Updated the documentation and pinned the UMDK dependency.
- Verified the Bazel build and all 61 test targets.

Please take another look when available. Kindly manually verify any potential issue before posting a review comment. please feel free ping me anytime.

wwbmmm:
Will this issue : https://github.com/apache/brpc/discussions/3432 resolved in this PR, or in further PR?

zchuango:
@wwbmmm We plan to handle #3432 in a follow-up PR.

We're currently testing the common handshake approach internally with RDMA and UBShm. For URMA, we'd like to get this PR merged first, since the handshake work will depend on how this PR progresses.

Once this pr merged, we plan to move the URMA handshake to the common handshake framework as well.

If there are any other concerns with this PR, feel free to ping me.

wwbmmm:
LGTM

src/brpc/urma/urma_endpoint.cpp
```diff
@@ -0,0 +1,1846 @@
...
+                    const ssize_t nr = ep->HandleCompletion(crs[i]);
+                    if (nr < 0) {
+                        return errno ? errno : EIO;
+                    }
+                    bytes += nr;
+                }
+            }
+        };
+
+        int completion_error = drain_cq();
```
dwh110, line 1137: Should it be activated only in polling mode， not in event_mode?

zchuango: This is mainly an optimization for `event_mode`. Once we get an event notification, we drain all currently available events in one go, which helps reduce event handling latency.

wwbmmm:
Overall the new UrmaTransport subsystem is well-structured and defensively written (handshake parsing bounds-checks, flow-control credits with overflow validation, robust TCP fallback). One concern in the Bazel build is worth addressing before merge.

- `BUILD.bazel:0` The cc_library `srcs` list appends `src/brpc/urma/mock_urma.cpp` unconditionally (in the base source list) and again in the `brpc_with_urma` select. Unlike the Makefile (`URMA_USE_MOCK=0` filters it out) and CMakeLists (`list(REMOVE_ITEM BRPC_SOURCES ... mock_urma.cpp )` when `NOT URMA_USE_MOCK_ENABLED`), the `brpc_with_urma_use_real` select maps to `[]` which does NOT remove the mock translation unit. Since mock_urma.cpp is guarded by `#if BRPC_WITH_URMA`, a real-liburma Bazel build (`--define=BRPC_WITH_URMA=true --define=BRPC_URMA_USE_MOCK=false -lurma`) would still compile and link the `urma_*` mock symbols and clash with the real `liburma`. Please exclude `mock_urma.cpp` under `brpc_with_urma_use_real` (e.g. via a `select` that maps the real case to an empty srcs list), matching the CMake/Make real-library behavior; also note the base list already has it so adding it again in `brpc_with_urma` is redundant.

---

zchuango:
@wwbmmm Thanks for pointing this out. Addressed in commit 07c4661d.
The Bazel source selection has been restructured:
- All src/brpc/urma/*.cpp files are excluded from the base source glob.
- URMA_SRCS explicitly excludes mock_urma.cpp.
- brpc_with_urma_use_real includes only the non-mock URMA sources and links liburma.
- The regular brpc_with_urma configuration appends mock_urma.cpp exactly once.
- URMA implementation sources are no longer compiled when URMA is disabled.

This now matches the CMake and Make behavior and prevents mock urma_* symbols from conflicting with the real liburma. bazel query successfully loaded and expanded the updated source selection.

zchuango:
@wwbmmm Addressed the URMA-related Copilot review findings in fe6ea6f3:
- Route writes through URMA only after negotiation reaches URMA_ON; TCP remains in use while the state is UNKNOWN.
- Replace the non-portable pthread_t reinterpret cast with std::hash<pthread_t>.
- Use GFLAGS_NAMESPACE::FlagSaver in the handshake-version test.
- Adapt UrmaEndpoint to the current per-stream InputMessengerProcessor model: URMA completions use a dedicated STREAM_URMA_JETTY processor, while fallback bytes return to the TCP fd processor.
- Clarify the dedicated transport poller comment for RDMA/UBRING/URMA.

src/brpc/input_messenger.cpp
```diff
@@ -15,35 +15,33 @@
...
+#include "butil/logging.h"    // CHECK
+#include "butil/time.h"       // cpuwide_time_us
+#include "bvar/bvar.h"        // bvar::Adder
+#include <gflags/gflags.h>
+
 
 namespace brpc {
 
-InputMessenger* g_messenger = nullptr;
+InputMessenger *g_messenger = nullptr;
```
wwbmmm: There's no need to change these formats

src/brpc/input_messenger_processor.cpp
```diff
@@ -15,279 +15,286 @@
...
+  case InputMessengerProcessor::STREAM_RDMA_QP:
+    return "rdma_qp";
+  case InputMessengerProcessor::STREAM_URMA_JETTY:
+    return "urma_jetty";
+  }
+  return "unknown";
 }
 
-InputMessengerProcessor::ParsingStreamGuard::ParsingStreamGuard(Socket* socket, StreamType type)
+InputMessengerProcessor::ParsingStreamGuard::ParsingStreamGuard(Socket *socket,
```
wwbmmm: don't change these formats

zchuango:
@wwbmmm  I just noticed a conflict between the format and master of these two files: `input_messenger.cpp` and `input_messenger_processor.cpp`. Now has been fixed. Please review them again.

src/brpc/urma/urma_endpoint.cpp
```diff
@@ -0,0 +1,1849 @@
...
+#include "butil/atomicops.h"
+#include "butil/iobuf.h"
+#include "butil/logging.h"
+#include "butil/macros.h"
+#include "butil/sys_byteorder.h"
+#include "butil/time.h"
+#include "bthread/bthread.h"
+#include "bthread/butex.h"
+
+#include "urma_api.h"
```
chenBright, line 46: #include "brpc/urma/urma_api.h"

src/brpc/urma/urma_endpoint.cpp
```diff
@@ -0,0 +1,1849 @@
...
+    if (pool) {
+        std::memcpy(out->seg_eid, pool->seg.ubva.eid.raw, 16);
+        out->seg_uasid = pool->seg.ubva.uasid;
+        out->seg_va = pool->seg.ubva.va;
+        out->seg_len = pool->seg.len;
+        out->seg_token_id = pool->seg.token_id;
+    }
+}
+
+void UrmaEndpoint::FillLocalHelloV2(v2_wire::HelloMessage* out) const {
```
chenBright, line 277: Why are there two versions (v2 and v3) of the new transport?

src/brpc/urma/urma_helper.h
```diff
@@ -0,0 +1,119 @@
...
+#include <functional>
+#include <string>
+
+#include "bthread/types.h"
+#include "butil/atomicops.h"
+
+#if BRPC_WITH_URMA
+
+#include "urma_api.h"
+#include "urma_types.h"
```
chenBright, line 32: #include "brpc/urma/urma_api.h"
#include "brpc/urma/urma_types.h"

Gzure: The Urma header file dependency environment has the urma-devel package installed, but the Urma header files are not placed in brpc.

src/brpc/urma/urma_endpoint.cpp
Gzure: The Urma header file dependency environment has the urma-devel package installed, but the Urma header files are not placed in brpc.

src/brpc/urma/urma_endpoint.cpp
Gzure: Urma supports v2 and v3 to remain consistent with RDMA. If v2 is confirmed to no longer support new transports, only the implementation of v3 can be retained.

chenBright: I don't think it's necessary to support v2 for the new transports.

src/brpc/urma/urma_helper.h
chenBright: ok

src/brpc/urma/urma_endpoint.cpp
chenBright: ok

src/brpc/urma/urma_endpoint.cpp
Gzure: > I don't think it's necessary to support v2 for the new transports.

I have deleted the urma v2 related code.

zchuango:
@wwbmmm @chenBright I would like to merge #3428 first so I can continue adapting the URMA handshake to the common handshake framework in #3535. If there are no further blocking concerns with the current version, I’ll go ahead and merge #3428.

chenBright (approved):
LGTM

PR #3447, opened 2026-08-14
Suppress false-positive LSan leaks in Span pool factory functions

chenBright:

Issue Number: resolve 

Problem Summary:

`brpc_controller_unittest` fails under LeakSanitizer/ASan with reports like:

```
==ERROR: LeakSanitizer: detected memory leaks
Direct leak ... in brpc::Span::CreateClientSpan(...) src/brpc/span.cpp:208
SUMMARY: AddressSanitizer: 110 byte(s) leaked in 4 allocation(s).
```

`Span` objects are recycled through `butil::object_pool` and are never truly
freed; `SpanDeleter` returns them to the pool instead of releasing them. As a
result, the `shared_ptr` control block, the `_full_method_name` string, and the
pooled `Span` itself are kept alive by the pool by design. LeakSanitizer cannot
follow the pointers held inside the pool and therefore reports these intentional
retentions as leaks.

Changed:

Add `ANNOTATE_SCOPED_MEMORY_LEAK` at the entry of the span factory functions
that allocate through the object pool, so LeakSanitizer ignores these intentional retentions:
  - `Span::CreateClientSpan`
  - `Span::CreateBthreadSpan`
  - `Span::CreateServerSpan` (the overload that performs the allocation)

This reuses the existing suppression mechanism already used in `Server::InitializeOnce` 
for `_keytable_pool`.

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/brpc/span.cpp
+++ b/src/brpc/span.cpp
@@ -31,6 +31,7 @@
 #include "butil/fast_rand.h"
 #include "butil/file_util.h"
 #include "butil/files/file_enumerator.h"
+#include "butil/debug/leak_annotations.h"
 #include "brpc/shared_object.h"
 #include "brpc/reloadable_flags.h"
 #include "brpc/span.h"
@@ -201,6 +202,11 @@ Span::~Span() {
 
 std::shared_ptr<Span> Span::CreateClientSpan(const std::string& full_method_name,
                                              int64_t base_real_us) {
+    // Span objects are recycled via butil::object_pool and are never truly
+    // freed; the shared_ptr control block and other allocations made here are
+    // kept alive by that pool by design. Instruct LeakSanitizer to ignore
+    // these intentional "leaks".
+    ANNOTATE_SCOPED_MEMORY_LEAK;
     Span* span_raw = butil::get_object<Span>(Forbidden());
     if (__builtin_expect(span_raw == NULL, 0)) {
         return nullptr;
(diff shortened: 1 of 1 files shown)


ZhengweiZhu:
Thanks for investigating this. I reached a different conclusion after tracing the lifetime in the failing tests.

The reported allocations are the root client spans submitted to the asynchronous bvar collector. `Controller::Reset` / `Controller::SubmitSpan` clears `Controller::_span`, but the collector can retain its reference until the next collection cycle. The short-lived unit-test binary may exit before that happens, so LeakSanitizer reports the still-pending span allocations.

Because the failure is caused by asynchronous test teardown, I think the fix should be scoped to the tests rather than suppressing LeakSanitizer for every allocation made by `CreateClientSpan`, `CreateBthreadSpan`, and `CreateServerSpan`. A factory-wide `ANNOTATE_SCOPED_MEMORY_LEAK` also risks hiding genuine leaks in those paths.

I opened #3448 with a narrower approach: after `Reset` and `SubmitSpan`, the tests wait with a bounded deadline for their existing `weak_ptr` to expire. This:

- verifies the intended lifetime contract instead of merely suppressing the report;
- avoids a fixed sleep;
- preserves LeakSanitizer coverage for the span factory paths;
- fails explicitly if the collector does not release the span within the deadline.

PR: https://github.com/apache/brpc/pull/3448

PR #3445, opened 2026-08-14
build: complete UBRING Bazel and CI support

ZhengweiZhu:
Add the missing BRPC_WITH_UBRING Bazel configuration, wire it through the library and example targets, and document the supported build commands.

Enable RDMA and UBRING in CI jobs whose all-options configurations did not exercise those features. Run both feature suites in the existing Bazel unit-test jobs and enable RDMA in the Make unit-test job.

Use unsigned literals for UBRING atomic counter operations to match the counter type and avoid template deduction failures on stricter compilers.

PR #3448, opened 2026-08-14
test: wait for asynchronous span collection

ZhengweiZhu:

Issue Number: N/A

Problem Summary:

`brpc_controller_unittest` may fail under LeakSanitizer with a 110-byte leak report after the root client span tests. The spans have already been submitted to the asynchronous bvar collector, but the short-lived test process can exit before the collector releases its references.

Changed:

- Add a bounded `WaitForSpanToExpire` helper that polls the test's `weak_ptr` for at most 5 seconds.
- Wait for asynchronous span collection after both `Controller::Reset` and `Controller::SubmitSpan`.
- Assert the observable lifetime contract instead of suppressing LeakSanitizer reports for all span factory allocations.

Side effects:
- Performance effects: Test-only. The two affected tests normally wait for approximately one collector interval and fail after a 5-second deadline if collection does not complete.

---
### Check List:

### DIFF
--- a/test/brpc_controller_unittest.cpp
+++ b/test/brpc_controller_unittest.cpp
@@ -42,6 +42,14 @@ void MyCancelCallback(bool* cancel_flag) {
     *cancel_flag = true;
 }
 
+bool WaitForSpanToExpire(const std::weak_ptr<brpc::Span>& span) {
+    const int64_t deadline_us = butil::gettimeofday_us() + 5000000L;
+    while (!span.expired() && butil::gettimeofday_us() < deadline_us) {
+        usleep(1000);
+    }
+    return span.expired();
+}
+
 TEST_F(ControllerTest, notify_on_failed) {
     brpc::SocketId id = 0;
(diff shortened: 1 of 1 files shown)


ZhengweiZhu:
CI update: all checks passed except `gcc-unittest-with-bazel`, which failed while running the unrelated `bthread_priority_queue_unittest` target. The first two `PriorityQueueTest` cases passed, then the process terminated during `PriorityQueueTest.start_foreground_priority_to_run` without a normal gtest assertion message.

This PR only changes `brpc_controller_unittest.cpp`, and Bazel runs these test targets in separate processes. The target check for this fix, `clang-unittest-asan`, passed. The same GCC/Bazel job also passed on the current upstream master CI, PR #3423, PR #3445, and PR #3447, so this appears to be an intermittent priority-queue scheduling failure rather than a regression from this PR.

Failed job: https://github.com/apache/brpc/actions/runs/31818586145/job/94826110622

 Could you maintainer please rerun the failed job? @chenBright

chenBright (approved):
LGTM

PR #3449, opened 2026-08-15
Fix SOFA PBRPC parser not limiting metadata size

wwbmmm:

Issue Number: null

Problem Summary:

The SOFA PBRPC parser (`ParseSofaMessage`) enforces `FLAGS_max_body_size`
only on the frame's `body_size` field. The `meta_size` field, and
therefore the whole frame size, is not bounded. A peer can send a frame
with a large `meta_size` and `body_size = 0`, which passes the current
check, so the connection keeps reading and buffering the declared frame
until the metadata is finally parsed (and rejected). This makes the
configured per-message limit ineffective and can consume much more memory
than expected.

Changed:

- In `ParseSofaMessage`, reject frames whose `meta_size` exceeds
  `FLAGS_max_body_size`, in addition to the existing `body_size` check.
- Add unit tests in `test/brpc_sofa_pbrpc_protocol_unittest.cpp` covering
  an oversized body and an oversized metadata block.

Side effects:

- Performance effects: none.
- Breaking backward compatibility: SOFA frames whose metadata exceeds
  `max_body_size` (previously buffered and then rejected at protobuf
  parsing time) are now rejected earlier with a "too big data" error and
  the connection is closed.

---

### Check List:

Tests run:

- `test/brpc_sofa_pbrpc_protocol_unittest` (all cases passed)

### DIFF
--- a/src/brpc/policy/sofa_pbrpc_protocol.cpp
+++ b/src/brpc/policy/sofa_pbrpc_protocol.cpp
@@ -189,11 +189,12 @@ ParseResult ParseSofaMessage(butil::IOBuf* source, Socket* socket,
                    << " + body_size=" << body_size;
         return MakeParseError(PARSE_ERROR_TRY_OTHERS);
     }
-    if (body_size > FLAGS_max_body_size) {
-        // We need this log to report the body_size to give users some clues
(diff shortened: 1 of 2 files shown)


chenBright (commented):
LGTM

PR #3450, opened 2026-08-15
Fix bug when parsing zero-length string field in mcpack2pb

wwbmmm:

Issue Number: null

Problem Summary:

A malformed mcpack2/compack request that contains a string field with a
zero value_size crashes the server. In `UnparsedValue::as_string()`,
`_size` (the string field's value_size, taken directly from the input)
is used in `out->resize(_size - 1)` without being validated first. When
value_size is 0, `_size - 1` underflows to SIZE_MAX, `resize()` throws
`std::length_error`, and since the exception is not caught anywhere on
the request parsing path the whole process aborts. This is reachable
from `UbrpcAdaptor::ParseNsheadMeta` and
`UbrpcAdaptor::ParseRequestFromIOBuf` for servers wired with the
nshead + ubrpc mcpack2 adaptor.

Changed:

- `src/mcpack2pb/parser.cpp`: reject string fields whose value_size is 0
  in `UnparsedValue::as_string()` by marking the input stream bad and
  returning, instead of underflowing and throwing.
- `test/brpc_mcpack2pb_unittest.cpp`: add unit tests covering the
  zero-size string field case and a normal string field parse.

Side effects:
- Performance effects: none (one extra comparison per string field).
- Breaking backward compatibility: none. Well-formed input is parsed
  exactly as before; only previously-crashing malformed input now fails
  gracefully.

---
### Check List:

### DIFF
--- a/src/mcpack2pb/parser.cpp
+++ b/src/mcpack2pb/parser.cpp
@@ -580,7 +580,9 @@ void UnparsedValue::as_string(std::string* out, const char* var) {
     if (_size < 1) {
         // A string field must contain at least the trailing '\0'.
         // Reject _size == 0 here, otherwise `_size - 1' underflows and
-        // resize() throws an uncaught exception.
+        // resize() throws an uncaught exception. Clear `out' so callers
+        // that reuse the string do not keep a stale value.
+        out->clear();
         _stream->set_bad();
         return;
     }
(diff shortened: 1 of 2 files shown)


chenBright (approved):
LGTM

PR #3451, opened 2026-08-15
Limit mcpack2pb array item count to the actual payload size

wwbmmm:

Issue Number: resolve

Problem Summary:
The mcpack2pb parser trusts the item count stored in an mcpack array
header without validating it against the actual payload. The code
generated by the mcpack2pb protoc plugin uses this count to preallocate
memory for repeated protobuf fields via `Reserve()`. A malformed request
that declares a huge item count (e.g. `0x7fffffff`) can therefore make
the server preallocate about 16GB of virtual memory, which may crash the
process on memory-constrained hosts.

Changed:
- Cap `ArrayIterator::item_count()` by the remaining bytes of the array
  in `mcpack2pb::ArrayIterator::init()`.
- Add unit tests covering huge and inconsistent item counts.

Side effects:
- Performance effects: none for well-formed input; the item count of a
  well-formed array never exceeds the remaining bytes.
- Breaking backward compatibility: none. The item count returned for
  malformed input is now clamped, but such input previously caused
  incorrect large preallocations or parse failures anyway.

---
### Check List:

### DIFF
--- a/src/mcpack2pb/parser-inl.h
+++ b/src/mcpack2pb/parser-inl.h
@@ -164,6 +164,13 @@ inline void ArrayIterator::init(InputStream* stream, size_t size) {
         return set_bad();
     }
     _item_count = items_head.item_count;
+    // The item count is read from the request and may be much larger than
+    // the actual payload. The generated code uses it to Reserve() memory for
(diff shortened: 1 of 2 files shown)


chenBright (approved):
LGTM

PR #3452, opened 2026-08-15
Fix RTMP abort message deleting the chunk stream being parsed

wwbmmm:

Issue Number: null

Problem Summary:

An RTMP Abort control message that references the chunk stream carrying
the message itself made the parser free that chunk stream while it was
still being processed. `RtmpChunkStream::Feed()` continues to access the
freed chunk stream right after `OnMessage()` returns, which may crash the
server or corrupt memory.

Changed:

- `RtmpChunkStream::OnAbortMessage` now skips `ClearChunkStream` when the
  abort targets the chunk stream that is currently being parsed.
- Added a unit test that feeds a self-referencing Abort message and
  verifies the chunk stream survives.

Side effects:
- Performance effects: None.

---
### Check List:


chenBright (approved):
LGTM

PR #3458, opened 2026-08-16
Refactor NULL with nullptr in brpc/policy

chenBright:

Issue Number: resolve 

Problem Summary:

Changed:

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/brpc/policy/auto_concurrency_limiter.cpp
+++ b/src/brpc/policy/auto_concurrency_limiter.cpp
@@ -97,7 +97,7 @@ AutoConcurrencyLimiter::AutoConcurrencyLimiter()
 }
 
 AutoConcurrencyLimiter* AutoConcurrencyLimiter::New(const AdaptiveMaxConcurrency&) const {
-    return new (std::nothrow) AutoConcurrencyLimiter;
+    return new AutoConcurrencyLimiter;
 }
 
 bool AutoConcurrencyLimiter::OnRequested(int current_concurrency, Controller*) {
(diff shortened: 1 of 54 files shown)


src/brpc/policy/mysql/mysql_auth_packet.cpp
```diff
@@ -43,7 +43,7 @@ size_t DecodeLengthEncodedInt(const butil::StringPiece& buf, uint64_t* out,
         return 1;
     }
     if (first == 0xfb) {
-        // 0xFB is the lenenc NULL marker, not a length prefix.  Report NULL
+        // 0xFB is the lenenc nullptr marker, not a length prefix.  Report nullptr
```
wasphin: Some `NULL`s in the comments are protocol-level terms, not C++ 
null pointers.

chenBright: Fixed.

src/brpc/policy/auto_concurrency_limiter.cpp
```diff
@@ -97,7 +97,7 @@ AutoConcurrencyLimiter::AutoConcurrencyLimiter()
 }
 
 AutoConcurrencyLimiter* AutoConcurrencyLimiter::New(const AdaptiveMaxConcurrency&) const {
-    return new (std::nothrow) AutoConcurrencyLimiter;
+    return new AutoConcurrencyLimiter;
```
wasphin, line 100: Not sure if this change is intentional or accidental?

wasphin: There are about 5 places with this change

chenBright: It's intentional. If `new` fails, it's basically unrecoverable; just throw a crash exception.

wasphin:
LGTM

src/brpc/policy/auto_concurrency_limiter.cpp
wasphin: Better to split this into different commits.

Issue #13, opened 2017-09-18
Brpc cannot link on Ubuntu LTS 16.04


beenchaos:
Linking echo_client
../..//output/lib/libbrpc.a(channel.o): In function `std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > butil::class_name_str<brpc::Describable>(brpc::Describable const&)':
/home/brpc/brpc/./src/butil/class_name.h:52: undefined reference to `butil::demangle(char const*)'
../..//output/lib/libbrpc.a(load_balancer.o): In function `std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > butil::class_name_str<brpc::NonConstDescribable>(brpc::NonConstDescribable const&)':
/home/brpc/brpc/./src/butil/class_name.h:52: undefined reference to `butil::demangle(char const*)'
/home/brpc/brpc/./src/butil/class_name.h:52: undefined reference to `butil::demangle(char const*)'
../..//output/lib/libbrpc.a(socket.o): In function `std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > butil::class_name_str<brpc::Describable>(brpc::Describable const&)':
/home/brpc/brpc/./src/butil/class_name.h:52: undefined reference to `butil::demangle(char const*)'
../..//output/lib/libbrpc.a(socket.o): In function `std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > butil::class_name_str<brpc::SocketUser>(brpc::SocketUser const&)':
/home/brpc/brpc/./src/butil/class_name.h:52: undefined reference to `butil::demangle(char const*)'
../..//output/lib/libbrpc.a(socket.o):/home/brpc/brpc/./src/butil/class_name.h:52: more undefined references to `butil::demangle(char const*)' follow
collect2: error: ld returned 1 exit status
Makefile:34: recipe for target 'echo_client' failed
make: *** [echo_client] Error 1

chenzhangyi:
Looks like it's duplicated with https://github.com/brpc/brpc/issues/6.  Please try update brpc.

PR #3468, opened 2026-08-20
docs: add Bazel build guide and bzlmod example

### LINKED ISSUE #2991: 官方文档建议增加bazel的编译介绍，目前只有cmake和config_brpc.sh的步骤

官方文档目前只有cmake和config_brpc.sh的步骤
建议增加bazel的编译介绍，或者描述通用的编译步骤方便开发者理解。

开发者真正的理解之后可以自行编写适合自己的编译流程。

zchuango:

Issue Number: resolve apache/brpc#2991

Problem Summary:

The build documentation currently focuses on config_brpc.sh and CMake. It does not provide a clear Bazel/bzlmod build path for users who want to build bRPC or use bRPC as a Bazel dependency.

Changed:
- Add Bazel/bzlmod build instructions to the English and Chinese getting started docs.
- Rewrite the English and Chinese Bazel support docs with bzlmod-first dependency guidance.
- Document the required registries and root MODULE.bazel setup for using bRPC as a Bazel dependency.
- Add an echo client target to example/build_with_bazel_module so the bzlmod example has both server and client targets.
- Add required leveldb and openssl single_version_override entries for aarch64 bzlmod builds, and document that users must repeat these overrides in their own root MODULE.bazel.

Side effects:
- Performance effects: None. This PR only changes documentation and the Bazel example.

Validation:
- GitHub Actions passed on PR #13.
- Linux x64: Bazel all-options jobs passed with both gcc and clang, including the external bzlmod example under `example/build_with_bazel_module`.
- macOS ARM64: Bazel build passed.
- Linux aarch64: manually verified with `cd example/build_with_bazel_module && bazel-7.2.1 $JVMARG build //:echo_c++_server //:echo_c++_client`.

---
### Check List:

### DIFF
--- a/example/build_with_bazel_module/MODULE.bazel
+++ b/example/build_with_bazel_module/MODULE.bazel
@@ -28,3 +28,14 @@ local_path_override(
     module_name = "brpc",
     path = "../..",
 )
+
+single_version_override(
+    module_name = "leveldb",
+    registry = "https://raw.githubusercontent.com/secretflow/bazel-registry/main",
+)
+
+single_version_override(
+    module_name = "openssl",
+    version = "3.3.2.bcr.1",
+    registry = "https://raw.githubusercontent.com/secretflow/bazel-registry/main",
+)
--- a/example/build_with_bazel_module/MODULE.bazel
+++ b/example/build_with_bazel_module/MODULE.bazel
@@ -29,6 +29,16 @@ local_path_override(
     path = "../..",
 )
 
+# Overrides only take effect in the root module. When this example is built,
+# `brpc-example` is the root module and `brpc` is just a dependency, so the
+# same two overrides in brpc's root MODULE.bazel are ignored and have to be
+# repeated here. Keep them in sync with the root MODULE.bazel.
+#
+# leveldb is published only in the secretflow registry — not in BCR, nor in
+# the babylon / apache-brpc registries added by .bazelrc. And brpc declares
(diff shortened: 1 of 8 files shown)


wwbmmm:
LGTM

example/build_with_bazel_module/MODULE.bazel
```diff
@@ -27,4 +27,15 @@ bazel_dep(name = 'brpc', version = '1.17.0', repo_name = 'apache_brpc')
 local_path_override(
     module_name = "brpc",
     path = "../..",
-)
\ No newline at end of file
+)
+
+single_version_override(
```
chenBright, line 42: Why use single_version_override?

zchuango: `single_version_override` is used here to ensure that this example uses the expected dependency versions during Bazel module resolution.

When building this example, `brpc-example` is the root module and `brpc` is only a dependency, so the overrides defined in brpc's own MODULE.bazel do not apply here. Without overriding these dependencies in the root module, Bazel may not be able to resolve the required versions from the available registries.

For example, leveldb 1.23 is only available from the secretflow registry, and openssl 3.3.2 is resolved there as 3.3.2.bcr.1. `single_version_override` allows us to explicitly select these required versions for this example.

I also added comments above these blocks to explain this behavior and noted that they should be kept in sync with the root MODULE.bazel configuration.

chenBright (approved):
LGTM

PR #3470, opened 2026-08-21
Fix potential bvar deadlock by running describe()/dump() outside the global VarMap lock

### LINKED ISSUE #2888: bvar 的全局 VarMap 锁为 pthread mutex, 可能引发死锁
**Describe the bug (描述bug)**
在从 `/brpc_metrics` 采集 metric 的时候, 请求由 bthread 执行, 而此时会为了锁住全局 VarMap 获取一个 pthread mutex, 当有一个用户定义的 metric 中出现了 bthread yield, 则会引起死锁.

**To Reproduce (复现方法)**
使用 `PassiveStatus`, 然后在这个 metric 的计算函数中引入一个 `bthread::Mutex` 的获取, 即会引起死锁. 其他的会产生 bthread yield 的同步原语均可引起同样行为.

**Expected behavior (期望行为)**
个人认为在这里出现 pthread mutex 是反直觉的, 尤其是在文档强调不要令 pthread mutex 和 bthread mutex 混用的情况下. 希望可以:  
 + 将这个锁改为 bthread mutex, 或者
 + 在文档中做补充, 强调不要在自定义 metric 中 yield bthread

**Versions (各种版本)**
OS:
Compiler:
brpc: 
protobuf:

chenBright:

Issue Number: resolve #2888 

Problem Summary:

bvar's global `VarMap` is guarded by a pthread mutex. `Variable::describe_exposed()`
(and `describe_series_exposed()`, `dump_exposed()`, plus the multi-dimension
`MVariableBase` counterparts) used to invoke `var->describe()` while holding that lock. 
For `PassiveStatus`, `describe()` runs a user-provided callback; if the callback yields 
the bthread (e.g. by acquiring a `bthread::Mutex`), the pthread mutex is never released 
and the process deadlocks.

Changed:

Run user callbacks OUTSIDE the global map lock via a small indirection handle:

- New `bvar/detail/exposed_ref.h`: `ExposedRef<T>` (a reference-counted handle
  guarding an exposed object). It uses `butil::Mutex` + `butil::ConditionVariable`.
- `describe_exposed()` / `describe_series_exposed()` / `get_exposed()`: under the
  map lock they now only `seek` + `acquire()` (ref-count +1, serialized with `hide()`'s 
  erase); the lock is released, `describe()` is called outside the lock, then `release()`.
- `hide()` now also invalidates the handle and blocks (`hide_and_wait()`) until
  all in-flight readers finish, so a Variable cannot be destroyed while a concurrent
  `describe()` is still using it. Each `expose()` rebuilds a fresh handle (the old one
  is single-use once hidden).
- `MVariableBase::describe_exposed()` / `dump_exposed()` get the same treatment;
  `dump()` in particular is moved outside the lock because `Dumper` is a
  user-overridable interface that may yield.
- Since callbacks no longer run under the lock, the recursive `VarMap` mutex is no
  longer needed and is reverted to a plain mutex (now consistent with `MVarMap`).

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/bvar/detail/exposed_ref.h
+++ b/src/bvar/detail/exposed_ref.h
@@ -1,0 +1,101 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#ifndef BVAR_DETAIL_EXPOSED_REF_H_
+#define BVAR_DETAIL_EXPOSED_REF_H_
+
+#include <memory>
+#include "butil/macros.h"
+#include "butil/scoped_lock.h"
+#include "butil/synchronization/condition_variable.h"
+
+namespace bvar {
+namespace detail {
+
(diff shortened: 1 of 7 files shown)


wwbmmm:
LGTM

PR #3476, opened 2026-08-22
Bound wire-declared sizes and decompressed output during parsing

wwbmmm:

Problem Summary:

Wire input to bRPC parsers contains lengths/counts that are declared
by the peer. Several parsers used those values to size buffers or
drive allocations before the corresponding bytes were known to exist,
and decompressors never bounded their output. With untrusted clients
this lets a small amount of traffic trigger disproportionate memory
usage.

Changed:

- Decompression (gzip/zlib/snappy) is now bounded: a new
  `-max_decompressed_body_size` flag (default: 32x
  `-max_body_size`, 0 = use the default) caps how much a single body
  may decompress to. Bodies that exceed the cap are rejected with a
  warning log instead of being fully materialized.
- AMF (rtmp) string reading grows the output as bytes arrive in
  64 KB chunks instead of allocating the declared length up front,
  so a truncated message no longer forces the full allocation.
- RTMP chunk parsing now rejects declared message lengths above
  `-max_body_size` and discards a partially assembled body when a
  new message header arrives mid-message.
- The mcpack2pb object iterator validates the wire-declared field
  count against the remaining bytes (each field head is at least 2
  bytes) and generated array-parse code caps its `Reserve()` size.

Side effects:

- Performance effects: negligible; the new bounds only cut
  allocations that were already on the path to failure.
- Breaking backward compatibility: none. Messages within
  `-max_body_size` are unaffected unless they also exceed the new
  decompression cap; operators with legitimately larger compressed
  bodies can raise `-max_decompressed_body_size`.

### Check List:

- Make sure your changes are compilable.
- Tests are added in `test/brpc_snappy_compress_unittest.cpp`,
  `test/brpc_rtmp_unittest.cpp`, `test/brpc_mcpack2pb_unittest.cpp`,
  `test/brpc_sofa_pbrpc_protocol_unittest.cpp`.
- Follow Contributor Covenant Code of Conduct.

### DIFF
--- a/src/brpc/amf.cpp
+++ b/src/brpc/amf.cpp
@@ -16,6 +16,7 @@
 // under the License.
 
 
+#include <algorithm>
 #include <google/protobuf/descriptor.h>
 #include "butil/sys_byteorder.h"
 #include "butil/logging.h"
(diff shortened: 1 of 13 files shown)


chenBright (approved):
LGTM

PR #3482, opened 2026-08-23
Run Bazel tests in parallel in CI

wasphin:

Issue Number: resolve

Problem Summary:

The Bazel unit-test jobs run independent test binaries serially or without an
explicit concurrency limit. This makes the jobs unnecessarily slow, while
unrestricted concurrency exposes tests that rely on timing, fixed ports, or
process-wide networking state to resource contention.

The Bazel disk-cache fallback can also restore an archive produced by a
different compiler or protobuf configuration. Such an archive consumes cache
transfer time and capacity without producing cache hits.

Changed:

- Run the GCC, Clang, and Clang with new protobuf Bazel unit-test jobs with
  `--test_output=errors` and `--local_test_jobs=3`.
- Run timing-sensitive tests and tests that manipulate process-wide networking
  state exclusively, while keeping independent test binaries parallel.
- Use ephemeral ports in the interceptor, HTTP expect, endpoint, and oversized
  message tests to avoid conflicts with other processes.
- Set and restore the oversized-message test's maximum body size explicitly.
- Declare the protobuf JSON test as a large Bazel test so its timeout matches
  its existing runtime under CI contention.
- Restrict Bazel disk-cache fallback restoration to caches produced by the
  same CI job.

Side effects:

- Performance effects: reduces Bazel unit-test wall time while limiting local
  test concurrency to avoid overloading GitHub-hosted runners.
- Breaking backward compatibility: No. Production library behavior is
  unchanged; this updates CI scheduling and test robustness.

---
### Check List:

### DIFF
--- a/.github/actions/setup-build-cache/action.yml
+++ b/.github/actions/setup-build-cache/action.yml
@@ -61,7 +61,6 @@ runs:
         key: ${{ runner.os }}-bazel-disk-${{ github.job }}-${{ github.sha }}
         restore-keys: |
           ${{ runner.os }}-bazel-disk-${{ github.job }}-
-          ${{ runner.os }}-bazel-disk-
 
     - name: Configure Bazel caches
       if: inputs.kind == 'bazel'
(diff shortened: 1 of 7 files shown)


wwbmmm:
LGTM

chenBright (approved):
LGTM

PR #3483, opened 2026-08-24
fix exp-golomb bit position in avc_nalu_read_uev

ubeddulla:
avc_nalu_read_uev adds every Exp-Golomb value bit at the constant shift leadingZeroBits-1 instead of the descending leadingZeroBits-1-i, so any ue(v) code with two or more leading zero bits decodes to the wrong number and a crafted SPS overflows the int32 accumulator past INT32_MAX (UBSan reports the signed overflow). The decoder runs on the SPS inside an untrusted RTMP AVC sequence header, reached through AVCDecoderConfigurationRecord::Create. Shift each bit into its correct position so the result matches the H.264 read_bits(leadingZeroBits) semantics and stays in range.

### DIFF
--- a/src/brpc/details/rtmp_utils.cpp
+++ b/src/brpc/details/rtmp_utils.cpp
@@ -46,7 +46,7 @@ int avc_nalu_read_uev(BitStream* stream, int32_t* v) {
             return -1;
         }
         int32_t b = stream->read_bit();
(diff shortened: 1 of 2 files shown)


wwbmmm:
LGTM

chenBright (approved):
LGTM

PR #3485, opened 2026-08-24
Validate returned stream identifiers

wasphin:

Issue Number: resolve

Problem Summary:

When a client creates multiple streams in one RPC, it expects the server response to return exactly one `extra_stream_id` for every stream after the first. The response path indexed this list using the number of locally created streams without first validating its size. A malformed response could therefore read past the returned identifiers, while surplus identifiers had no valid client-stream mapping.

Changed:

- Require the number of returned `extra_stream_ids` to exactly match the number of additional client streams.
- Fail the RPC with `ERESPONSE` and use the existing stream cleanup path when the counts differ.
- Add full client/server regression coverage for both missing and surplus returned stream identifiers.

Side effects:
- Performance effects: One integer comparison is added when completing a multi-stream RPC.

- Breaking backward compatibility: Malformed responses with missing or surplus stream identifiers are now rejected instead of being partially processed.

---
### Check List:

### DIFF
--- a/src/brpc/controller.cpp
+++ b/src/brpc/controller.cpp
@@ -1711,6 +1711,12 @@ void Controller::HandleStreamConnection(Socket *host_socket) {
             if (!FailedInline()) {
                 SetFailed(EREQUEST, "The server didn't accept the stream");
             }
+        } else if (_remote_stream_settings->extra_stream_ids_size() !=
+                   (int)stream_num - 1) {
+            SetFailed(ERESPONSE, "Server returned %d extra_stream_ids, "
+                      "expected %d",
+                      _remote_stream_settings->extra_stream_ids_size(),
+                      (int)stream_num - 1);
         } else {
             for (size_t i = 0; i < stream_num; ++i) {
                 if (Stream::Address(_request_streams[i], &ptrs[i]) != 0) {
--- a/src/brpc/controller.cpp
+++ b/src/brpc/controller.cpp
@@ -1705,18 +1705,20 @@ void Controller::HandleStreamConnection(Socket *host_socket) {
         return;
     }
     size_t stream_num = _request_streams.size();
+    const size_t expected_extra_streams = stream_num - 1;
     std::vector<StreamUniquePtr> ptrs(stream_num);
     if (!FailedInline()) {
         if (_remote_stream_settings == nullptr) {
             if (!FailedInline()) {
                 SetFailed(EREQUEST, "The server didn't accept the stream");
             }
-        } else if (_remote_stream_settings->extra_stream_ids_size() !=
-                   (int)stream_num - 1) {
+        } else if (static_cast<size_t>(
+                       _remote_stream_settings->extra_stream_ids_size()) !=
+                   expected_extra_streams) {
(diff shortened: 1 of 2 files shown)


src/brpc/controller.cpp
```diff
@@ -1711,6 +1711,12 @@ void Controller::HandleStreamConnection(Socket *host_socket) {
             if (!FailedInline()) {
                 SetFailed(EREQUEST, "The server didn't accept the stream");
             }
+        } else if (_remote_stream_settings->extra_stream_ids_size() !=
+                   (int)stream_num - 1) {
+            SetFailed(ERESPONSE, "Server returned %d extra_stream_ids, "
+                      "expected %d",
+                      _remote_stream_settings->extra_stream_ids_size(),
+                      (int)stream_num - 1);
```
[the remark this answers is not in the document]
wasphin: Addressed in 186ec05b. `stream_num == 0` is already excluded by the early `_request_streams.empty()` return. The expected count is now retained as `size_t`, the protobuf count is converted to the common type for comparison, and the diagnostic uses `%zu`.

test/brpc_streaming_rpc_unittest.cpp
```diff
@@ -1130,6 +1131,74 @@ class MyServiceWithExtraStream : public test::EchoService {
...
+        ASSERT_EQ(0, brpc::StreamCreate(request_streams, STREAM_COUNT, cntl,
+                                        nullptr));
+        ASSERT_EQ(STREAM_COUNT, request_streams.size());
+
+        test::EchoService_Stub stub(&channel);
+        stub.Echo(&cntl, &request, &response, nullptr);
+        ASSERT_TRUE(cntl.Failed());
+        ASSERT_EQ(brpc::ERESPONSE, cntl.ErrorCode());
+        ASSERT_NE(std::string::npos,
+                  cntl.ErrorText().find("extra_stream_ids, expected 2"));
```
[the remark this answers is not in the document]
wasphin: Addressed in 186ec05b. The expected diagnostic substring is now derived from `STREAM_COUNT - 1` instead of hard-coding `2`.

wwbmmm:
LGTM

test/brpc_streaming_rpc_unittest.cpp
```diff
@@ -1130,6 +1131,76 @@ class MyServiceWithExtraStream : public test::EchoService {
...
+        ASSERT_EQ((int)_stream_count + _adjustment,
+                  (int)response_streams.size());
+    }
+
+private:
+    size_t _stream_count;
+    int _adjustment;
+};
+
+TEST_F(StreamingRpcTest, reject_mismatched_returned_stream_identifiers) {
```
chenBright, line 1166: @wasphin This test case sometimes hangs.

https://github.com/apache/brpc/actions/runs/33306389856/job/99243587923?pr=3507

wasphin: Will re-check it

wasphin: Fixed in https://github.com/apache/brpc/pull/3513

PR #3486, opened 2026-08-24
Speed up CI with build caching and parallel jobs

wasphin:

Issue Number: resolve

Problem Summary:
CI recompiles the same sources, downloads the same Bazel dependencies, and
runs the Make and CMake configurations serially on every run, making build
validation unnecessarily slow.

Changed:
- Add a shared composite action that configures ccache for Make/CMake builds.
- Cache Bazel repository downloads separately from Bazel build outputs.
- Reuse caches in Linux and macOS build jobs.
- Report ccache statistics so cache behavior can be verified from CI logs.
- Run the four Linux CMake compiler/configuration combinations as independent
  matrix jobs.
- Run the four Linux Make compiler/configuration combinations as independent
  matrix jobs.
- Run the GCC and Clang Protobuf compatibility builds as independent matrix
  jobs.
- Consolidate each Linux job's APT dependencies into one update/install
  transaction.
- Upgrade checkout to the Node.js 24-based action release.
- Give each ccache job/configuration an unambiguous namespace so restore
  prefixes cannot select another job's cache.
- Disable Bazel test-result caching so every CI run executes all tests while
  still reusing cached compilation actions.

Measured results:
- The latest successful Linux run reduced the slowest job from a 38:25
  baseline average to 21:43 (43.4%). Total runner time across the measured
  Linux jobs fell from 270:09 to 143:45 (46.8%).
- The Make and CMake matrix critical paths were 4:45 and 4:06, compared with
  serial baseline averages of 16:09 and 18:38. Their total runner times were
  15:30 and 14:37, respectively.
- GCC and Clang Make/Protobuf matrix critical paths were 1:59 and 2:11,
  compared with serial baseline averages of 19:32 and 20:39.
- With Bazel test-result caching disabled, the three Bazel unit-test jobs took
  16:36-20:22, reductions of 35.0%-49.8% from their baseline averages. Every
  test still executed; only downloads and compilation actions were reused.
- The latest successful macOS run reduced total runner time from a 28:28
  baseline average to 9:47 (65.6%).

The baseline is the average of 10 successful runs. The post-change values are
from the latest successful run, so they represent the current configuration
rather than a multi-run post-change average.

Side effects:
- Performance effects: Subsequent CI runs can reuse compiler outputs and Bazel
  downloads/build outputs. Ccache is bounded to 2 GiB. Bazel repository caches
  are shared per platform, while build caches remain separated by job. Bazel
  tests always execute; only their compilation actions remain cacheable.

---
### Check List:

### DIFF
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -24,6 +24,7 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
+      max-parallel: 2
       matrix:
         include:
           - name: gcc-default
@@ -68,6 +69,7 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
+      max-parallel: 2
       matrix:
         include:
           - name: gcc-default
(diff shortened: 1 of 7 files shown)


.github/workflows/ci-linux.yml
```diff
@@ -192,10 +233,17 @@ jobs:
         protobuf-install-dir: /protobuf-3.21.12
         config-brpc-options: --cc=clang --cxx=clang++ --werror
 
+    - name: Show ccache statistics
+      if: always()
+      run: ccache --show-stats
+
   clang-unittest-with-bazel:
```
chenBright, line 248: Can the bazel test be performed without caching?

wasphin: Bazel caches test results as well, but I think reusing them is acceptable when the test and all of its relevant inputs are unchanged. Bazel will invalidate the cached result when those inputs change.

If we want CI to always execute tests, we can disable test-result caching separately, e.g. with --cache_test_results=no, while still keeping the build cache.

chenBright: I think it's best to always execute tests because some problems are sporadic.

wasphin: Updated. CI now sets `test --cache_test_results=no`, so Bazel reuses compilation actions but executes every test on each run.

chenBright:
@wwbmmm once mentioned that Apache recommends not using too many CIs.

wasphin:
1. The overall runner time is down 46.8% (from 270 min to ~144 min per workflow), so in terms of total resource consumption, this is a clear win.
2. I'm aware of the ASF concurrency limit (≤ 20 [1]). The current matrix brings it to 25, but the matrix also helps simplify configuration and centralize maintenance across different triggers. I'll look into other ways to limit the concurrency while keeping the matrix structure intact.

[1] https://infra.apache.org/github-actions-policy.html

chenBright:
The unittest ci takes 20 minutes to execute, so merging other build ci (e.g., protobuf build ci) into one does not affect the overall execution time.

wwbmmm:
> The unittest ci takes 20 minutes to execute, so merging other build ci (e.g., protobuf build ci) into one does not affect the overall execution time.

I think we can just remove some param combinations to keep the job num <= 20. Merging jobs into one may increase failure rate.

BTW, I think the unit test takes 20 minutes is not ideal. Maybe we can optimize those slow cases in the future.

wasphin:
@wwbmmm @chenBright Agreed. Reducing the unit-test execution time is the next step after this PR. I have identified serial test-binary execution and several fixed-duration performance/stress cases as the main bottlenecks, and I will address them separately to keep this PR focused on build caching and compilation-job parallelism.

wasphin:
Updated in ca9e3d65 to comply with the [ASF GitHub Actions policy](https://infra.apache.org/github-actions-policy.html). The Make and CMake matrices now each use `max-parallel: 2`, and the six GCC/Clang Protobuf configurations share one matrix with `max-parallel: 2`. This preserves all existing build coverage while limiting the maximum combined concurrency to 17 jobs: 13 Linux, 3 macOS, and 1 license job.

chenBright (approved):
LGTM

PR #3473, opened 2026-08-22
Fix unstable hpack UT

wwbmmm:
### What problem does this PR solve?

Issue Number: resolve 

Problem Summary: Fix unstable hpack UT brought in https://github.com/apache/brpc/pull/3343

### What is changed and the side effects?

Changed: remove fork() call in UT

Side effects:
- Performance effects: No

- Breaking backward compatibility: No

---
### Check List:
- Please make sure your changes are compilable.
- When providing us with a new feature, it is best to add related tests.
- Please follow [Contributor Covenant Code of Conduct](https://github.com/apache/brpc/blob/master/CODE_OF_CONDUCT.md).

PR #3343, opened 2026-06-14
Fix HPACK dynamic table size update recursive decoding

wwbmmm:
### What problem does this PR solve?

Issue Number: null

Problem Summary:

HPACK decoding handled Dynamic Table Size Update entries by recursively calling
the decoder again. A header block containing many consecutive table size updates
could therefore use a large amount of call stack before reaching a real header
field or the end of input.

### What is changed and the side effects?

Changed:
- Decode consecutive Dynamic Table Size Update entries iteratively.
- Preserve the total consumed byte count when table size updates precede a
  decoded header field.
- Add HPACK regression tests for long update sequences and update-before-header
  decoding.

Side effects:
- Performance effects: avoids recursive call overhead for consecutive table
  size updates.

- Breaking backward compatibility: none.

---
### Check List:
- Please make sure your changes are compilable.
- When providing us with a new feature, it is best to add related tests.
- Please follow [Contributor Covenant Code of Conduct](https://github.com/apache/brpc/blob/master/CODE_OF_CONDUCT.md).

PR #3488, opened 2026-08-25
Remove noisy INFO log in ResetMaxSize and fix hpack UT instability

wwbmmm:

Related to https://github.com/apache/brpc/pull/3473

Problem Summary:

The `HPackTest.many_dynamic_table_size_updates` test (introduced in
PR #3343) was unstable in clang-unittest-asan CI, frequently hanging.

Root cause:

**200000 LOG(INFO) lines in a single test** — `ResetMaxSize()` printed
   a `LOG(INFO)` on every call. The test feeds 200000 consecutive Dynamic
   Table Size Update entries, each triggering an INFO log, flooding the
   output with 200000+ lines. Under ASAN this log storm was slow enough to
   timeout CI jobs.

Changed:
- Remove the `LOG(INFO)` from `IndexTable::ResetMaxSize()`. This was a
  debug-only log that printed on every HTTP/2 table size update — it had
  no place in production code and was the direct cause of the log storm.

Side effects:
- Performance effects: Removing the `LOG(INFO)` from `ResetMaxSize()` is a
  minor improvement for all HTTP/2 decoding paths that handle table size
  updates.

### How to test?

- The test case `HPackTest.many_dynamic_table_size_updates` now runs under
  ASAN in ~1ms with 12 lines of log output (previously ~527ms with
  200012 lines).

### Check List:

### DIFF
--- a/src/brpc/details/hpack.cpp
+++ b/src/brpc/details/hpack.cpp
@@ -177,8 +177,6 @@ public:
     }
 
     void ResetMaxSize(size_t new_max_size) {
-        LOG(INFO) << this << ".size=" << _size << " new_max_size=" << new_max_size
-                  << " max_size=" << _max_size;
         if (new_max_size > _max_size) {
             //LOG(ERROR) << "Invalid new_max_size=" << new_max_size;
             //return -1;


chenBright (approved):
LGTM

PR #3489, opened 2026-08-25
Limit reassembled stream message size

wasphin:

Issue Number: N/A

Problem Summary:

Each stream data frame is bounded by `max_body_size`, but continuation frames
can be accumulated into a message larger than that limit.

Changed:

- Limit the accumulated payload of a fragmented stream message to
  `max_body_size`.
- Release pending fragments and close the stream with `EMSGSIZE` when the
  accumulated payload exceeds the limit.
- Cover the boundary and rejection behavior through a complete client/server
  RPC stream.

Side effects:

- Performance effects: one constant-time size check per stream data frame.
- Breaking backward compatibility: fragmented stream messages larger than
  `max_body_size` are now rejected instead of being accumulated without a
  bound.

---
### Check List:

### DIFF
--- a/src/brpc/stream.cpp
+++ b/src/brpc/stream.cpp
@@ -35,6 +35,7 @@
 namespace brpc {
 
 DECLARE_bool(usercode_in_pthread);
+DECLARE_uint64(max_body_size);
 DECLARE_int64(socket_max_streams_unconsumed_bytes);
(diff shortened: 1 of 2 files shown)


wwbmmm:
LGTM

chenBright (approved):
LGTM

PR #3490, opened 2026-08-26
refactor(bthread): make TaskControl metric ownership explicit

darion-yaphet:

  Problem Summary:

  TaskControl stored per-tag bvar metrics as raw pointers. In addition, each cumulative-time PassiveStatus received a heap-allocated callback argument with no explicit  owner. Since PassiveStatus does not own that argument, its lifetime was unclear and could leak when TaskControl is destroyed.

  Changed:

  - Replace the four per-tag bvar raw-pointer vectors with std::unique_ptr containers.
  - Add explicit ownership for each CumulatedWithTagArgs callback argument.
  - Preserve metric names, access patterns, and scheduler behavior.
  - Keep callback arguments alive until their corresponding PassiveStatus is destroyed.

  Side effects:

  - Performance effects: No steady-state impact expected. Ownership changes occur only during TaskControl initialization/destruction.
  - Breaking backward compatibility: None; no public API or metric name changes.

  ———

  ### Check List:

  - [x] The modified task_control.cpp CMake object target compiles successfully.
  - [x] git diff --check passes.
  - [ ] Full bthread test suite was not run because the current build is blocked by an unrelated brpc::EPROGREADTIMEOUT compilation error.
  - [ ] No new feature; no additional behavior test required.

### DIFF
--- a/src/bthread/task_control.h
+++ b/src/bthread/task_control.h
@@ -41,6 +41,7 @@ DECLARE_int32(task_group_ntags);
 namespace bthread {
 
 class TaskGroup;
+struct CumulatedWithTagArgs;
 
 // Control all task groups
 class TaskControl {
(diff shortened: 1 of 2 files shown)


wwbmmm:
LGTM

src/bthread/task_control.h
```diff
@@ -171,10 +172,14 @@ friend bthread_t init_for_pthread_stack_trace();
     bvar::PassiveStatus<std::string> _status;
     bvar::Adder<int64_t> _nbthreads;
 
-    std::vector<bvar::Adder<int64_t>*> _tagged_nworkers;
-    std::vector<bvar::PassiveStatus<double>*> _tagged_cumulated_worker_time;
-    std::vector<bvar::PerSecond<bvar::PassiveStatus<double>>*> _tagged_worker_usage_second;
-    std::vector<bvar::Adder<int64_t>*> _tagged_nbthreads;
+    std::vector<std::unique_ptr<bvar::Adder<int64_t>>> _tagged_nworkers;
```
chenBright, line 175: Wouldn't it be better to use `std::vector<bvar::Adder<int64_t>>`?

darion-yaphet: > Wouldn't it be better to use `std::vector<bvar::Adder<int64_t>>`?

bvar::Adder inherits bvar::Variable, which is non-copyable, and Adder is also effectively non-movable. A std::vector<bvar::Adder<int64_t>> must be able to relocate elements when it grows, so it cannot safely support this pattern. More importantly, each exposed bvar must retain a stable address.

chenBright (approved):
LGTM

PR #3493, opened 2026-08-26
Limit streams accepted per request

wasphin:

Issue Number: N/A

Problem Summary:

`StreamAccept` creates one response stream for every stream identifier declared
by a request, without limiting the number created by a single RPC.

Changed:

- Add `stream_max_streams_per_request` with a default value of 64.
- Reject requests exceeding the limit before creating any response streams.
- Cover the boundary, rejection, and a subsequent valid request over the same
  client/server connection.

Side effects:

- Performance effects: one constant-time count check per `StreamAccept` call.
- Breaking backward compatibility: applications accepting more than 64
  streams in one RPC must raise `stream_max_streams_per_request`.

---
### Check List:

### DIFF
--- a/test/brpc_streaming_rpc_unittest.cpp
+++ b/test/brpc_streaming_rpc_unittest.cpp
@@ -1260,12 +1260,15 @@ public:
         response->set_message(request->message());
 
         brpc::StreamIds response_streams;
-        accept_result = brpc::StreamAccept(response_streams, *cntl, nullptr);
-        accepted_streams = response_streams.size();
+        accept_result.store(
+            brpc::StreamAccept(response_streams, *cntl, nullptr),
+            std::memory_order_release);
+        accepted_streams.store(
+            response_streams.size(), std::memory_order_release);
     }
 
-    int accept_result{0};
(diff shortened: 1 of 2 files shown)


test/brpc_streaming_rpc_unittest.cpp
```diff
@@ -1249,6 +1249,83 @@ class MyServiceWithMismatchedExtraStreamIds : public test::EchoService {
...
+            std::memory_order_release);
+        accepted_streams.store(
+            response_streams.size(), std::memory_order_release);
+    }
+
+    std::atomic<int> accept_result{0};
+    std::atomic<size_t> accepted_streams{0};
+};
+
+TEST_F(StreamingRpcTest, limit_streams_accepted_per_request) {
```
chenBright, line 1274: @wasphin This test case sometimes hangs.

https://github.com/apache/brpc/actions/runs/34106462745/job/101692505463?pr=3525

wasphin: Fixed in https://github.com/apache/brpc/pull/3531

PR #3494, opened 2026-08-26
Reject invalid stream feedback frames

wasphin:

Issue Number: resolve

Problem Summary:

Malformed stream feedback frames can reach `CHECK` statements when they carry
an unexpected payload or when flow control is disabled.

Changed:

- Reject malformed feedback with `EPROTO` and close only the affected stream.
- Use scoped cleanup while updating remote-consumption state.
- Cover both invalid feedback cases through a client/server RPC connection.

Side effects:
- Performance effects: No expected impact on valid feedback frames.

---
### Check List:


chenBright (approved):
LGTM

PR #3497, opened 2026-08-27
check chunk header bounds in ubring recv paths

ubeddulla:

Issue Number: resolve N/A

Problem Summary:

UbrTrxRecvBlockMode and StartReadv read a chunk's msg_len and cur_index straight from the receive-ring header that the remote peer writes into shared memory, then memcpy msg_len minus cur_index bytes out of the fixed 60-byte payload.inner without checking either value. A peer that sets msg_len above UBR_MSG_PAYLOAD_LEN (60), or cur_index above msg_len so the uint8_t subtraction wraps, makes that copy read past the payload into the adjacent chunk header and neighbouring ring slots. Under ASAN this shows up as a heap-buffer-overflow read of size 255 at 0 bytes after the 64-byte chunk. The send path already caps msg_len at UBR_MSG_PAYLOAD_LEN, so only the receive side was missing the check.

Changed:

Added IsRecvChunkHeaderValid (msg_len <= UBR_MSG_PAYLOAD_LEN and cur_index <= msg_len) and apply it at both receive sites before the copy, rejecting a malformed chunk with errno=EBADMSG and the same UBRING_ERR the existing pre-checks return. Added a regression test in brpc_ubring_unittest.cpp.

Side effects:
- Performance effects: one comparison per chunk, negligible.

- Breaking backward compatibility: none. Valid chunks (msg_len <= 60, cur_index <= msg_len) behave exactly as before.

---
### Check List:

### DIFF
--- a/src/brpc/ubshm/ub_ring.cpp
+++ b/src/brpc/ubshm/ub_ring.cpp
@@ -424,6 +424,12 @@ int UBRing::UbrTrxRecvBlockMode(uint8_t *dest, uint32_t buf_len)
         }
         uint8_t chunk_msg_len = current_chunk->header[UBR_MSG_LEN_INDEX];
         uint8_t cur_index = current_chunk->header[UBR_MSG_CUR_INDEX];
+        if (UNLIKELY(!IsRecvChunkHeaderValid(chunk_msg_len, cur_index))) {
+            LOG(ERROR) << "Trx recv failed, invalid chunk header msg_len="
+                       << (uint32_t)chunk_msg_len << " cur_index=" << (uint32_t)cur_index;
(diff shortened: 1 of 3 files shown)


wwbmmm:
@zchuango Have a look if you have time

zchuango:
> @zchuango Have a look if you have time

OK

zchuango:
@ubeddulla Thanks for fixing this! The bounds checks in both `UbrTrxRecvBlockMode` and `StartReadv` close the malformed chunk-header over-read issue. Appreciate the clear regression tests and the contribution.

zchuango (approved):

PR #3499, opened 2026-08-28
Limit recursion depth of mcpack2pb parser

wwbmmm:

Problem Summary:

The mcpack2pb parser (used by nshead_mcpack / ubrpc / public_pbrpc
requests) has no limit on the nesting depth of the input object/array.
For a protobuf message that references itself (e.g. a `repeated`
message field), the generated parser recurses once per nesting level,
so an input nested deep enough makes the recursion grow unbounded and
overflow the stack, crashing the process. The serializer has enforced
MAX_DEPTH (`src/mcpack2pb/field_type.h`) since the beginning, the
parser just never did.

Changed:

- Thread the current nesting depth through `mcpack2pb::UnparsedValue`
  and the `ObjectIterator` / `ArrayIterator` constructors.
- Fail the parse (mark the input stream bad) as soon as the depth
  exceeds `MAX_DEPTH` (128), exactly like the serializer already does,
  so that the recursion stops at a bounded level instead of exhausting
  the stack.
- Add unit tests covering both a deeply-nested payload (rejected
  without crashing) and a moderately-nested one (parsed normally).

The depth limit is enforced on the iterator level, so existing code
generated by `protoc-gen-mcpack` is protected without regeneration:
no generated-code API is changed.

Side effects:
- Parsing now fails for mcpack input nested deeper than 128 levels
  (the same limit the serializer already applies). Such input is
  abnormal for any real workload, roughly matching
  `json2pb_max_recursion_depth` (100) used by the JSON parser.
- Performance effects: none (one integer compare per iterator).

- Breaking backward compatibility: no.

### Check List:


chenBright (approved):
LGTM

PR #3500, opened 2026-08-28
Contain malformed Thrift T_EXCEPTION replies on the client

wwbmmm:
## What problem does this PR solve?

Problem Summary:

A bRPC Thrift client can be brought down by a peer reply that carries a
malformed `T_EXCEPTION`. The exception struct is decoded in
`ReadThriftException` without a try/catch; a bad field or an oversized
length makes the underlying thrift library throw (`TProtocolException`,
`TTransportException` or `std::length_error`). The exception is not caught
anywhere on the return path and eventually reaches the bthread task frame,
where `std::terminate()` is called, terminating the whole process.

This only affects builds with `WITH_THRIFT` enabled (off by default).

## What is changed and the side effects?

Changed:

`ReadThriftException` now wraps the decode of the exception struct in the
same `try/catch` that `ReadThriftStruct` already uses for the normal reply
body. A malformed `T_EXCEPTION` reply is reported as a failed RPC instead
of crashing the process.

Side effects:

- None. The fix only contains an otherwise-uncaught exception.

- Performance effects: none.

## Check List

- [x] The change is compilable.
- [x] A unit test is added in `test/brpc_thrift_protocol_unittest.cpp`
      that feeds a malformed `T_EXCEPTION` reply through the client
      response path and asserts the call does not crash and marks the
      controller failed.
- [x] Follows the Contributor Covenant Code of Conduct.

### DIFF
--- a/test/BUILD.bazel
+++ b/test/BUILD.bazel
@@ -24,12 +24,7 @@ COPTS = [
     "-Wno-unused-parameter",
     "-fno-omit-frame-pointer",
     "-fno-access-control",
-] + select({
-    # The thrift framed protocol (and the unit test that exercises it) is only
-    # compiled when the build has Thrift support enabled, matching src/BUILD.
-    "//bazel/config:brpc_with_thrift": ["-DENABLE_THRIFT_FRAMED_PROTOCOL=1"],
-    "//conditions:default": [],
-})
+]
 
 TEST_BUTIL_SOURCES = [
     "at_exit_unittest.cc",
--- a/test/BUILD.bazel
+++ b/test/BUILD.bazel
@@ -24,7 +24,12 @@ COPTS = [
     "-Wno-unused-parameter",
     "-fno-omit-frame-pointer",
     "-fno-access-control",
-]
+] + select({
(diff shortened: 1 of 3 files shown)


test/BUILD.bazel
```diff
@@ -24,7 +24,12 @@ COPTS = [
     "-Wno-unused-parameter",
     "-fno-omit-frame-pointer",
     "-fno-access-control",
-]
+] + select({
+    # The thrift framed protocol (and the unit test that exercises it) is only
+    # compiled when the build has Thrift support enabled, matching src/BUILD.
+    "//bazel/config:brpc_with_thrift": ["-DENABLE_THRIFT_FRAMED_PROTOCOL=1"],
+    "//conditions:default": [],
+})
```
chenBright: It seems this select is not needed. unit test targets will inherit this definition of bRPC source targets automatically.

https://github.com/apache/brpc/blob/4a429cdc2153eff616902edfd2349823d6d7c5b5/BUILD.bazel#L45-L48

wwbmmm: removed

chenBright (approved):
LGTM

PR #3501, opened 2026-08-28
Release the RTMP peer key after key agreement

wasphin:

Issue Number: resolve

Problem Summary:

`DHWrapper::copy_shared_key` only released the temporary peer public key
when `DH_compute_key` failed. Successful RTMP complex handshakes retained the
allocation.

Changed:

Release the temporary peer public key immediately after `DH_compute_key`
returns, covering both success and failure paths.

Side effects:
- Performance effects: Negligible; this only releases an allocation when its
  last use completes.

---
### Check List:


chenBright (approved):
LGTM

PR #3502, opened 2026-08-29
Add bounded-load consistent hashing load balancer (c_murmurhash_bl)

rajvarun77:
# What & Why

Classic consistent hashing has no relief valve for hot keys: one server saturates while its ring neighbors sit idle. This PR adds `c_murmurhash_bl`, an implementation of ["Consistent Hashing with Bounded Loads"](https://arxiv.org/abs/1608.01350) (Mirrokni et al., CACM 2017) on the existing murmurhash ring, so hot-key overflow spills to deterministic ring successors instead of overloading a single server.

# Usage

```
options.load_balancer_type = "c_murmurhash_bl";              // default factor from -chash_bounded_load_factor (1.25, must be > 1)
options.load_balancer_type = "c_murmurhash_bl:load_factor=1.5";  // per-channel override; replicas= also supported
```

# Design

- Subclass of `ConsistentHashingLoadBalancer`; only `SelectServer` and load accounting differ, ring construction is shared.
- Per-server capacity = `ceil(load_factor * (total_inflight + 1) / server_count)`; an at-capacity server overflows clockwise to the next server with spare capacity, with first-seen fallback if all are at capacity.
- In-flight counters are relaxed atomics shared across both DoublyBufferedData buffers, incremented at selection, decremented exactly once in `Feedback()` (total stays consistent even if the server was removed in between), and resynced on membership change.

# Tests & Docs

7 new cases in `test/brpc_ch_bounded_load_balancer_unittest.cpp` (factor/replicas validation, hot-key capping, overflow to ring successor, feedback decrement, removal consistency, no accounting without changable_weights); existing consistent-hashing tests unchanged and passing. Documented in `docs/cn/client.md` and `docs/en/client.md`.

cc @chenBright (reviewer of the sibling p2c load balancer, #3367)

### DIFF
--- a/docs/cn/client.md
+++ b/docs/cn/client.md
@@ -277,9 +277,15 @@ locality-aware，优先选择延时低的下游，直到其延时高于其他机
 
 注意甄别请求中的“主键”部分和“属性”部分，不要为了偷懒或通用，就把请求的所有内容一股脑儿计算出哈希值，属性的变化会使请求的目的地发生剧烈的变化。另外也要注意padding问题，比如struct Foo { int32_t a; int64_t b; }在64位机器上a和b之间有4个字节的空隙，内容未定义，如果像hash(&foo, sizeof(foo))这样计算哈希值，结果就是未定义的，得把内容紧密排列或序列化后再算。
 
+每台服务器的虚拟节点数默认由-chash_num_replicas控制（默认100），可按channel覆盖：`c_murmurhash:replicas=300`。
+
 实现原理请查看[Consistent Hashing](consistent_hashing.md)。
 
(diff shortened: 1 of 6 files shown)


rajvarun77:
@chenBright @wwbmmm gentle ping when you have a moment — CI has been fully green for a week and the Copilot pass is addressed. This is the bounded-load follow-up to the consistent-hashing family (same ring, only `SelectServer` + in-flight accounting differ), so it should be a fairly contained review. Happy to adjust the default `load_factor` or the option-string parsing if you'd prefer a different shape.

wwbmmm:
LGTM

chenBright (approved):
LGTM

PR #3504, opened 2026-08-29
fix(bthread): drop unimplemented barrier and rwlockattr APIs

darion-yaphet:

Problem Summary:
`bthread.h` declared POSIX-style barrier APIs (`bthread_barrier_init/destroy/wait`) and rwlock attribute APIs (`bthread_rwlockattr_init/destroy/getkind_np/setkind_np`) that were never defined in `libbrpc`. Compiling against the headers succeeded, then linking failed with `undefined reference`. The headers advertised an API that the library does not provide.

Changed:
- Remove unimplemented `bthread_barrier_*` declarations and the unused `bthread_barrier_t` / `bthread_barrierattr_t` types.
- Remove unimplemented `bthread_rwlockattr_*` declarations.
- Keep `bthread_rwlockattr_t` because it is still the argument type of the implemented `bthread_rwlock_init()`.
- Document that `bthread_rwlock_init()` currently ignores `attr`.

Side effects:
- Performance effects: none.

- Breaking backward compatibility: source-incompatible only for code that called these never-linked symbols. Working programs that already linked against brpc could not have been using them. `bthread_rwlock_init()` signature is unchanged.

---
### Check List:

### DIFF
--- a/src/bthread/bthread.h
+++ b/src/bthread/bthread.h
@@ -249,6 +249,7 @@ extern int bthread_cond_timedwait(
 
 // Initialize read-write lock `rwlock' using attributes `attr', or use
 // the default values if later is nullptr.
+// NOTE: attr is not used in the current implementation.
 extern int bthread_rwlock_init(bthread_rwlock_t* __restrict rwlock,
                                const bthread_rwlockattr_t* __restrict attr);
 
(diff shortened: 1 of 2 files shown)


wwbmmm:
LGTM

chenBright (commented):
Commenting it out and adding a todo tag instead of deleting it is more appropriate.

darion-yaphet:
> Commenting it out and adding a todo tag instead of deleting it is more appropriate.

Because brpc is an RPC system rather than a thread library, I think since this API has not been implemented, it should not be exposed. What do you think about this?

chenBright:
I think keeping it in the comments is fine. We might implement these APIs later if needed.

```c
// ---------------------------------------------------
// Functions for handling read-write lock attributes.
// ---------------------------------------------------

// Initialize attribute object `attr' with default values.
// extern int bthread_rwlockattr_init(bthread_rwlockattr_t* attr);

// Destroy attribute object `attr'.
// extern int bthread_rwlockattr_destroy(bthread_rwlockattr_t* attr);

// Return current setting of reader/writer preference.
// extern int bthread_rwlockattr_getkind_np(const bthread_rwlockattr_t* attr,
                                         int* pref);

// Set reader/write preference.
// extern int bthread_rwlockattr_setkind_np(bthread_rwlockattr_t* attr, int pref);
```

chenBright (approved):
LGTM

PR #3505, opened 2026-08-29
Fix data race on  Socket::_read_buf of RDMA server 

### LINKED ISSUE #3479: RDMA server: race condition on _read_buf between PollCq and OnNewMessages causes core dump
**Describe the bug**
When a brpc server is started with `SOCKET_MODE_RDMA`, the server-side RDMA socket's _read_buf can be concurrently accessed by two independent bthreads, causing IOBuf internal state corruption and intermittent core dumps.

There are 4 race scenarios, all stemming from the same root cause: `PollCq`(CQ socket bthread) and `OnNewMessages`(main socket bthread) can run concurrently and both operate on the same Socket::_read_buf, which is not thread-safe.

**Scenario 1: After RDMA handshake is established**

RdmaTransport::Init sets the server-side edge trigger to `InputMessenger::OnNewMessages` (to drive the handshake via the standard InputMessenger path). After the handshake reaches ESTABLISHED, the edge trigger is never switched. If the TCP fd becomes readable (e.g., TCP keepalive, peer close), OnNewMessages is called and reads from the TCP fd into _read_buf. At the same time, PollCq processes RDMA completions and writes RDMA data into _read_buf via HandleCompletion. Both bthreads modify _read_buf concurrently → IOBuf corruption → core dump in IOBuf::cutn or IOBuf::clear.
```
 Bthread A: PollCq (CQ socket)              Bthread B: OnNewMessages (main socket)
 ┌─────────────────────────────┐          ┌──────────────────────────────────┐
 │ ibv_poll_cq → got completion │          │ TCP fd readable (keepalive etc) │
 │ HandleCompletion:            │          │ DoRead:                          │
 │   _rbuf.cutn(                │  RACE!   │   _read_buf.append_from_fd(...)   │
 │     &_socket->_read_buf)  ★──┼──────────┼──>  ★ writes _read_buf           │
 │   ★ writes _read_buf         │          │ ProcessNewMessage:               │
 │                              │          │   cutn → reads _read_buf  ★       │
 │                              │          │                                  │
 │ IOBuf internal state corrupted → core dump in cutn() or clear()           │
 └─────────────────────────────┘          └──────────────────────────────────┘
```

**Scenario 2: During RDMA handshake (after QP is brought up)**

In ExecuteServerHandshake Phase 1, `BringUpQp`puts the QP into RTS state. The client may complete its handshake and start sending RDMA data before the server's handshake finishes. PollCq receives the RDMA completion and HandleCompletion writes to _read_buf. Meanwhile, the server's OnNewMessages is still driving the handshake (Phase 2: draining the 4-byte ACK) and accessing _read_buf via source->cutn(). Concurrent access → core dump in IOBuf::cutn.
```
 Client                              Server
 ─────                               ──────
  │                                    │
  │── RDMA hello ──TCP──────────────→  │ Phase 1: parse hello, BringUpQp (QP→RTS)
  │←── server hello ──TCP────────────  │         SendLocalHello, state=S_ACK_WAIT
  │                                    │
  │── ACK (4B) ──TCP────────────────→  │ Phase 2: OnNewMessages reads ACK...
  │                                    │
  │══ RDMA data ════════════════════→  │ PollCq: HandleCompletion writes _read_buf ★
  │                                    │ OnNewMessages: source->cutn(&flags, 4)  ★
  │                                    │ → CONCURRENT _read_buf access → CRASH
```

**Scenario 3: Handshake ACK rejection**

ExecuteServerHandshake Phase 2 checks source->size() > HELLO_ACK_LEN and rejects the connection. When a client falls back to TCP, it sends the 4-byte ACK followed immediately by the first baidu_std RPC request. If both arrive before the server's readv(), they are read into _read_buf in one call. The server sees source->size() > 4 and drops the connection. This causes mass handshake failures for all TCP-fallback clients.
```
 Client                              Server
 ─────                               ──────
  │── ACK (4B) ──┐                   │
  │── RPC (NB) ──┤                   │
  │              ↓ (same readv)      │ source->size() = 4 + N > 4
  │══════════════════════════════→  │ "Too many bytes in handshake ACK"
  │                                │ → PARSE_ERROR_ABSOLUTELY_WRONG → drop!
  │                                │ (should drain 4B ACK, let N bytes
  │                                │  be parsed by baidu_std parser)
```

**Scenario 4: PollCq calling ProcessNewMessage with bytes == 0 during handshake**
PollCq unconditionally calls ProcessNewMessage after processing completions, even when bytes == 0 (e.g., IBV_WC_SEND completions, or IBV_WC_RECV dropped during handshake). ProcessNewMessage calls CutInputMessage, which invokes the handshake parser (ExecuteServerHandshake) and calls cutn() on _read_buf. Meanwhile, OnNewMessages is also calling ProcessNewMessage → CutInputMessage → ExecuteServerHandshake → cutn() on the same _read_buf. Both bthreads call cutn() on the same IOBuf concurrently → corruption → core dump.
```
 Bthread A: PollCq (CQ socket)              Bthread B: OnNewMessages (main socket)
 ┌─────────────────────────────┐           ┌──────────────────────────────────┐
 │ ibv_poll_cq → IBV_WC_SEND    │           │ DoRead → reads ACK to _read_buf │
 │ HandleCompletion → return 0  │           │ ProcessNewMessage:              │
 │ bytes = 0                    │           │   CutInputMessage:              │
 │ ProcessNewMessage(bytes=0):  │  RACE!    │     ExecuteServerHandshake:      │
 │   CutInputMessage:           │           │       source->cutn(&flags, 4) ★ │
 │     ExecuteServerHandshake:  │           │                                  │
 │       source->cutn(&flags,4)★┼───────────┼──>                               │
 │                              │           │                                  │
 │ Both bthreads cutn() the same _read_buf → IOBuf corruption → CRASH        │
 └─────────────────────────────┘           └──────────────────────────────────┘
```
Stack trace from production:
```
#0  butil::IOBuf::cutn
#1  brpc::rdma::RdmaEndpoint::ExecuteServerHandshake
#2  brpc::rdma::ExecuteServerHandshake
#3  brpc::policy::ParseRdmaHandshake
#4  brpc::InputMessenger::CutInputMessage
#5  brpc::InputMessenger::ProcessNewMessage
#6  brpc::rdma::RdmaEndpoint::PollCq          ← PollCq, not OnNewMessages
#7  brpc::Transport::OnEdge
#8  bthread::TaskGroup::task_runner
```

**To Reproduce**
1. Start a brpc server with ServerOptions.socket_mode = `SOCKET_MODE_RDMA`
2. Have RDMA clients connect; some will negotiate RDMA successfully, others may fall back to TCP
3. After RDMA is established, if the TCP fd becomes readable (keepalive, peer close, etc.), the race triggers
4. Core dump occurs intermittently (~20% probability at startup, sporadically during runtime)
5. Stack trace shows crash in `IOBuf::clear()` or `IOBuf::cutn()` during `Socket::BeforeRecycled()`, `ExecuteServerHandshake`, or `PollCq`

**Expected behavior**
No core dump. _read_buf should not be accessed concurrently by PollCq and OnNewMessages. Handshake should not fail when ACK and RPC data arrive in the same readv() call.

**Versions**
OS: Rocky 9.2
Compiler: GCC 13.3
brpc: master 2026.08.23
protobuf: -

The client-side RDMA socket does not have this issue because its edge trigger is RdmaEndpoint::OnNewDataFromTcp, which checks the RDMA state and only reads 1 byte for EOF detection in ESTABLISHED state (without touching _read_buf).

chenBright:

Issue Number: resolve #3479

Problem Summary:

Changed:

1. One input stream, one buffer. New `InputMessengerProcessor` holds the 
state of a single input stream: its `butil::IOPortal` and the message-size statistics 
that size the next read. `Socket::_read_buf`, `_last_msg_size` and `_avg_msg_size` 
move into it, and `InputMessenger::CutInputMessage()` / `ProcessNewMessage()` 
become its methods. A `Socket` owns one for its fd, a `RdmaEndpoint` owns one for 
its QP. `Socket::DoRead()` now takes the destination `IOPortal*` instead of always
filling `_read_buf`.

2. The server stops parsing the TCP fd once RDMA is on (scenario 1).
`RdmaTransport::Init()` installs `RdmaEndpoint::OnNewDataFromTcp` for both sides
rather than only for the client; it dispatches on `Socket::CreatedByConnect()`.
In `ESTABLISHED` the fd is only probed for EOF, never parsed. This is needed on
top of the buffer split because `preferred_index` and `parsing_context` are
still per-Socket, so two streams must not parse at once.

3. CQ events start after the handshake, not during it (scenarios 2).
`RdmaEndpoint::StartCqEvents()` is split out of `DoAllocateResources()`. The
server calls it from `OnNewDataFromTcpAtServer()` once `OnNewMessages()` has
returned, the client from `ProcessHandshakeAtClient()`. No CQE is lost by
deferring: `BringUpQp()` fills the RQ and `DoAllocateResources()` arms both CQs
before the QP reaches RTS, and adding an already readable fd to an edge-triggered 
epoll reports it immediately.

4. The handshake ACK no longer swallows what follows it (scenario 3). Phase
2 consumes exactly `HELLO_ACK_LEN` bytes and leaves the rest to the real
protocol when the connection falls back to TCP. When RDMA is on the fd is not an
RPC channel any more, so bytes on it stay a protocol error, and a guard at the
top of `ExecuteServerHandshake()` turns them away once the endpoint has left the
handshake.

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/brpc/input_message_base.h
+++ b/src/brpc/input_message_base.h
@@ -53,6 +53,7 @@ protected:
 
 private:
 friend class InputMessenger;
+friend class InputMessengerProcessor;
 friend void* ProcessInputMessage(void*);
 friend class Stream;
 friend class Transport;
--- a/src/brpc/input_messenger.cpp
+++ b/src/brpc/input_messenger.cpp
@@ -74,98 +74,6 @@ DEFINE_int32(socket_tcp_user_timeout_ms, -1,
 
 DECLARE_bool(usercode_in_pthread);
 DECLARE_bool(usercode_in_coroutine);
-DECLARE_uint64(max_body_size);
-
-const size_t MSG_SIZE_WINDOW = 10;  // Take last so many message into stat.
-const size_t MIN_ONCE_READ = 4096;
-const size_t MAX_ONCE_READ = 524288;
-
-ParseResult InputMessenger::CutInputMessage(
-        Socket* m, size_t* index, bool read_eof) {
-    const int preferred = m->preferred_index();
-    const int max_index = (int)_max_index.load(butil::memory_order_acquire);
-    // Try preferred handler first. The preferred_index is set on last
-    // selection or by client.
-    if (preferred >= 0 && preferred <= max_index
-            && _handlers[preferred].parse != nullptr) {
-        int cur_index = preferred;
-        do {
-            ParseResult result =
-                _handlers[cur_index].parse(&m->_read_buf, m, read_eof, _handlers[cur_index].arg);
-            if (result.is_ok() ||
-                result.error() == PARSE_ERROR_NOT_ENOUGH_DATA) {
-                m->set_preferred_index(cur_index);
-                *index = cur_index;
-                return result;
-            } else if (result.error() != PARSE_ERROR_TRY_OTHERS) {
-                // Critical error, return directly.
-                LOG_IF(ERROR, result.error() == PARSE_ERROR_TOO_BIG_DATA)
-                    << "A message from " << m->remote_side()
-                    << "(protocol=" << _handlers[cur_index].name
-                    << ") is bigger than " << FLAGS_max_body_size
-                    << " bytes, the connection will be closed."
-                    " Set max_body_size to allow bigger messages";
-                return result;
-            }
-
-            if (m->CreatedByConnect()) {
-                if((ProtocolType)cur_index == PROTOCOL_BAIDU_STD && cur_index == preferred) {
-                    // baidu_std may fall to streaming_rpc.
-                    cur_index = (int)PROTOCOL_STREAMING_RPC;
-                    continue;
-                } else if((ProtocolType)cur_index == PROTOCOL_STREAMING_RPC && cur_index == preferred) {
-                    // streaming_rpc may fall to baidu_std.
-                    cur_index = (int)PROTOCOL_BAIDU_STD;
-                    continue;
-                } else {
-                    // The protocol is fixed at client-side, no need to try others.
-                    LOG(ERROR) << "Fail to parse response from " << m->remote_side()
-                        << " by " << _handlers[preferred].name 
-                        << " at client-side";
-                    return MakeParseError(PARSE_ERROR_ABSOLUTELY_WRONG);
-                }
-            } else {
-                // Try other protocols.
-                break;
-            }
-        } while (true);
-        // Clear context before trying next protocol which probably has
-        // an incompatible context with the current one.
-        if (m->parsing_context()) {
-            m->reset_parsing_context(nullptr);
-        }
-        m->set_preferred_index(-1);
-    }
-    for (int i = 0; i <= max_index; ++i) {
-        if (i == preferred || _handlers[i].parse == nullptr) {
-            // Don't try preferred handler(already tried) or invalid handler
-            continue;
-        }
-        ParseResult result = _handlers[i].parse(&m->_read_buf, m, read_eof, _handlers[i].arg);
-        if (result.is_ok() ||
-            result.error() == PARSE_ERROR_NOT_ENOUGH_DATA) {
-            m->set_preferred_index(i);
-            *index = i;
-            return result;
-        } else if (result.error() != PARSE_ERROR_TRY_OTHERS) {
-            // Critical error, return directly.
-            LOG_IF(ERROR, result.error() == PARSE_ERROR_TOO_BIG_DATA)
-                << "A message from " << m->remote_side()
-                << "(protocol=" << _handlers[i].name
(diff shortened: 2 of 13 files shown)


bzs1118:
Thx for your solution. I tested it in our environment and it works.

chenBright:
> Thx for your solution. I tested it in our environment and it works.

@bzs1118 Are there still any problems?

bzs1118:
> > Thx for your solution. I tested it in our environment and it works.
> 
> @bzs1118 Are there still any problems?

No problems for now, I'll follow up here if we hit anything in production.

PR #3506, opened 2026-08-29
fix(cmake): prefer generated proto headers over in-source leftovers

darion-yaphet:

Issue Number: N/A

Problem Summary:

CMake searched `${PROJECT_SOURCE_DIR}/src` before the build directory. Make generates protobuf headers in-tree (`src/**/*.pb.h`, gitignored). After a `.proto` update, CMake already writes the new header into the build directory, but the compiler still picks the leftover in-source file.

This showed up after #3469 (`EPROGREADTIMEOUT`):

```
src/brpc/controller.cpp:76:28: error: no member named 'EPROGREADTIMEOUT' in namespace 'brpc'
```

`src/brpc/errno.proto` and `build/brpc/errno.pb.h` both had `EPROGREADTIMEOUT = 1019`; the stale `src/brpc/errno.pb.h` still stopped at `EREJECT = 1018`.

Changed:

- Add `${CMAKE_CURRENT_BINARY_DIR}` as a `BEFORE` include on `brpc_common_config`, so generated protobuf headers are searched before `src/` and leftover Make-generated `*.pb.h` cannot shadow them.

Side effects:
- Performance effects: none. Compile-time include order only; no runtime change.

- Breaking backward compatibility: no. Generated headers remain the source of truth. In-source `*.pb.h` from Make is still valid when it is up to date; it just no longer wins over a newer build-dir copy. Existing Make and Bazel flows are unchanged.

---
### Check List:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,7 +66,6 @@ set(BRPC_COMMON_COMPILE_OPTIONS)
 set(BRPC_COMMON_DEFINITIONS)
 set(BRPC_COMMON_INCLUDE_DIRS
     ${PROJECT_SOURCE_DIR}/src
-    ${CMAKE_CURRENT_BINARY_DIR}
 )
 set(BRPC_COMMON_LINK_OPTIONS)
 set(BRPC_CXX_STANDARD 14)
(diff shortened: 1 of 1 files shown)


darion-yaphet:
It seems there's no available limit left. Would you like to try again? @chenBright

chenBright (approved):
LGTM

wasphin:
Could the mixed directories of the generated *.pb.h files be caused by using both brpc_config.sh and CMake to build the project?

darion-yaphet:
> Could the mixed directories of the generated *.pb.h files be caused by using both brpc_config.sh and CMake to build the project?

Yes.

config_brpc.sh configures the Make-based build, which generates protobuf files in the source tree, while CMake generates them in the CMake build directory and copies public headers to build/output/include. 

Running both build systems in the same checkout can therefore leave generated *.pb.h files in multiple directories. CMake accounts for this by prioritizing build-directory headers, but a clean rebuild using only one build system is recommended.

Issue #3463, opened 2026-08-18
UBRing transport: timer lifecycle, backend-specific chunking, and maintainability issues


ZhengweiZhu:
**Describe the bug**

After reviewing and testing the UBRing/UBShmTransport introduced by #3290, we found several correctness, scalability, portability, performance, and maintainability issues.

The two shared-memory backends should be considered separately:

- `--ub_shm_type=1`: local POSIX shared memory (IPC)
- `--ub_shm_type=2`: remote shared memory provided by UBS

Some constraints that may be necessary for UBS do not apply to local IPC shared memory.

### 1. Deleting a timer may block indefinitely

Linux timerfds are created as blocking file descriptors:

```cpp
timerfd_create(CLOCK_MONOTONIC, 0);
```

Both `DeleteTimerInner()` and `DeleteTimerSafe()` execute a blocking read before closing the fd:

```cpp
uint64_t exp = 0;
read(fd, &exp, sizeof(exp));
close(fd);
```

There is a deterministic problematic path for one-shot timers:

1. `TimerEpoll()` reads and consumes the timerfd expiration.
2. It calls `UnifiedCallback()`.
3. For a non-periodic timer, `UnifiedCallback()` calls `DeleteTimerInner()`.
4. `DeleteTimerInner()` reads the same timerfd again.
5. Since the one-shot timer has already expired and was drained, the second blocking `read()` may never return.

For periodic timers, deletion may also block until the next expiration. For example, deleting a heartbeat timer may block for up to the heartbeat interval.

This can stall connection cleanup and delay releasing resources after a peer disconnects or crashes.

Timer deletion should not require a blocking read. If the custom implementation is retained, timerfds should at least use `TFD_NONBLOCK | TFD_CLOEXEC`, and `EAGAIN` should be treated as normal. Removing the fd from epoll and closing it without draining it may be sufficient.

### 2. The custom timer subsystem should preferably be replaced with brpc's existing timer implementation

The timer problems are not isolated defects. They are consequences of introducing a second timer framework based on timerfd/epoll alongside the timer framework already maintained by brpc.

brpc already provides:

- `bthread::TimerThread`
- `bthread_timer_add()` and `bthread_timer_del()`
- Versioned timer task IDs
- Cancellation semantics that account for a callback already running
- A shared global timer thread

[...]

PR #3507, opened 2026-08-30
Add UBRing data format negotiation

zchuango:

Issue Number: #3463

Problem Summary:

This is Phase 3 PR1 of Issue #3463.

Phase 3 is split into two parts:

- PR1 adds an explicit and backward-safe UBRing data format negotiation mechanism.
- A follow-up PR will introduce an IPC-specific data format and optimize the IPC data path.

UBRing currently uses the same legacy data format for both IPC and UBS, but the handshake only negotiates `hello_ver` and `impl_ver`. Before a follow-up PR can introduce an IPC-specific format, both peers must explicitly agree on the same data format.

The existing base Hello is already 64 bytes. Directly appending extension bytes is unsafe because an old peer may consume only 64 bytes and leave the extra bytes in the TCP stream, corrupting the following ACK or application data.

This PR only establishes the format-negotiation foundation. It does not introduce the new IPC data format or change the existing UBRing data path.

Changed:

- Bump the UBRing hello protocol version from V2 to V3 while keeping the base Hello wire format unchanged at 64 bytes.
- Add `UBR_DATA_FORMAT_NONE` and `UBR_DATA_FORMAT_LEGACY_64`.
- Add a separate fixed-size 4-byte format extension containing a 16-bit length and a 16-bit format identifier, both serialized in network byte order.
- Exchange the format extension only after both peers confirm V3 compatibility through the base Hello.
- Let the client propose `LEGACY_64`, and let the server select `LEGACY_64` or `NONE`.
- Map remote shared memory only after a supported, matching, non-`NONE` format is selected.
- Send ACK=0 and fall back to TCP when format negotiation fails.
- Do not exchange format-extension bytes between V2 and V3 peers.
- Require the base Hello `msg_len` to be exactly 64 so that unsupported extra Hello bytes cannot remain unread in the TCP stream.
- Record and reset the negotiated data format in `UBShmEndpoint`.
- Add focused tests for serialization/deserialization, network byte order, `NONE`, unknown format values, and negotiated-format state cleanup.
- Keep the existing UBRing message layout, 64-byte slot, 60-byte payload, `Copy64Byte`, memory ordering, and send/receive data path unchanged.

The V3 format extension is intentionally fixed at 4 bytes. A different extension wire size requires negotiation through a future hello version.

Tests:

```text
bazel test --config=ubring //test:brpc_ubring_unittest
```

```text
15 tests from 4 test suites passed.
```

Smoke-test results:

- V3 client + V3 server negotiated `LEGACY_64` and completed RPC requests.
- V3 client + baseline V2 server fell back to TCP and completed 10 RPC requests.
- Baseline V2 client + V3 server fell back to TCP and completed 10 RPC requests.

Side effects:
- Performance effects:

  V3-to-V3 connections add one fixed 4-byte request, one fixed 4-byte response, and one handshake round trip. The established UBRing data path has no additional per-message overhead.

  No unsafe wire compatibility break is introduced. V2 and V3 peers intentionally fall back to TCP without exchanging format-extension bytes. TCP RPC traffic remains functional.

---
### Check List:

### DIFF
--- a/src/brpc/ubshm/ub_endpoint.cpp
+++ b/src/brpc/ubshm/ub_endpoint.cpp
@@ -61,12 +61,32 @@ static const size_t MAGIC_STR_LEN = 2;
 static const size_t HELLO_MSG_LEN_MIN = 64;
 static const size_t ACK_MSG_LEN = 4;
 static uint16_t g_ub_hello_msg_len = 64;
-static uint16_t g_ub_hello_version = 2;
+static uint16_t g_ub_hello_version = 3;
 static uint16_t g_ub_impl_version = 1;
 
 static const uint32_t ACK_MSG_UB_OK = 0x1;
 
 static butil::Mutex* g_ubring_resource_mutex = nullptr;
+
+void HelloFormatExtension::Serialize(void* data) const {
+    char* current_pos = static_cast<char*>(data);
+    const uint16_t net_extension_len = butil::HostToNet16(extension_len);
+    memcpy(current_pos, &net_extension_len, sizeof(net_extension_len));
+    current_pos += sizeof(net_extension_len);
+    const uint16_t net_format_id = butil::HostToNet16(format_id);
+    memcpy(current_pos, &net_format_id, sizeof(net_format_id));
+}
+
+void HelloFormatExtension::Deserialize(const void* data) {
+    const char* current_pos = static_cast<const char*>(data);
+    uint16_t net_extension_len;
+    memcpy(&net_extension_len, current_pos, sizeof(net_extension_len));
+    extension_len = butil::NetToHost16(net_extension_len);
+    current_pos += sizeof(net_extension_len);
+    uint16_t net_format_id;
+    memcpy(&net_format_id, current_pos, sizeof(net_format_id));
+    format_id = butil::NetToHost16(net_format_id);
+}
 
 void HelloMessage::Serialize(void* data) const {
     char* current_pos = static_cast<char*>(data);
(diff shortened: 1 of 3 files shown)


zchuango:
@wwbmmm @chenBright Could you help confirm the scope of the two Copilot comments about UB resources not being released immediately after TCP fallback?
I checked the existing code and found that this behavior is not specific to the new format negotiation. The existing version-mismatch and remote-mapping-failure paths also keep the resources until the socket is destroyed.
Simply calling DeallocateResources() on fallback may not be safe, since it calls UbrTrxClose() and may wait for a peer even though the UBRing connection was never established.
Do you think this should be addressed in this PR, or would a separate cleanup PR covering all handshake fallback paths be more appropriate?

wwbmmm:
> @wwbmmm @chenBright Could you help confirm the scope of the two Copilot comments about UB resources not being released immediately after TCP fallback? I checked the existing code and found that this behavior is not specific to the new format negotiation. The existing version-mismatch and remote-mapping-failure paths also keep the resources until the socket is destroyed. Simply calling DeallocateResources() on fallback may not be safe, since it calls UbrTrxClose() and may wait for a peer even though the UBRing connection was never established. Do you think this should be addressed in this PR, or would a separate cleanup PR covering all handshake fallback paths be more appropriate?

I think you can ignore these Copilot comments.

wwbmmm:
LGTM

wwbmmm:
LGTM

src/brpc/ubshm/ub_endpoint.cpp
```diff
@@ -352,7 +374,7 @@ void* UBShmEndpoint::ProcessHandshakeAtClient(void* arg) {
     }
 
     ep->_state = C_HELLO_SEND;
-    HelloMessage local_msg;
+    HelloMessage local_msg = {};
```
chenBright: HelloMessage local_msg{};

chenBright (approved):
LGTM

PR #3511, opened 2026-08-31
Enforce internal_port gating of builtin services in pb protocols

chenBright:

Issue Number: resolve 

Problem Summary:

  `ServerOptions.internal_port` is documented to make builtin services (and
  Tabbed services, see the comment on `internal_port`) reachable only from the
  internal port. `ServerOptions::security_mode()` returns true once it is set.

  Only the http/h2 dispatch path actually enforced this. Every pb protocol
  computed the same `security_mode` flag but used it just for address obfuscation
  in `AppendServerIdentiy()`, and had no check at all between method resolution
  and `CallMethod()`. Since builtin services share `_method_map` with user
  services, they can be addressed by name over pb.

  With `internal_port` configured, a client on the public port can therefore reach builtin 
  services that the http path returns `403 Forbidden` for on the very same server. For 
  example, `brpc.vars` dumps the full metrics set, and`hotspots` is reachable the same 
  way.

  Affected dispatch paths: `baidu_std`, `hulu_pbrpc`, `sofa_pbrpc`, and all
  nshead-based pb protocols (`public_pbrpc`, `nshead_mcpack`, `nova_pbrpc`,
  `ubrpc2pb`), which share `NsheadPbServiceAdaptor::ProcessNsheadRequest`.

Changed:

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/brpc/details/server_private_accessor.h
+++ b/src/brpc/details/server_private_accessor.h
@@ -104,6 +104,29 @@ private:
     const Server* _server;
 };
 
+// Reject accesses to builtin services when the server is in security mode,
+// in which case they are only reachable from ServerOptions.internal_port.
+// Returns true if the access was rejected, in which case `cntl` was already
+// SetFailed() and the caller must stop dispatching the request immediately.
+// NOTE: Call this after ControllerPrivateAccessor::set_security_mode() and
+// before the method is counted by MethodStatus::OnRequested(), so that
+// rejected accesses do not pollute the stats of the method. `mp` may point
+// to BadMethodService which is builtin as well and lists the methods of the
+// requested service, so protocols dispatching to BadMethodService must call
+// this beforehand, or make sure the listing is hidden in security mode.
+inline bool RejectBuiltinAccess(Controller* cntl, const Server& server,
+                                const Server::MethodProperty* mp) {
+    if (!cntl->is_security_mode() ||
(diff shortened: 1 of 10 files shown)


chenBright:
@wasphin PTAL.

src/brpc/details/server_private_accessor.h
```diff
@@ -104,6 +104,29 @@ class ServerPrivateAccessor {
...
+// in which case they are only reachable from ServerOptions.internal_port.
+// Returns true if the access was rejected, in which case `cntl` was already
+// SetFailed() and the caller must stop dispatching the request immediately.
+// NOTE: Call this after ControllerPrivateAccessor::set_security_mode() and
+// before the method is counted by MethodStatus::OnRequested(), so that
+// rejected accesses do not pollute the stats of the method. `mp` may point
+// to BadMethodService which is builtin as well and lists the methods of the
+// requested service, so protocols dispatching to BadMethodService must call
+// this beforehand, or make sure the listing is hidden in security mode.
+inline bool RejectBuiltinAccess(Controller* cntl, const Server& server,
```
wasphin, line 117: Maybe we can add a PreCallValidator like or so that check all rpcs inside in the future.

wasphin (approved):
LGTM

src/brpc/details/server_private_accessor.h
chenBright: Good idea!

PR #3512, opened 2026-08-31
Close two paths that reach services past their intended gate

chenBright:

Issue Number: resolve 

Problem Summary:

Paths that differ only in empty segments dispatched identically. `FindMethodPropertyByURIImpl` splits 
the path with a `StringSplitter` that skips empty fields, so `//flags`, `/flags//` and `/flags//port` all resolved 
to the same method as `/flags`. RFC 3986 treats `//foo` and `/foo` as distinct paths, so a front proxy whose 
ACL matches the collapsed form does not match the padded one and passes it through — `//flags?setvalue=` 
reaches a builtin service that `/flags` cannot. Normalizing the path server-side would not help, because the 
proxy has already forwarded the padded literal; only rejecting it removes the differential.

The same audit found that HTTP/2 never validated `:path`. RFC 9113 8.3.1 requires  it to be non-empty and, 
apart from the asterisk-form used by `OPTIONS`, to begin  with `/`. 

Changed:

  - `FindMethodPropertyByURI` rejects any path containing `//`. The check lives in
    the funnel rather than in `FindMethodPropertyByURIImpl`, so it also covers the
    global restful map, which does its own `NormalizeSlashes`.
  - `H2StreamContext::ConsumeHeaders` rejects a `:path` that is empty or does not
    begin with `/`, unless it is exactly `*`.

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/brpc/policy/http2_rpc_protocol.cpp
+++ b/src/brpc/policy/http2_rpc_protocol.cpp
@@ -1273,6 +1273,16 @@ int H2StreamContext::ConsumeHeaders(butil::IOBufBytesIterator& it) {
             case 'p':
                 if (strcmp(name + 2, /*p*/"ath") == 0) {
                     matched = true;
+                    // RFC 9113 8.3.1: :path MUST NOT be empty and MUST begin
+                    // with '/', the only exception being the asterisk-form
+                    // that OPTIONS uses. '*' is accepted for any method here,
+                    // as http_parser does for HTTP/1: pinning it to OPTIONS
+                    // would take the whole header block, since HPACK does not
+                    // order pseudo-headers and :method may not have arrived.
+                    if (pair.value != "*" && (pair.value.empty() || pair.value[0] != '/')) {
+                        LOG(ERROR) << "Invalid path=" << pair.value;
+                        return -1;
+                    }
                     // Including path/query/fragment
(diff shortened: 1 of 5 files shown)


chenBright:
@wasphin PTAL.

src/brpc/policy/http2_rpc_protocol.cpp
```diff
@@ -1373,6 +1273,14 @@ int H2StreamContext::ConsumeHeaders(butil::IOBufBytesIterator& it) {
             case 'p':
                 if (strcmp(name + 2, /*p*/"ath") == 0) {
                     matched = true;
+                    // RFC 9113 8.3.1: :path MUST NOT be empty and, apart from
+                    // the asterisk-form that OPTIONS may use, MUST begin with
+                    // '/'.
+                    if (pair.value != "*" &&
```
wasphin: Any chance here to check the `OPTIONS` requests? Not a must though.

wasphin (approved):
LGTM

wwbmmm:
LGTM

src/brpc/policy/http2_rpc_protocol.cpp
chenBright: No, :method may not have arrived.

PR #3513, opened 2026-08-31
Fix hang in stream ID validation test

wasphin:

Issue Number: N/A

Problem Summary:

The mismatched returned stream ID test could hang in `Server::Join`. The
intentionally rejected response left server-side accepted streams open.

Related discussion: https://github.com/apache/brpc/pull/3485#discussion_r3889720756

Changed:

The test now retains and closes the server-side accepted streams before
stopping and joining the server.

Side effects:
- Performance effects: None. The change only affects a unit test.

---
### Check List:


chenBright (commented):
LGTM

PR #3514, opened 2026-09-01
Fix SelectiveChannel response copy for non-reflectable messages

### LINKED ISSUE #3372: selective_channel SubDone::Run最后的message交换可能会有内存问题
**Describe the bug**
selective_channel.cpp:
Image
Image

对于一些无反射的Message（RedisResponse）, GetRelection()会返回空指针。
看了下PopFree()方法，之前的版本response成员，会使用用户传下来的response，
从而不会触发message交换。

**To Reproduce**
CallMethod使用RedisRequest和RedisResponse时，由于触发该问题。

**Expected behavior**
会引用空指针，触发SIGSEGV，反生coredump

**Versions**
OS: CentOS
Compiler: gcc
brpc: 1.17.0
protobuf: 3.25.8

zchuango:

Issue Number: resolve #3372

Problem Summary:

`SelectiveChannel` uses protobuf reflection to swap a successful sub-call response back to the main response. For non-reflectable messages such as `RedisResponse`, `GetReflection()` is unavailable, which may lead to a null pointer dereference and crash.

Changed:

- Add a common interface for `NonreflectableMessage` to copy responses of the same type without protobuf reflection.
- Use the non-reflectable copy path in `SelectiveChannel`, while keeping the existing reflection-based `Swap` for normal protobuf messages.
- Add protobuf 29+ `MessageCreator` support so `Message::New()` can create non-reflectable responses correctly.
- Add a Redis-backed regression test for using `RedisResponse` with `SelectiveChannel`.

Side effects:
- Performance effects:

No significant performance impact is expected. The additional type check only occurs when transferring the completed sub-call response.

No. The existing reflection-based path for normal protobuf messages remains unchanged.

---
### Check List:
- Please follow [[Contributor Covenant Code of Conduct](https://github.com/apache/brpc/blob/master/CODE_OF_CONDUCT.md)](https://github.com/apache/brpc/blob/master/CODE_OF_CONDUCT.md).

### DIFF
--- a/src/brpc/nonreflectable_message.h
+++ b/src/brpc/nonreflectable_message.h
@@ -25,6 +25,11 @@
 
 namespace brpc {
 
+class NonreflectableMessageBase : public ::google::protobuf::Message {
+public:
+    virtual bool CopyFromSameType(const ::google::protobuf::Message& other) = 0;
+};
+
 //
 // In bRPC, some non-Protobuf based protocol messages are also designed to implement
 // Protobuf Message interfaces, to provide a unified protocol message.
@@ -38,7 +43,7 @@ namespace brpc {
 // and use only #if version_check #endif, to make maintenance easier.
 //
 template <typename T>
-class NonreflectableMessage : public ::google::protobuf::Message {
+class NonreflectableMessage : public NonreflectableMessageBase {
 public:
     inline NonreflectableMessage() = default;
     inline NonreflectableMessage(const NonreflectableMessage&) : NonreflectableMessage() {}
@@ -95,6 +100,18 @@ public:
         }
         Clear();
         MergeFrom(other);
+    }
+
+    bool CopyFromSameType(const ::google::protobuf::Message& other) override {
+        if (&other == this) {
+            return true;
+        }
+        if (other.GetDescriptor() != descriptor()) {
+            return false;
+        }
+        Clear();
+        MergeFrom(static_cast<const T&>(other));
+        return true;
     }
 
     void MergeFrom(const ::google::protobuf::Message& other) PB_526_OVERRIDE {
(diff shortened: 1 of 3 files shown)


src/brpc/nonreflectable_message.h
```diff
@@ -25,6 +25,11 @@
 
 namespace brpc {
 
+class NonreflectableMessageBase : public ::google::protobuf::Message {
+public:
+    virtual bool CopyFromSameType(const ::google::protobuf::Message& other) = 0;
+};
+
```
chenBright, line 32: It seems that NonreflectableMessageBase is unnecessary; it's sufficient to use NonreflectableMessage directly.

src/brpc/selective_channel.cpp
```diff
@@ -394,8 +395,14 @@ void SubDone::Run() {
...
-            main_cntl->_response->GetReflection()->Swap(
-                main_cntl->_response, _cntl._response);
+            NonreflectableMessageBase* nr_msg =
+                dynamic_cast<NonreflectableMessageBase*>(main_cntl->_response);
+            if (nr_msg != nullptr) {
+                CHECK(nr_msg->CopyFromSameType(*_cntl._response));
+            } else {
+                main_cntl->_response->GetReflection()->Swap(
+                    main_cntl->_response, _cntl._response);
+            }
```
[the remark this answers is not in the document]
zchuango: Handled in the latest update. CopyFromSameType failure now sets ERESPONSE instead of aborting on CHECK.

wasphin (approved):
LGTM

src/brpc/nonreflectable_message.h
wasphin: `NonreflectableMessage` is a tempalted class that can not be directly used for base checking.

src/brpc/nonreflectable_message.h
```diff
@@ -229,6 +246,17 @@ class NonreflectableMessage : public ::google::protobuf::Message {
 private:
     static T _instance;
 
+#if GOOGLE_PROTOBUF_VERSION >= 5029000
+    static void* PlacementNew_(const void*, void* mem,
```
chenBright, line 250: `PlacementNew_ ` -> `PlacementNew`

src/brpc/nonreflectable_message.h
chenBright: Got it.

src/brpc/nonreflectable_message.h
cw20050111-prog: Thanks for taking a look. I may be misunderstanding this comment, but there doesn’t appear to be a space between `PlacementNew_` and `(` in my original code. Could you please clarify what issue or change you’re referring to here?

chenBright: > `PlacementNew_ ` -> `PlacementNew`

@cw20050111-prog fixed.

Huixxi:
LGTM

PR #3517, opened 2026-09-03
Replace popen('uname -ap') with uname() syscall in bvar kernel_version

weim0000:

Problem Summary:

When a brpc server has allocated a large amount of memory, the first request to the `/vars` endpoint can cause a significant latency stall. This is because the `kernel_version` bvar variable is lazily initialized on first access, and its constructor calls `popen("uname -ap")` to read the kernel version.

Internally, `popen()` calls `fork()` to spawn a child process. On Linux, `fork()` needs to duplicate the parent process's page tables. For a server with a large memory footprint (e.g., tens of GBs), this can take hundreds of milliseconds or even longer, effectively blocking the bthread that handles the `/vars` request.

This caused a production incident in our environment, where the service appeared to hang when the monitoring system first scraped the `/vars` endpoint after the server had been running for a while with heavy memory usage.

Changed:

Replace `butil::read_command_output(oss, "uname -ap")` (which shells out via `popen` → `fork` → `exec`) with the POSIX `uname()` syscall in `src/bvar/default_variables.cpp`. The `uname()` syscall reads the same kernel information directly via `struct utsname`, without creating any child process. The output format remains equivalent to `uname -ap`. A unit test is added in `test/bvar_variable_unittest.cpp`.

Side effects:
- Performance effects: Eliminates the `fork()` overhead entirely. The `uname()` syscall completes in microseconds regardless of the process's memory usage, whereas the previous `popen()` approach could stall for 100ms+ on large-memory processes.

- Breaking backward compatibility: No. The output format of the `kernel_version` bvar remains the same.

### DIFF
--- a/src/bvar/default_variables.cpp
+++ b/src/bvar/default_variables.cpp
@@ -20,6 +20,7 @@
 #include <unistd.h>                        // getpagesize
 #include <sys/types.h>
 #include <sys/resource.h>                  // getrusage
+#include <sys/utsname.h>                   // uname
 #include <dirent.h>                        // dirent
 #include <iomanip>                         // setw
 #include <stdio.h>
@@ -617,11 +618,15 @@ static void get_cmdline(std::ostream& os, void*) {
 struct ReadVersion {
     std::string content;
     ReadVersion() {
-        std::ostringstream oss;
-        if (butil::read_command_output(oss, "uname -ap") != 0) {
+        struct utsname buf;
+        if (uname(&buf) != 0) {
             LOG(ERROR) << "Fail to read kernel version";
             return;
         }
+        std::ostringstream oss;
+        oss << buf.sysname << ' ' << buf.nodename << ' '
+            << buf.release << ' ' << buf.version << ' '
+            << buf.machine << ' ' << buf.machine;
         content.append(oss.str());
     }
 };
--- a/src/bvar/default_variables.cpp
+++ b/src/bvar/default_variables.cpp
@@ -633,7 +633,11 @@ struct ReadVersion {
         std::ostringstream oss;
         oss << buf.sysname << ' ' << buf.nodename << ' '
             << buf.release << ' ' << buf.version << ' '
-            << buf.machine << ' ' << processor << '\n';
+            << buf.machine << ' ' << processor;
+#if !defined(__APPLE__)
+        oss << " GNU/Linux";
+#endif
+        oss << '\n';
         content.append(oss.str());
     }
 };
--- a/src/bvar/default_variables.cpp
+++ b/src/bvar/default_variables.cpp
@@ -623,10 +623,17 @@ struct ReadVersion {
             LOG(ERROR) << "Fail to read kernel version";
             return;
         }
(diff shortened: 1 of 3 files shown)


src/bvar/default_variables.cpp
```diff
@@ -617,11 +618,26 @@ static void get_cmdline(std::ostream& os, void*) {
...
+        const char* processor = "arm";
+#elif defined(__APPLE__) && defined(__x86_64__)
+        const char* processor = "i386";
+#else
+        const char* processor = buf.machine;
+#endif
+        std::ostringstream oss;
+        oss << buf.sysname << ' ' << buf.nodename << ' '
+            << buf.release << ' ' << buf.version << ' '
+            << buf.machine << ' ' << processor;
```
wwbmmm, line 636: The reconstructed output is not byte-equivalent to `uname -ap` as stated in the PR. On Linux, `uname -ap` appends the operating-system identifier as the final field, e.g. `... aarch64 GNU/Linux`; this new code stops at `machine processor` and drops that OS field entirely. Additionally, `uname(1)` omits a field whose value is the literal string "unknown" (common for `uname -p`), whereas this code always appends `processor` (here equal to `buf.machine`, producing a duplicated `machine` value).

If exact parity with the previous output is required (e.g. for scrapers/keying off the string), either append the proper OS name field (`GNU/Linux` on Linux; none on macOS) and mirror uname's handling of "unknown", or update the PR description to not claim the output format is preserved. Since these fields are informational this may be acceptable, but the claim and the format should match.

---

wwbmmm (commented):
Good fix that removes the fork()/popen stall on the /vars hot path; however the output is not strictly equivalent to `uname -ap`, so the no-compatibility-break claim needs to be corrected or the format adjusted.

---

src/bvar/default_variables.cpp
weim0000: 'GNU/Linux' has been added here. Please take a look at the following four lines of code

weim0000: @wwbmmm Please check the latest code，thanks

src/bvar/default_variables.cpp
```diff
@@ -617,11 +618,26 @@ static void get_cmdline(std::ostream& os, void*) {
...
+#if !defined(__APPLE__)
+        oss << " GNU/Linux";
+#endif
+        oss << '\n';
```
weim0000, line 640: @wwbmmm here

wwbmmm:
LGTM. The header-only make_kernel_version_string helper is shared by production and tests, the GNU/Linux suffix is correctly gated to Linux (non-Android), the failure log now carries errno, and my earlier unused-variable concern is resolved. No remaining blockers.

---

weim0000:
@wwbmmm copilot has triggered the limit

weim0000:
@wwbmmm copilot has reached the limit

wwbmmm:
> @wwbmmm copilot has reached the limit

Sorry for the copilot spam, I have fixed it.

wwbmmm:
LGTM
The CI failure seems unrelated to this PR.

yanglimingcn:
LGTM

PR #3521, opened 2026-09-04
Reject HTTP/1 header forms that desync framing, drop chunked trailers

chenBright:

Issue Number: resolve 

Problem Summary:

  Four header forms are parsed here in a way a conforming front-end does not, and
  one legal form is not parsed at all. In a `client -> proxy -> bRPC` chain both
  parties parse the same bytes independently, so every disagreement about where a
  message ends turns the tail of message N into the head of message N+1 for one of
  them. The primitive behind request smuggling and response desync.

  1. **`Content-Length : 5`** — this fork's `TOKEN(c)` maps SP to itself. [Upstream](https://github.com/nodejs/http-parser)
     has a `STRICT_TOKEN()` and a `HTTP_PARSER_STRICT` branch. Both were dropped
     here, leaving the lenient path as the only path. A space before the colon
     therefore keeps `header_state` on the framing header, so the value still
     drives our framing while the field name we report is `"Content-Length "`.
     RFC 7230 3.2.6 excludes SP from `tchar`, and 3.2.4 forbids whitespace before
     the colon.

  2. **`Content-Length: 1 3`** — interior blanks were skipped, so we read 13 where
     a proxy reads 1 or rejects. Same class as CVE-2022-32213, fixed upstream by
     http-parser https://github.com/nodejs/http-parser/commit/01da95f and https://github.com/nodejs/http-parser/commit/cd88eef.
     Neither was picked up here.

  3. **`Transfer-Encoding:\r\n chunked`, `Content-Length:\r\n 5`** — both obs-fold
     states (RFC 7230 3.2.4) carry `header_state` across the fold, so a folded
     framing value still drove our framing while a front-end that does not unfold
     sees a different message.

  4. **chunked trailers** — after the last chunk the parser sets `F_TRAILING` and
     returns to `s_header_field_start`, so trailer fields reach the same
     `on_header_field` / `on_header_value` callbacks as the header section. Neither
     looked at `F_TRAILING`, so a trailer landed in the header map like a real
     header: appended to an existing entry of the same name, or added as a new one.
     A front-end only filters the header section, so any chunked request could put
     an `Authorization` or `X-Forwarded-For` past it:

         POST / HTTP/1.1
         Host: a.com
         X-Forwarded-For: 10.0.0.1        <- set by the proxy
         Transfer-Encoding: chunked

         5
         hello
         0
         X-Forwarded-For: 6.6.6.6         <- invisible to the proxy, appended by us
         Authorization: Bearer stolen

  5. **`Transfer-Encoding: gzip, chunked`** — found while auditing the `TOKEN()`
     call sites for (1). `TOKEN(' ')` is truthy, which made the
     `else if (c == ' ' || c == '\t')` arm of
     `h_matching_transfer_encoding_token_start` dead code: the OWS after the comma
     started a new coding name, so `chunked` never matched. This is the form
     RFC 7230 7 spells out. Requests were rejected with
     `HPE_INVALID_TRANSFER_ENCODING` per RFC 7230 3.3.3, and responses were read to
     EOF instead of as chunked.

Changed:
 
- Reintroduce `STRICT_TOKEN()`. This fork's `tokens[]` already omits SP as
   RFC 7230 3.2.6 does, so the strict form is the plain lookup and `TOKEN()` is
   the leniency layered on top.

- Reject SP before the colon of `Content-Length`, `Transfer-Encoding`,
  `Connection` and `Upgrade`. Those four are the only terminal `header_state`s
   reachable in that switch arm. Any other name is `h_general` or still
   matching, and keeps the historical leniency unless the new flag is on.
- Backport https://github.com/nodejs/http-parser/commit/01da95f and https://github.com/nodejs/http-parser/commit/cd88eef.

- Reject obs-fold of a value whose `header_state` decides where the body ends
   (`is_framing_header_state()`, i.e. the Content-Length and Transfer-Encoding
   states). Both fold states need the check: `s_header_value_lws` for a value
   that has started, `s_header_value_discard_lws` for one that has not.

    - Use `STRICT_TOKEN()` unconditionally in `h_matching_transfer_encoding_token_start`, 
       as upstream does, so the OWS after a comma is skipped instead of starting a coding name.

    - Add the `h_content_length*`, `h_transfer_encoding`, `h_upgrade` and
      `h_matching_transfer_encoding_*` cases missing from
      `http_parser_header_state_name()`, which printed `h_unknown` for them.

- `on_header_field` and `on_header_value` return early when `F_TRAILING` is
    set, so trailers are discarded rather than merged. `parser->flags` is
    cleared at the start of each message, so the check is scoped to one trailer.

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/brpc/details/http_message.cpp
+++ b/src/brpc/details/http_message.cpp
@@ -33,6 +33,13 @@ namespace brpc {
 
 DEFINE_bool(allow_chunked_length, false,
             "Allow both Transfer-Encoding and Content-Length headers are present.");
+DEFINE_bool(http_allow_obs_fold, false,
+            "Allow obsolete line folding (RFC 7230 3.2.4) of Content-Length "
+            "and Transfer-Encoding, which front proxies may parse differently.");
+DEFINE_bool(http_strict_header_token, false,
+            "Reject a space inside any header field name (RFC 7230 3.2.6), "
+            "not just inside the names that decide framing. Names carrying no "
+            "framing semantics are accepted by default for compatibility.");
 DEFINE_bool(allow_http_1_1_request_without_host, true,
             "Allow HTTP/1.1 request without host which violates the HTTP/1.1 specification.");
 DEFINE_bool(http_verbose, false,
@@ -83,6 +90,15 @@ int HttpMessage::on_status(http_parser *parser, const char *, const size_t) {
 // change the order of these field values when a message is forwarded. 
 int HttpMessage::on_header_field(http_parser *parser,
                                  const char *at, const size_t length) {
+    if (parser->flags & F_TRAILING) {
+        // Drop chunked trailers instead of merging them into the header map.
+        // A front-end that strips or rewrites headers (Authorization,
+        // X-Forwarded-For, ...) only looks at the header section, so anything
+        // accepted here would be a header smuggled past it. HTTP/1 trailers
+        // have no reader API in bRPC anyway. Supporting them later should add
+        // a separate map rather than reuse this one.
+        return 0;
+    }
     HttpMessage *http_message = (HttpMessage *)parser->data;
     if (http_message->_stage != HTTP_ON_HEADER_FIELD) {
         http_message->_stage = HTTP_ON_HEADER_FIELD;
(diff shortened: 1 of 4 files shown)


wwbmmm:
LGTM

PR #3523, opened 2026-09-04
Freeze bvar dump path flags and cap pprof duration by FLAGS_max_profiling_seconds

chenBright:

Issue Number: resolve 

Problem Summary:

  Two builtin services let whoever reaches them do more than the surrounding
  configuration intends.

  #### `/flags` can repoint the bvar dump files.

  The bvar dump flags carry an accept-anything validator whose only job is to 
  wake the dumping thread:

  ```cpp
  // src/bvar/variable.cpp
  static bool wakeup_dumping_thread(const char*, const std::string&) {
      pthread_cond_signal(&dump_cond);
      return true;
  }
  ```

  The dumping thread re-reads those flags every round and ends up in
  `FileDumper::dump_impl()`:

  ```cpp
  butil::CreateDirectoryAndGetError(dir, &error);   // recursive mkdir
  _fp = fopen(_filename.c_str(), "w");              // truncate
  ```

  So two requests against a server with default options:

  ```
  GET /flags/bvar_dump_file?setvalue=/victim/dir/target
  GET /flags/bvar_dump?setvalue=true
  ```

  create a directory chain anywhere and truncate a file inside it, with the
  privileges of the server. `setvalue` is a GET, so this is reachable from a
  browser as well.

  The primitive is bounded but real. `bvar_dump_file` and `bvar_dump_tabs` get a
  `.data` suffix forced on them by `FilePath::AddExtension()`, and the bytes
  written are bvar names and values, not attacker input. `mbvar_dump_file` is
  used verbatim with no suffix. `bvar_dump_tabs` names files too: its tab names
  go through the same `AddExtension()`, which appends them to the path with no
  sanitization, so `../` in a tab name escapes the configured directory. The
  recursive mkdir and the truncation are unrestricted in every case.

  #### `/pprof` ignores `FLAGS_max_profiling_seconds`, which `/hotspots` enforces.

  ```cpp
  // src/brpc/builtin/pprof_service.cpp, ReadSeconds(), before this PR
      return seconds;                                        // no upper bound
  ```
  ```cpp
  // src/brpc/builtin/hotspots_service.cpp:233
      seconds = std::min(seconds, FLAGS_max_profiling_seconds);
  ```

  `?seconds=2147483647` reaches `bthread_usleep(sleep_sec * 1000000L)`, roughly
  68 years. The profilers are process wide -- gperftools' `ProfilerStart` and
  bthread's `g_cp` are each a single global, so this is not merely one hung
  RPC: it holds the only CPU or contention profiler for as long as it asked for,
  and `/hotspots/cpu` and `/hotspots/contention` answer 503 the whole time. The
  300 second limit was enforced on only one of the two paths to the same
  resource, so the path without it decides the limit for both.

Changed:

- Drop the validator registration for -bvar_dump_file`, `-bvar_dump_tabs` and 
   `FLAGS_mbvar_dump_file`.
- Clamp `ReadSeconds()` to `FLAGS_max_profiling_seconds`.

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/brpc/builtin/common.cpp
+++ b/src/brpc/builtin/common.cpp
@@ -26,12 +26,16 @@
 #include "butil/third_party/murmurhash3/murmurhash3.h"
 #include "butil/process_util.h"              // ReadCommandLine
 #include "brpc/server.h"
+#include "brpc/reloadable_flags.h"
 #include "brpc/builtin/common.h"
 
 namespace brpc {
 
 DEFINE_string(rpc_profiling_dir, "./rpc_data/profiling",
               "For storing profiling results.");
+
+DEFINE_int32(max_profiling_seconds, 300, "upper limit of running time of profilers");
+BRPC_VALIDATE_GFLAG(max_profiling_seconds, PositiveInteger);
 
 bool UseHTML(const HttpHeader& header) {
     const std::string* console = header.uri().GetQuery(CONSOLE_STR);
--- a/src/brpc/builtin/common.h
+++ b/src/brpc/builtin/common.h
@@ -56,6 +56,7 @@ enum ProfilingType {
 };
 
 DECLARE_string(rpc_profiling_dir);
+DECLARE_int32(max_profiling_seconds);
 
 bool UseHTML(const HttpHeader& header);
 bool MatchAnyWildcard(const std::string& name,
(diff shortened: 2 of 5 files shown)


wwbmmm:
LGTM

PR #3525, opened 2026-09-05
Serve builtin services only on ServerOptions.internal_port

chenBright:

Issue Number: resolve 

Problem Summary:

 Authentication configured through `ServerOptions.auth` can be bypassed
  entirely on a server that also sets `ServerOptions.internal_port`.

  `VerifyHttpRequest()` returns true without checking any credential when a
  builtin service is requested on `internal_port`. That verdict is not
  per-request: `verify()` runs only for the first message of a connection
  and the result latches the connection. Sending `GET /status` to 
  `internal_port` first therefore authenticates the connection, and everything 
  sent next on it is dispatched without ever calling `verify()`.

Changed:

  `internal_port` now carries builtin and Tabbed services only. Requests for
  ordinary services are rejected with `EPERM` (HTTP 403) and must go to the
  port passed to `Server::Start()`. With nothing but builtin services served
  there, the latch has nothing left to unlock.

Side effects:
- Performance effects:

---
### Check List:

### DIFF
--- a/src/brpc/policy/http_rpc_protocol.cpp
+++ b/src/brpc/policy/http_rpc_protocol.cpp
@@ -1558,7 +1558,8 @@ void ProcessHttpRequest(InputMessageBase *msg) {
         return;
     }
 
-    if (server->options().http_master_service) {
+    if (server->options().http_master_service &&
+        !IsInternalPort(*server, cntl->local_side())) {
         // If http_master_service is on, just call it.
         google::protobuf::Service* svc = server->options().http_master_service;
         const google::protobuf::MethodDescriptor* md =
@@ -1605,7 +1606,8 @@ void ProcessHttpRequest(InputMessageBase *msg) {
         mp->service->CallMethod(mp->method, cntl, &breq, &bres, nullptr);
         return;
     }
-    if (RejectBuiltinAccess(cntl, *server, mp)) {
+    if (RejectBuiltinAccess(cntl, *server, mp) ||
+        RejectNonBuiltinAccessFromInternalPort(cntl, *server, mp)) {
         return;
     }
     // Switch to service-specific error.
--- a/test/brpc_socket_unittest.cpp
+++ b/test/brpc_socket_unittest.cpp
@@ -118,6 +118,17 @@ class CheckRecycle : public brpc::SocketUser {
     }
 };
 
+// Acceptor::Join() returns as soon as Acceptor::BeforeRecycle() cleared
+// _listened_fd, which Socket::BeforeRecycled() calls several statements before
+// it closes the fd, so the fd may still be open when Join() returns.
+void WaitForClosedFd(int fd) {
+    for (int i = 0; i < 100 && fcntl(fd, F_GETFD) != -1; ++i) {
+        bthread_usleep(10000);
+    }
+    ASSERT_EQ(-1, fcntl(fd, F_GETFD));
+    ASSERT_EQ(EBADF, errno);
+}
+
 // Never hardcode a port in tests: it may be taken by another process or by
 // another test in this binary, making the test fail for unrelated reasons.
 // Listen on port 0 and ask the kernel which port it picked instead.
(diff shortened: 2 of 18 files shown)


src/brpc/policy/http_rpc_protocol.cpp
```diff
@@ -1605,7 +1605,8 @@ void ProcessHttpRequest(InputMessageBase *msg) {
         mp->service->CallMethod(mp->method, cntl, &breq, &bres, nullptr);
         return;
     }
-    if (RejectBuiltinAccess(cntl, *server, mp)) {
+    if (server->RejectBuiltinAccess(cntl, mp) ||
```
wwbmmm, line 1608: The new gate is placed after the `server->options().http_master_service` early-dispatch block (around line 1543), which calls into the user-provided http_master_service and returns before reaching this check. http_master_service is a non-builtin user service, so on a server that sets both `internal_port` and `http_master_service`, an ordinary (non-builtin) service is still reachable on internal_port, contradicting both the new docs and the guarantee that only builtin/tabbed services are served there. The same pattern also exists for `baidu_master_service` in baidu_rpc_protocol.cpp, whose dispatch happens before the new gate. Please move the internal-port restriction ahead of these master-service dispatch blocks (or apply it centrally) so the gate covers every non-builtin path.

---

wwbmmm (commented):
The change correctly narrows internal_port to builtin/tabbed services and adds tests, but the new RejectNonBuiltinAccessFromInternalPort gate is not applied on every request-dispatch path, so the stated guarantee that only builtin services are served on internal_port does not fully hold.

---

src/brpc/policy/http_rpc_protocol.cpp
chenBright: fixed.

Skip `baidu_master_service` and  `http_master_service ` on internal_port breaks the equivalence.

chenBright:
> ### 🔵 Needs a closer look
> The new fd-close polling helper in `test/brpc_socket_unittest.cpp` can be flaky due to missing EINTR handling in the `fcntl(F_GETFD)` loop.
> 
> Review details
> ### Suppressed comments (1)
> **test/brpc_socket_unittest.cpp:130**
> 
> * WaitForClosedFd() polls fcntl(F_GETFD) but doesn’t handle EINTR; if a signal interrupts fcntl, the loop may exit early and the final ASSERT expects EBADF, causing flaky failures. Also, errno should be reset before the final fcntl so the EBADF check is guaranteed to correspond to that call.
> 
> ```
> void WaitForClosedFd(int fd) {
>     for (int i = 0; i < 100 && fcntl(fd, F_GETFD) != -1; ++i) {
>         bthread_usleep(10000);
>     }
>     ASSERT_EQ(-1, fcntl(fd, F_GETFD));
>     ASSERT_EQ(EBADF, errno);
> }
> ```
> 
> * **Files reviewed:** 18/18 changed files
> * **Comments generated:** 0 new
> * **Review effort level:** Lite

`fcntl` only returns `EINTR` for the blocking lock commands (`F_SETLKW/F_OFD_SETLKW`); 
`F_GETFD` just reads the `FD_CLOEXEC` bit from the fd table and can't be interrupted, so 
`EBADF` is its only realistic failure.

test/brpc_socket_unittest.cpp
```diff
@@ -118,6 +118,17 @@ class CheckRecycle : public brpc::SocketUser {
...
 
+// Acceptor::Join() returns as soon as Acceptor::BeforeRecycle() cleared
+// _listened_fd, which Socket::BeforeRecycled() calls several statements before
+// it closes the fd, so the fd may still be open when Join() returns.
+void WaitForClosedFd(int fd) {
+    for (int i = 0; i < 100 && fcntl(fd, F_GETFD) != -1; ++i) {
+        bthread_usleep(10000);
+    }
+    ASSERT_EQ(-1, fcntl(fd, F_GETFD));
+    ASSERT_EQ(EBADF, errno);
```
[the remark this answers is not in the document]
chenBright: `fcntl` only returns `EINTR` for the blocking lock commands (`F_SETLKW/F_OFD_SETLKW`); 
`F_GETFD` just reads the `FD_CLOEXEC` bit from the fd table and can't be interrupted, so 
`EBADF` is its only realistic failure.

chenBright:
I've verified the remaining comments against the code. They're about comment phrasing and test details rather than real defects, so I'm leaving them as they are. I think the PR is ready.

wwbmmm:
LGTM

PR #3532, opened 2026-09-08
Ignore canceled requests in the circuit breaker

wushinanqiyi:

Canceled RPCs use ECANCELED, which currently contributes to circuit-breaker error statistics and immediately reopens a half-open circuit. Caller cancellation does not indicate a downstream server failure.

Ignore ECANCELED alongside ELIMIT at the start of CircuitBreaker::OnCallEnd. Canceled requests do not affect sampling, error cost, latency, or the successful half-open probe count.

Add regression tests for cancellation during initialization, interleaved cancellation after initialization, and cancellation in the half-open state. The tests also check that genuine errors still trigger isolation.

Performance effects: one additional error-code comparison per call.
Breaking backward compatibility: no API changes; cancellation no longer causes circuit-breaker isolation.

### Validation

- git diff --check: passed.
- Native brpc unit tests: not run locally; this Windows host has no configured Linux/WSL build environment or brpc dependencies. Linux/macOS CI validation is still required.

### DIFF
--- a/src/brpc/circuit_breaker.cpp
+++ b/src/brpc/circuit_breaker.cpp
@@ -18,6 +18,7 @@
 #include "brpc/circuit_breaker.h"
 
 #include <cmath>
+#include <errno.h>
 #include <gflags/gflags.h>
 
 #include "brpc/errno.pb.h"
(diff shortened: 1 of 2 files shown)


wwbmmm:
LGTM

chenBright (approved):
LGTM

PR #3534, opened 2026-09-10
refactor(butil): replace integer string conversions

darion-yaphet:

  Problem Summary:

  Replace internal uses of legacy `butil` integer-to-string helpers with `std::to_string` to reduce duplicated conversion logic.

  Changed:

  - Use `std::to_string` for internal `IntToString` and `UintToString` call sites.
  - Keep the public `butil::IntToString(int)` and `butil::UintToString(unsigned)` APIs as compatibility wrappers backed by `std::to_string`.
  - Update integer conversion tests while preserving unsigned conversion behavior for negative `int` inputs.

  Side effects:
  - Performance effects: No benchmark was run; no expected material impact.

  - Breaking backward compatibility: No. The existing public helper APIs remain available.

  ---
  ### Check List:
    - `cmake --build build --target BUTIL_LIB -j6`
    - `cmake --build build --target test_butil -j6`
  - [x] Related tests updated and passed.
    - `StringNumberConversionsTest.IntToString`
    - `HttpTest.http_head`

### DIFF
--- a/example/baidu_proxy_and_generic_call/proxy.cpp
+++ b/example/baidu_proxy_and_generic_call/proxy.cpp
@@ -17,9 +17,10 @@
 // todo
 // A proxy to receive EchoRequest and send back EchoResponse.
 
+#include <string>
+
 #include <gflags/gflags.h>
 #include <butil/logging.h>
-#include <butil/strings/string_number_conversions.h>
 #include <brpc/server.h>
 #include <brpc/controller.h>
 #include <brpc/channel.h>
@@ -65,7 +66,7 @@ public:
             FLAGS_load_balancer.c_str(), &options) != 0) {
             LOG(ERROR) << "Fail to initialize channel";
             (*cntl->response_user_fields())["x-bd-proxy-error-code"] =
-                butil::IntToString(brpc::EINTERNAL);
+                std::to_string(brpc::EINTERNAL);
             (*cntl->response_user_fields())["x-bd-proxy-error-text"] =
                 "Fail to initialize channel";
             return;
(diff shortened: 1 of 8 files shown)


darion-yaphet:
`IntToString` and `UintToString` were not removed directly because they are publicly exposed butil APIs. Removing them directly would cause external BRPC users to fail during compilation or linking.

chenBright (commented):
I don't see a clear advantage or necessity in switching to std::to_string here. It's just a stylistic swap. I'd suggest not making this change.

darion-yaphet:
Why should we re-implement the methods that are already provided by the standard library? There is no benefit at all.

wasphin:
Thanks for the cleanup. However, I think we need more benchmark results before making this change.

Replacing an existing implementation with std::to_string simply because the standard library provides it is not, by itself, a sufficient reason to change. We should demonstrate that the replacement is at least comparable in performance and does not introduce regressions on the toolchains we support.

In particular, the implementation of std::to_string varies across libstdc++ versions. Modern versions have optimized integer conversion, while older versions may implement it through vsnprintf, which can be significantly slower than the current butil implementation.

Could you please provide benchmarks covering representative integer types/values and, ideally, both older and newer supported GCC/libstdc++ versions? This would give us enough evidence to evaluate the trade-off rather than making the switch solely for API modernization.

darion-yaphet:
I have already explained that this PR is not intended to enhance performance, but rather to avoid duplication.

wwbmmm:
> I don't see a clear advantage or necessity in switching to std::to_string here. It's just a stylistic swap. I'd suggest not making this change.

+1

darion-yaphet:
OK. It seems it's not necessary anymore. Close.

PR #3535, opened 2026-09-11
feat: add common transport-level handshake for RDMA and UBSHM

zchuango:

Issue Number: N/A

Related Discussion: https://github.com/apache/brpc/discussions/3432

Problem Summary:

RDMA, UBSHM, and URMA use similar connection setup flows: establish a TCP control connection, exchange handshake messages, prepare transport-specific resources, negotiate whether the high-speed transport can be used, and fall back to TCP when necessary.

These handshake flows were previously implemented separately in individual transports, resulting in duplicated framing, TCP handshake I/O, state transitions, fallback handling, and connection orchestration.

As discussed in the related Discussion, the handshake orchestration should be moved above individual high-speed transports, while transport-specific resource management and data-plane operations remain inside each transport.

This PR introduces the common transport-level handshake framework and migrates both RDMA and UBSHM to it. URMA can be migrated to the same framework in a follow-up change.

Changed:

- Add `AdapterTransport` as the top-level transport for TCP, RDMA, and UBSHM sockets.
- Add common handshake session, framing, I/O, and adapter abstractions.
- Centralize handshake orchestration, state transitions, upgrade selection, and TCP fallback in the common transport layer.
- Move RDMA handshake orchestration out of `RdmaEndpoint`.
- Move UBSHM handshake orchestration out of `UBShmEndpoint`.
- Keep transport-specific resource allocation, negotiation, activation, and data-plane operations inside their respective transports and endpoints.
- Preserve the existing RDMA protocol ID and registered `rdma_handshake` protocol name.
- Preserve the existing RDMA v2/v3 wire formats.
- Preserve the existing UBSHM handshake wire format.
- Preserve TCP fallback when the requested high-speed transport is unavailable or negotiation fails.
- Add common handshake tests and transport-specific compatibility/fallback coverage for RDMA and UBSHM.

Side effects:
- Performance effects:
  - No intended data-plane performance change.
  - The common handshake framework only affects connection setup and transport negotiation.

  - No intended breaking change.
  - Existing RDMA v2/v3 wire formats are preserved.
  - Existing UBSHM wire format is preserved.
  - Existing TCP fallback behavior is preserved.

---
### Check List:
- Please follow [[Contributor Covenant Code of Conduct](https://github.com/apache/brpc/blob/master/CODE_OF_CONDUCT.md)](https://github.com/apache/brpc/blob/master/CODE_OF_CONDUCT.md).


src/brpc/rdma/rdma_endpoint.cpp
```diff
@@ -1480,16 +915,16 @@ void RdmaEndpoint::DeallocateResources() {
...
 
-        if (_resource->comp_channel != nullptr) {
-            if (_cq_sid != INVALID_SOCKET_ID) {
+    if (nullptr != _resource->comp_channel) {
                 // Destroy send_comp_channel will destroy this fd,
                 // so that we should remove it from epoll fd first
                 int fd = _resource->comp_channel->fd;
-                GetGlobalEventDispatcher(fd, _socket->_io_event.bthread_tag()).RemoveConsumer(fd);
+      GetGlobalEventDispatcher(fd, _socket->_io_event.bthread_tag())
+          .RemoveConsumer(fd);
```
wwbmmm: [replied by brpc-oncall robot] The inner `if (_cq_sid != INVALID_SOCKET_ID)` guard was dropped during the reformatting. `RemoveConsumer(comp_channel->fd)` is now called even when this endpoint never registered that fd with the event dispatcher, i.e. when `DeallocateResources()` runs after `AllocateResources()` but before `StartCqEvents()` (which is exactly what happens on TCP fallback via `DeactivateUpgrade()` -> `RdmaEndpoint::Reset()`). Calling `epoll_ctl(EPOLL_CTL_DEL)` on an fd that was never added fails with ENOENT and logs misleading "Fail to remove fd" warnings. Please restore the `_cq_sid` guard (and keep `remove_consumer` as before).

src/brpc/rdma_transport.cpp
```diff
@@ -70,31 +68,54 @@ int RdmaTransport::Reset(int32_t expected_nref) {
...
 
 std::shared_ptr<AppConnect> RdmaTransport::Connect() {
-    if (_default_connect == nullptr) {
-        return  std::make_shared<rdma::RdmaConnect>();
-    }
-    return _default_connect;
+  return _default_connect;
+}
+
+void RdmaTransport::SetHighSpeedAvailable(bool available) {
```
wwbmmm, line 74: [replied by brpc-oncall robot] This new code uses 2-space indentation while the rest of the file/repo follows the 4-space style, and the PR reindents many unrelated blocks (rdma_transport.cpp, rdma_endpoint.cpp, ub_endpoint.cpp). Please keep 4-space indentation and separate pure reformatting from functional changes, so the actual logic diffs stay reviewable and accidental drops such as the `_cq_sid` guard above are not buried in whitespace.

wwbmmm (commented):
[replied by brpc-oncall robot] The handshake refactor looks reasonable and the latest revision addresses the earlier upgrade-selection, synchronization and resource-release concerns, but an accidental logic drop in RdmaEndpoint and the unrelated 2-space reformatting should be fixed.

src/brpc/rdma/rdma_endpoint.cpp
zchuango: Fixed. The _cq_sid != INVALID_SOCKET_ID guard has been restored before removing the RDMA completion-channel consumer.

src/brpc/rdma_transport.cpp
zchuango: Fixed. The affected RDMA and UBSHM transport/endpoint code has been restored to 4-space indentation, and the accidental formatting noise was cleaned up.

src/brpc/adapter_transport.cpp
```diff
@@ -0,0 +1,612 @@
...
+    adapter->CompleteConnection(handshake::FAILED);
+    task->done(EPROTO, task->data);
+    return NULL;
+}
+
+void AdapterTransport::Init(Socket* socket, const SocketOptions& options) {
+    CHECK_EQ(_mode, options.socket_mode);
+    _socket = socket;
+    _default_connect = options.app_connect;
+    _on_edge_trigger = options.on_edge_triggered_events;
```
wwbmmm, line 365: [replied by brpc-oncall robot] This line does not leave `InputMessenger::OnNewMessages` active for client high-speed sockets, so I don't think the claimed race with `ProcessClientHandshake`'s blocking reads exists. Client sockets are created through `InputMessenger::Create(SocketOptions, SocketId*)` (Channel via `SocketMap`, and `Socket::GetPooledSocket`/`GetShortSocket`), which only sets `need_on_edge_trigger = true` and leaves `options.on_edge_triggered_events` NULL; the overload that assigns `OnNewMessages` has no callers. For a client high-speed socket `options.user == get_client_side_messenger()`, so this branch falls through to `_on_edge_trigger = OnNewDataFromTcp`, which only calls `NotifyReadable()` and never reads the fd. Pooled/short sockets propagate the same NULL value, so they also get `OnNewDataFromTcp`. Please double-check whether any caller sets `on_edge_triggered_events` for these sockets before changing this.

wwbmmm (commented):
[replied by brpc-oncall robot] Copilot's two critical findings are valid and must be fixed (AdapterConnect must not be re-wrapped for pooled/short sockets, and the server must reject a non-NUL-terminated remote.shm_name before strncpy/strlen), but the client edge-trigger race claim does not apply to the sockets this PR creates.

wwbmmm:
please resolve conflicts

docs/cn/handshake_common_design.md
```diff
@@ -0,0 +1,499 @@
...
+
+```text
+Socket -> AdapterTransport -> TcpTransport / HighSpeedTransport
+Socket -> AdapterTransport -> ConnectionUpgradeCoordinator
+ConnectionUpgradeCoordinator -> TransportUpgradeOps
+ConnectionUpgradeCoordinator -> HandshakeSession
+TransportUpgradeOps -> transport-specific endpoint/resource
+```
+
+禁止以下反向依赖：
```
wwbmmm, line 57: AI废话太多。

wwbmmm (commented):
This PR was intended to reduce duplicated code logic, but actually it added more lines of code than removed code.

test/brpc_rdma_unittest.cpp
```diff
@@ -169,139 +168,122 @@ static void ConnectToServer(butil::fd_guard* sockfd) {
...
-            LOG(INFO) << "sleep " << req->sleep_us() << "us...";
-            bthread_usleep(req->sleep_us());
-        }
-        res->set_message("MyEchoService");
-        if (req->code() != 0) {
-            res->add_code_list(req->code());
-        }
-        cntl->response_attachment().append(cntl->request_attachment());
-    }
+  void Echo(google::protobuf::RpcController *cntl_base,
```
chenBright, line 171: Do not change the code indentation from 4 to 2.

chenBright (changes requested):

docs/cn/handshake_common_design.md
zchuango: Thanks for the feedback. We’re looking into this and will continue simplifying the common handshake design to remove unnecessary complexity and duplication. We’ll update the PR once the changes are ready.

test/brpc_rdma_unittest.cpp
zchuango: Thanks for pointing this out. We’re cleaning up the indentation and unrelated formatting changes as part of the current update.

docs/cn/handshake_common_design.md
zchuango: Updated. We reworked the common handshake abstraction and replaced the callback-based compatibility layer with explicit HandshakeProtocol and HandshakeTransport participant interfaces.

This removes the callback conversion wrappers and makes the protocol-specific wire handling and transport resource lifecycle responsibilities explicit. The design document has also been updated accordingly.

Could you please take another look when you have time?

test/brpc_rdma_unittest.cpp
zchuango: Fixed. The RDMA test code has been restored to the existing 4-space indentation, and the unrelated formatting changes have been removed.

PR #3536, opened 2026-09-11
Reject empty MySQL response packets

wasphin:

Issue Number:

Problem Summary:

The MySQL response parser reads the response type byte without first
rejecting packets whose declared payload length is zero. When another packet
follows, the parser may treat that packet's first byte as the empty packet's
type.

Changed:

- Reject zero-length MySQL response packets before reading their type.
- Apply the same validation after a fast-auth marker.
- Cover standalone and followed-by-another-packet cases.

Side effects:

- Performance effects: One payload-length check while determining a response
  type.
- Breaking backward compatibility: Invalid zero-length response packets are
  rejected instead of being interpreted using following bytes.

---
### Check List:


wwbmmm:
LGTM

wasphin:
The two flaky unit tests are not related to this change, and I will fix them separately.

PR #3540, opened 2026-09-13
Stabilize timer and socket unit tests

wasphin:

Issue Number:

Problem Summary:

Two unit tests can fail under CI load because they rely on tight timing
assumptions. The timer test requires shutdown within 15 milliseconds, and the
socket test releases its socket after a fixed 50-millisecond delay without
ensuring that all asynchronous write callbacks have completed.

Changed:

- Keep the timer shutdown latency check well below the next scheduled task
  while allowing for CI runner scheduling delays.
- Wait for all shutdown-write callbacks with a bounded countdown event before
  releasing the socket.
- Stop repeated shutdown-write runs after the first test failure.

Side effects:

- Performance effects: Normal runs no longer incur the fixed 50-millisecond
  delay in every shutdown-write iteration.

---
### Check List:

### DIFF
--- a/test/brpc_socket_unittest.cpp
+++ b/test/brpc_socket_unittest.cpp
@@ -30,6 +30,7 @@
 #include "butil/fd_utility.h"
 #include "butil/debug/leak_annotations.h"
 #include <butil/fd_guard.h>
+#include "bthread/countdown_event.h"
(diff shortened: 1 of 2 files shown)


chenBright (approved):
LGTM

PR #3062, opened 2025-08-11
Add SingleIOBuf for efficient flatbuffers serialization/deserialization

Q1ngbo:
### What problem does this PR solve?

Issue Number:
#2354  #978

Problem Summary:
Hi, We plan to migrate the flatbuffers protocol used internally by Baidu to the community, which is expected to be divided into the following three steps:
1. Submit the data structure SingleIOBuf, which is used for Flatbuffers serialization and deserialization, to the apache/brpc repository;
2. Submit message construction and serialization/deserialization methods to the google/flatbuffers repository;
3. Add Flatbuffers-based RPC handling policy to apache/brpc.
These steps have a sequential dependency, and this commit completes the first step.
Flatbuffers requires that data to be serialized/deserialized be stored in a contiguous memory buffer. To meet this requirement, we designed SingleIOBuf. Unlike brpc’s existing IOBuf, SingleIOBuf uses a single Block to hold contiguous data, which better aligns with Flatbuffers' memory layout requirements. We also defined corresponding memory allocation and release interfaces to support efficient management of SingleIOBuf.

### What is changed and the side effects?

Changed:

Side effects:
- Performance effects:

- Breaking backward compatibility: 

---
### Check List:
- Please make sure your changes are compilable.
- When providing us with a new feature, it is best to add related tests.
- Please follow [Contributor Covenant Code of Conduct](https://github.com/apache/brpc/blob/master/CODE_OF_CONDUCT.md).

PR #3558, opened 2026-09-21
Add optional FlatBuffers messages and code generation

AnDiXL:

Related discussion: #2354.

This is the message/codegen follow-up to #3196, opened at @wwbmmm's [request](https://github.com/apache/brpc/pull/3196#issuecomment-5761639973). It submits the two commits already shared there as a focused, independently reviewable PR.

Background and related work:

- #3062 (merged): the `SingleIOBuf` foundation used by this implementation.
- #3196: the original FlatBuffers message-construction proposal and review discussion.
- #3197: the associated RPC proposal. Network transport and Channel/Controller/Server integration remain separate and are **not included in this PR**.
- [LinQuickDev/brpc#45](https://github.com/LinQuickDev/brpc/pull/45): related FlatBuffers integration and validation work by @Spicy-cream.

Thanks to @Q1ngbo for the original work, and to @Spicy-cream for the continued interest in FlatBuffers support and collaboration around these follow-ups.

Changed:

- `3f2550c6`: add IOBuf-backed `Message`, allocator-aware `MessageBuilder`, stable service/method descriptors, optional CMake/Make/Bazel integration, documentation and regression tests.
- `6e5a3b32`: fix generated header-guard collisions and relocation, and make standalone codegen acceptance use the appropriate Protobuf includes, link target and C++ standard.
- `Message` derives from `NonreflectableMessage<Message>`; existing protobuf-facing `Protocol` callback signatures remain unchanged.
- The standalone `brpc_flatc` binding generator uses the upstream FlatBuffers parser rather than requiring a compiler fork.

Scope and side effects:

- FlatBuffers support is disabled by default.
- This is message construction and generated in-process dispatch, not an `fb_rpc` transport or a network RPC example. The RPC follow-up will carry those separately.
- No measured performance improvement is claimed.

### Validation

Recorded checks, not rerun for PR creation:

- Exact PR head (`6e5a3b32`), macOS/CMake: 21 message tests and both `flatbuffers_codegen_acceptance` and `flatbuffers_codegen_runtime` passed, with no failures or skips.
- Related full message/RPC snapshots were also validated on Linux with CMake, Make and Bazel. This is integration evidence, not a separate Linux validation of this reduced two-commit PR.
- Full-repository tests, sanitizer/performance results and hosted CI for this new PR are not claimed here.

Build and usage instructions are in [docs/en/flatbuffers.md](https://github.com/AnDiXL/brpc/blob/6e5a3b3248fc10671873497fc56574eb4ef8b5c9/docs/en/flatbuffers.md) and [tools/flatbuffers/README.md](https://github.com/AnDiXL/brpc/blob/6e5a3b3248fc10671873497fc56574eb4ef8b5c9/tools/flatbuffers/README.md).

---
### Check List:

- [x] Focused build and test validation is recorded above.
- [x] Message and code-generation regression tests are included.
- [x] Build instructions and ownership/verification limitations are documented.
- [ ] Hosted CI for this PR has completed.


docs/en/flatbuffers.md
```diff
@@ -0,0 +1,123 @@
+# FlatBuffers messages
```
wwbmmm, line 1: Could you also add a chinese version of this document?

wwbmmm:
I think this PR is more comprehensive than https://github.com/apache/brpc/pull/3196 . @Q1ngbo What do you think of this PR? Does it conflict with your previous work on flatbuffer?

docs/en/flatbuffers.md
AnDiXL: > Could you also add a chinese version of this document?

Of course, I will submit it tomorrow.

MODULE.bazel
```diff
@@ -79,3 +79,17 @@ git_repository(
...
+flatbuffers_http_archive = use_repo_rule(
+    '@bazel_tools//tools/build_defs/repo:http.bzl',
+    'http_archive',
+)
+flatbuffers_http_archive(
+    name = 'com_github_google_flatbuffers',
+    sha256 = 'b9c2df49707c57a48fc0923d52b8c73beb72d675f9d44b2211e4569be40a7421',
+    strip_prefix = 'flatbuffers-25.2.10',
+    urls = ['https://github.com/google/flatbuffers/archive/refs/tags/v25.2.10.tar.gz'],
+)
```
chenBright: What is different about `bazel_dep(name = "flatbuffers", version = "25.2.10")`?

docs/en/flatbuffers.md
AnDiXL: Added `docs/cn/flatbuffers.md` and linked it from the English document. This Chinese version is scoped to the message/codegen part of this PR; the RPC/example documentation remains in the follow-up draft.

MODULE.bazel
AnDiXL: `bazel_dep(name = "flatbuffers", version = "25.2.10")` is not equivalent here. The BCR module for FlatBuffers 25.2.10 imports gRPC and several JS/Go/Swift tooling dependencies for the full upstream build, while this PR only needs `//:runtime_cc` and `//:flatc`. Pulling that module would also bring gRPC's dependency graph, including another BoringSSL version, into Bzlmod resolution even when `BRPC_WITH_FLATBUFFERS` is false.

I kept the checksum-pinned `http_archive` and expanded the comment in `MODULE.bazel` to make this explicit.

AnDiXL:
For the hosted FlatBuffers ON matrix: I intentionally kept it in the follow-up draft #3559 rather than this message/codegen PR. #3559 adds `.github/workflows/flatbuffers-on.yml` plus the validation runner and covers the FlatBuffers-enabled CMake/Make/Bazel paths together with the RPC/example changes. This PR keeps the smaller message/codegen scope.

CMakeLists.txt
```diff
@@ -82,6 +83,17 @@ if(WITH_GLOG)
     set(BRPC_WITH_GLOG 1)
 endif()
 
+set(WITH_FLATBUFFERS_VAL "0")
+if(WITH_FLATBUFFERS)
+    find_path(FLATBUFFERS_INCLUDE_DIR NAMES flatbuffers/flatbuffers.h)
```
wwbmmm: [replied by brpc-oncall robot] `WITH_FLATBUFFERS` is validated only by the existence of `flatbuffers/flatbuffers.h`, but the runtime does not use FlatBuffers purely through its public API: `src/brpc/flatbuffers/message.cpp` reaches into builder internals (`buf_`, `buf_.swap_allocator()`, `scratch_push_small()`, `string_pool`, `minalign_`), and `SlabAllocator` asserts on the exact `allocate()`/`reallocate_downward()` size bookkeeping (only one live allocation, `old_size == _capacity`). Bazel pins 25.2.10 while CMake and `config_brpc.sh` accept any installed version, so with a different FlatBuffers an unsupported version fails deep in compilation or aborts inside the allocator. Please add a compile-time guard in `src/brpc/flatbuffers/message.h` (e.g. `#if !defined(FLATBUFFERS_VERSION_MAJOR) || FLATBUFFERS_VERSION_MAJOR < X` -> `#error`) and document the supported range, so users get a clear diagnostic instead of an obscure failure.

src/brpc/flatbuffers/message.h
```diff
@@ -0,0 +1,166 @@
...
+    template <typename T> const T* GetRoot() const {
+        return data() ? ::flatbuffers::GetRoot<T>(data()) : nullptr;
+    }
+    template <typename T> T* GetMutableRoot() {
+        return data() ? ::flatbuffers::GetMutableRoot<T>(mutable_data()) : nullptr;
+    }
+
+    // Failure leaves the old message unchanged. A fragmented or unaligned
+    // payload is copied into aligned storage; aligned contiguous input is shared.
+    bool parse_msg_from_iobuf(const butil::IOBuf& buf, size_t msg_size,
```
wwbmmm: [replied by brpc-oncall robot] `parse_msg_from_iobuf` (and therefore `ParseFbFromIOBUF`) accepts any `msg_size` below `FLATBUFFERS_MAX_BUFFER_SIZE` and, for fragmented or insufficiently aligned input, copies that many bytes into a fresh allocation before anything is verified. Once the RPC transport lands this length comes from an untrusted peer. Please state the caller contract here (and in `docs/en/flatbuffers.md`): the caller must bound `msg_size` with its own max-message-size setting and must call `Verify<T>()` before reading, so the receive-side limit is not left implicit.

wwbmmm (commented):
[replied by brpc-oncall robot] I agree with Copilot's two findings (the Make-generated header can shadow CMake's output, and hosted CI does not exercise any of the new FlatBuffers tests). My own remaining concerns are: the runtime binds to FlatBuffers internals without any version validation, Message::MergeFrom aborts the process rather than reporting an error, and the receive path's size/verification obligations are not stated. Please also confirm the service.h dispatch contract is what the RPC transport PR will consume, since it becomes public API here.

